PR #33, opened 2015-01-02
Fix dependencies for spark 1.2

zeitos:


zeitos:
this is interesting, I didn't get that error on my environment, looking into it

PR #35, opened 2015-01-05
Support storing rdds that have a storeLevel already set

zeitos:
The code was trying to override the StorageLevel on the given RDD, but some rdds are cached by spark (ie: the ones in a MatrixFactorization result of ALS.train)

### DIFF
--- a/job-server/src/spark.jobserver/JobServerNamedRdds.scala
+++ b/job-server/src/spark.jobserver/JobServerNamedRdds.scala
@@ -88,8 +88,8 @@ class JobServerNamedRdds(val rddManager: ActorRef) extends NamedRdds {
     val rdd = rddGen
     rdd.setName(name)
     rdd.getStorageLevel match {
-      case StorageLevel.NONE => rdd.persist(storageLevel) 
-      case currentLevel => rdd.persist(currentLevel) 
+      case StorageLevel.NONE => rdd.persist(storageLevel)
+      case currentLevel => rdd.persist(currentLevel)
     }
     // TODO: figure out if there is a better way to force the RDD to be computed
     if (forceComputation) rdd.count()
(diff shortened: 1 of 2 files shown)


job-server/test/spark.jobserver/NamedRddsSpec.scala
```diff
@@ -1,10 +1,11 @@
 package spark.jobserver
 
-import akka.actor.{PoisonPill, Props, ActorRef, ActorSystem}
+import akka.actor.{ActorRef, ActorSystem, Props}
 import akka.testkit.{ImplicitSender, TestKit}
 import org.apache.spark.SparkContext
+import org.apache.spark.storage.StorageLevel
 import org.scalatest.matchers.ShouldMatchers
-import org.scalatest.{FunSpec, BeforeAndAfterAll, BeforeAndAfter}
+import org.scalatest.{BeforeAndAfter, BeforeAndAfterAll, FunSpec}
```
zeitos: ups, looks like I have auto import enabled, should I revert the import change?

job-server/src/spark.jobserver/JobServerNamedRdds.scala
```diff
@@ -87,7 +87,11 @@ class JobServerNamedRdds(val rddManager: ActorRef) extends NamedRdds {
       "forceComputation implies storageLevel != NONE")
     val rdd = rddGen
     rdd.setName(name)
-    if (storageLevel != StorageLevel.NONE) rdd.persist(storageLevel)
+    val level = rdd.getStorageLevel match {
+      case StorageLevel.NONE => storageLevel
+      case _ => rdd.getStorageLevel
```
velvia: I would just combine the match and the next persist line.  Also no need to call `getStorageLevel` twice.
```
rdd.getStorageLevel match {
    case StorageLevel.NONE => rdd.persist(storageLevel)
    case alreadySet                => rdd.persist(alreadySet)
}
```

job-server/test/spark.jobserver/NamedRddsSpec.scala
velvia: Sure.

velvia:
Thanks, just minor code style comment.

zeitos:
Fixed the comments

velvia:
:+1:   going to merge, thanks!

PR #49, opened 2015-01-19
SPARK_CONF_HOME to SPARK_CONF_DIR

WangTaoTheTonic:
In Spark we normally use SPARK_CONF_DIR to represent Spark's config directory. Maybe changing the name here is better.


velvia:
Thanks, LGTM.

PR #52, opened 2015-01-25
Velvia/new sparkjob context api

omriiluz:


omriiluz:
Wrong direction, sorry!

PR #63, opened 2015-02-04
Change "mem-per-node" to "memory-per-node"

KaiXinXiaoLei:
"mem-per-node" is not valid, because in the config, using "memory-per-node". For example, using  "memory-per-node" in SparkJobUtils.scala, not "mem-per-node".


velvia:
Thanks @KaiXinXiaoLei esp for updating README as well.

PR #64, opened 2015-02-05
Timer utility to create Timer metric

petro-rudenko:
BTW maybe it's better to use [metrics-scala](https://github.com/erikvanoosten/metrics-scala) library

### DIFF
--- a/akka-app/src/ooyala.common.akka/metrics/YammerMetrics.scala
+++ b/akka-app/src/ooyala.common.akka/metrics/YammerMetrics.scala
@@ -1,7 +1,7 @@
 package ooyala.common.akka.metrics
 
 import com.yammer.metrics.Metrics
-import com.yammer.metrics.core.{Histogram, Meter, Gauge}
+import com.yammer.metrics.core.{Histogram, Meter, Gauge, Timer}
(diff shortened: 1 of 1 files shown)


seregasheypak:
Could you please advice how can we collect metrics for the jobs? default spark history-server doesn't work because SparkContext is never stopped.

petro-rudenko:
I'm just using  metricsWriter.processTimer utility to log into file:
```scala
val logger = (new Log4jLoggerFactory).getLogger(this.getClass.getName)
val metricsWriter = new CompactMetricsWriter(logger)
val modelFitTimer = Metrics.newTimer(this.getClass, "Model fit time", TimeUnit.SECONDS, TimeUnit.MILLISECONDS)
val cvModel = modelFitTimer.time(new Callable[CrossValidationModel] {
      def call = train(data)
})
 metricsWriter.processTimer(new MetricName(this.getClass, "Model fit time"), modelFitTimer, logger)
```

this will output the statistics on timer metrics.

seregasheypak:
We are trying to reuse spark-history server functionality or publish smth
useful using default spark metrics interface.

2015-04-20 12:29 GMT+02:00 Peter Rudenko <notifications@github.com>:

PR #65, opened 2015-02-06
Add uris of dependence jars when creating context

KaiXinXiaoLei:
When creating context, I want to add uris of dependent jars.  And i want to add different jars for different context. If option "dependent-jar-uris" is a list, there is impossible to using "POST /contexts/<name>?dependent-jar-uris=[file:///some/path/of/my-foo-lib.jar]" to create context. So i change the type of "dependent-jar-uris".

For example:
curl -d "" 'localhost:8090/contexts/test-context?dependent-jar-uris=file:///some/path/of/my-foo-lib.jar,file:///some/path/of/my-boo-lib.jar'

In this way, when creating test-context, i pass the dependence jars for this context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,9 @@ the REST API.
 ### Context configuration
 
 A number of context-specific settings can be controlled when creating a context (POST /contexts) or running an
-ad-hoc job (which creates a context on the spot).
+ad-hoc job (which creates a context on the spot).  For example, add urls of dependent jars for a context.
+
+    POST '/contexts/my-new-context?dependent-jar-uris=file:///some/path/of/my-foo-lib.jar'
 
 When creating a context via POST /contexts, the query params are used to override the default configuration in
 spark.context-settings.  For example,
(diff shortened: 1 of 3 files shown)


velvia:
Hi @KaiXinXiaoLei 

Thanks for the PR.   

I understand your desire to pass in the dependent URIs as a query param.  However, I think the right base type in a config file is a string list, because this is much less errorprone.  So, here is a compromise:

- Don't change the base type in the application.conf, leave it as a string list
- In `JobManagerActor`, parse it first as a StringList, then try it as a String.  For example:

```
   Try(config.getStringList("dependent-jar-uris").asScala.toSeq).
     orElse(Try(config.getString("dependent-jar-uris").split(",").toSeq)).getOrElse(Nil)
```

Also would you mind adding a note of this to the README?  Thanks.

KaiXinXiaoLei:
@velvia 
I had change. Thanks.

velvia:
Otherwise this looks great to me!

PR #75, opened 2015-02-26
Update local.sh.template for YARN config

cerebraldna:
Uncomment YARN_CONF_DIR and replace pathToRemoteConf when running remotely connecting to YARN


velvia:
@cerebraldna thanks.

PR #77, opened 2015-03-02
File line length exceeds 110 characters

KaiXinXiaoLei:
This lines not satisfy scala-style


velvia:
Thanks, I fixed this just now.  If you pull the latest from master it'll work.

PR #78, opened 2015-03-02
uris of jars is a string list, or a string separated by commas ','

KaiXinXiaoLei:
When creating context, I want to add uris of dependent jars. And i want to add different jars for different context. If option "dependent-jar-uris" is a list, there is impossible to using "POST /contexts/?dependent-jar-uris=[file:///some/path/of/my-foo-lib.jar]" to create context. So i change the type of "dependent-jar-uris".

For example:
curl -d "" 'localhost:8090/contexts/test-context?dependent-jar-uris=file:///some/path/of/my-foo-lib.jar,file:///some/path/of/my-boo-lib.jar'

In this way, when creating test-context, i pass the dependence jars for this context.

See: https://github.com/spark-jobserver/spark-jobserver/pull/65

### DIFF
--- a/job-server/config/local.conf.template
+++ b/job-server/config/local.conf.template
@@ -41,7 +41,6 @@ spark {
 
     # uris of jars to be loaded into the classpath for this context. Uris is a string list, or a string separated by commas ','
     # dependent-jar-uris = ["file:///some/path/present/in/each/mesos/slave/somepackage.jar"]
-    # dependent-jar-uris = "file:///some/path/of/my-foo-lib.jar,file:///some/path/of/my-boo-lib.jar"
     
     # If you wish to pass any settings directly to the sparkConf as-is, add them here in passthrough,
     # such as hadoop connection settings that don't use the "spark." prefix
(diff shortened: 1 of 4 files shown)


KaiXinXiaoLei:
@velvia 
i change from https://github.com/spark-jobserver/spark-jobserver/pull/65 to this.

job-server/config/local.conf.template
```diff
@@ -39,8 +39,9 @@ spark {
     # in case spark distribution should be accessed from HDFS (as opposed to being installed on every mesos slave)
     # spark.executor.uri = "hdfs://namenode:8020/apps/spark/spark.tgz"
 
-    # uris of jars to be loaded into the classpath for this context
+    # uris of jars to be loaded into the classpath for this context. Uris is a string list, or a string separated by commas ','
     # dependent-jar-uris = ["file:///some/path/present/in/each/mesos/slave/somepackage.jar"]
+    # dependent-jar-uris = "file:///some/path/of/my-foo-lib.jar,file:///some/path/of/my-boo-lib.jar"
```
velvia: I'd prefer to keep the example as a string list.  Much more type or parser safe.

velvia:
Looks good for the most part, just a comment on the doc.  Also do you want to at least add a comment to `application.conf`?  All settings should be documented there as well.  Thx.

KaiXinXiaoLei:
I  think there is not need to add a comment to the file  application.conf. Because users see the local.conf.template to get the config option.

KaiXinXiaoLei:
@velvia  I haved changed application.conf

velvia:
@KaiXinXiaoLei thanks.  LGTM.

PR #79, opened 2015-03-02
Swagger integration

zeitos:
Moved routes to their own traits and added some doc.

Also fixed line width in YammerMetrics (not sure how it got into master, sbt package failed)

There's a lot of html and css that actually belongs to swagger-ui

I hardcoded the url to be /docs... we might want to make this configurable though


zeitos:
Just a side comment: The annotations of spray-swagger are super verbose :S

zeitos:
I think I'll close this one since it's super outdated

velvia:
We should pick it up again when we have time. 

-Evan

PR #80, opened 2015-03-02
Update job-server-tests jar path

curran:
The original test command `curl --data-binary @job-server-tests/target/job-server-tests-0.4.1.jar localhost:8090/jars/test` failed, because the jar path was wrong. The correct path is `job-server-tests/target/job-server-tests-0.4.2-SNAPSHOT.jar`. This is the jar that was generated by running `sbt job-server-tests/package` as instructed in the README, so it seems the documentation needs to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Then go ahead and start the job server using the instructions above.
 
 Let's upload the jar:
 
-    curl --data-binary @job-server-tests/target/job-server-tests-0.4.1.jar localhost:8090/jars/test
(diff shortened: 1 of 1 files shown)


curran:
Alternatively, the README should include (under "Getting Started")

First clone the repository and check out the 0.4.1 branch by running:

    git clone https://github.com/spark-jobserver/spark-jobserver.git
    cd spark-jobserver
    git checkout tags/v0.4.1

PR #81, opened 2015-03-02
Improve Quick Start Guide

curran:
I faced some issues when getting started that could have been avoided if a note about version tag checkout had been in the README, so I thought I'd add it. My issue was that the command `curl --data-binary @job-server-tests/target/job-server-tests-0.4.1.jar localhost:8090/jars/test` was failing with

    Warning: Couldn't read data from file 
    Warning: "job-server-tests/target/job-server-tests-0.4.1.jar", this makes an 
    Warning: empty POST.

I had to change it to `curl --data-binary @job-server-tests/target/job-server-tests-0.4.2-SNAPSHOT.jar localhost:8090/jars/test`. Then I realized I should be using v0.4.1 in the first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,13 @@ For release notes, look in the `notes/` directory.  They should also be up on [l
 
 ## Quick start / development mode
 
-You need to have [SBT](http://www.scala-sbt.org/release/docs/Getting-Started/Setup.html) installed.
+First clone the repository and check out the v0.4.1 version by running:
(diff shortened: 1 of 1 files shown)


velvia:
@curran I just released 0.5.0 and modified the README to have what I suggested to you.  Check it out and see if it helps.

curran:
The new version looks great, thank you!

PR #89, opened 2015-03-15
Spark-1.3 support

petro-rudenko:


velvia:
@petro-rudenko thanks for the PR.  Looks like there is another one which is exactly the same, but without the log4j line.   I think the other one will be merged soon, either way master will be on 1.3.0 soon.

PR #90, opened 2015-03-15
added HiveContextFactory, Hive dependencies and SparkHiveJob trait

ajmerlob:
I've cleared the way for jobs that require a HiveContext instead of a SQLContext.  That said, this PR forces all builds to include the Hive dependencies, which are unnecessary for many projects.  The issue with excluding the dependencies by default is that SparkJob.scala and HiveContextFactory.scala depend on Hive at compile-time; to optionally exclude Hive dependencies, the SparkHiveJob trait in SparkJob.scala and essentially all of HiveContextFactory.scala would need to be conditionally excluded from compiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ See [Troubleshooting Tips](doc/troubleshooting.md).
 ## Features
 
 - *"Spark as a Service"*: Simple REST interface for all aspects of job, context management
-- Support for Spark SQL Contexts/jobs and custom job contexts!  See [Contexts](doc/contexts.md).
+- Support for Spark SQL and Hive Contexts/jobs and custom job contexts!  See [Contexts](doc/contexts.md).
 - Supports sub-second low-latency jobs via long-running job contexts
 - Start and stop job contexts for RDD sharing and low-latency jobs; change resources on restart
 - Kill running jobs via stop context
--- a/job-server-api/src/spark.jobserver/SparkJob.scala
+++ b/job-server-api/src/spark.jobserver/SparkJob.scala
@@ -3,6 +3,7 @@ package spark.jobserver
 import com.typesafe.config.Config
 import org.apache.spark.SparkContext
 import org.apache.spark.sql.SQLContext
+import org.apache.spark.sql.hive.HiveContext
 
 sealed trait SparkJobValidation {
   // NOTE(harish): We tried using lazy eval here by passing in a function
@@ -50,4 +51,8 @@ trait SparkJob extends SparkJobBase {
 
 trait SparkSqlJob extends SparkJobBase {
   type C = SQLContext
+}
+
+trait SparkHiveJob extends SparkJobBase {
+  type C = HiveContext
 }
--- a/job-server-tests/src/spark.jobserver/HiveTestJob.scala
+++ b/job-server-tests/src/spark.jobserver/HiveTestJob.scala
@@ -1,0 +1,51 @@
+package spark.jobserver
+
+import com.typesafe.config.{Config, ConfigFactory}
+import org.apache.spark._
+import org.apache.spark.sql.hive.HiveContext
+
+/**
+ * A test job that accepts a HiveContext, as opposed to the regular SparkContext.
+ * Initializes some dummy data into a table, reads it back out, and returns a count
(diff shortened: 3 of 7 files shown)


zeitos:
Hi Aaron, can you include a test  (take a look at SqlTestJob or SqlJobSpec) ?

ajmerlob:
Drat, I knew I missed something :-)  No problem.

velvia:
@ajmerlob thanks.  Hmm....   could we make the dependencies "provided" like the Spark dependencies?

velvia:
@ajmerlob sorry I spoke too soon.  It's already provided, LGTM!   (once tests are added... it'd be great to have an example.)   Also, perhaps add a note to README, in the initial section can say that Hive jobs are supported?

ajmerlob:
scalastyle and I didn't see eye-to-eye on this one.  Can somebody show me the light on my 200 character Hive-SQL query?

ajmerlob:
FYI: On my machine, one test is failing:  "can add contexts from jobConfig" under LocalContextSupervisorSpec.

velvia:
@ajmerlob hmm... not sure where ScalaStyle is flagging you.  You could potentially turn off sections with //scalastyle:off ... 

Also looks like Travis CI failed because it ran out of memory.  Does Hive use that much more memory?

velvia:
Jenkins please retest

ajmerlob:
@velvia Not sure why it would use that much memory, but it's possible - the first time Hive launches, I can imagine it might consume more resources than normal.  (It's going to do a bunch of checks, and ultimately create the metastore_db and Hive warehouse on HDFS if they don't exist / weren't configured - that's why I moved the timeout provision to 120 seconds vs. 6 seconds in the original SQL script).

zeitos:
@ajmerlob do you think we could use https://github.com/klarna/HiveRunner instead of running hive?

ajmerlob:
Looks plausible, and... like a pain to switch over...  I guess I'll take a
look at it later today.  In the meantime, can ya just add some memory and
say "retest this please" :-)

Aaron

ajmerlob:
Yea .. so I couldn't even import that package correctly.  If someone can
give me a hand with getting it to compile, I can take a look at actually
using it.

velvia:
Hmm....  so I'm not sure we can get Travis CI to increase the memory.  If anyone can help that'd be great.
https://travis-ci.com/plans

@ajmerlob do the tests pass on your box now?

ajmerlob:
I'm just using a low-grade EC2 instance as my box. I just ran sbt test, and
"can add contexts from jobConfig" fails (which I've noted elsewhere), but
all the Hive tests pass.

ajmerlob:
It's not obvious to me why "can add contexts from jobConfig" is failing, since it fails on my machine even with a clean build

ajmerlob:
So I did some snooping on this "can add contexts from jobConfig" test failing.  Here's a fun fact: if I copy-pasted the test, so that it ran twice in a row.  It works the second time.  Is there some service that's taking it's time to spin up while the test is trying to run?

ajmerlob:
I get two unique log messages on the first try (the time it fails).  Here's the log, minus everything that's the same.

[2015-03-19 04:15:36,356] WARN  doop.util.NativeCodeLoader [] [akka://test/user/$b/olap-demo] - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
[2015-03-19 04:15:37,295] WARN  k.jobserver.JobResultActor [] [] - Shutting down spark.jobserver.JobResultActor
[2015-03-19 04:15:40,349] ERROR ocalContextSupervisorActor [] [] - Exception after sending Initialize to JobManagerActor

velvia:
@ajmerlob it might be because of the timeout.  The LocalContextSupervisorActor has to wait for an Initialized message from the JobManagerActor to come back, and on the box you are on which is a small VM it might take extra time.  Try increasing the timeout, I think the default in the test might be 3 seconds...   the exception should give us a clue.

I have restarted the build, to see if Travis CI will pass this time.

junyo-aaron:
I'm closing this for now - I'll try to pick it up back up later.

PR #92, opened 2015-03-16
fix: Serialize empty maps.

jonbuffington:
Use a pattern guard to filter cases where a map is empty. Empty maps
cause an exception without the guard in place.

This patch was inspired by 434337e in the ooyala master branch.

### DIFF
--- a/akka-app/test/ooyala.common.akka/web/JsonUtilsSpec.scala
+++ b/akka-app/test/ooyala.common.akka/web/JsonUtilsSpec.scala
@@ -28,6 +28,24 @@ class JsonUtilsSpec extends FunSpec with Matchers {
       JsonUtils.listFromJson(json) should equal (batch)
     }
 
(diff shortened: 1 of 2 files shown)


akka-app/test/ooyala.common.akka/web/JsonUtilsSpec.scala
```diff
@@ -28,6 +28,24 @@ class JsonUtilsSpec extends FunSpec with Matchers {
       JsonUtils.listFromJson(json) should equal (batch)
     }
 
+    it("should serialize first-level empty maps to JSON") {
```
zeitos: Do we need to test for 
```
val expected = "{}" 
```

?

jonbuffington: Great idea. I updated the pull request.

velvia:
@jonbuffington thanks.

PR #93, opened 2015-03-17
the path of spark-job-server.jar is changed when building

KaiXinXiaoLei:
when running the script server-deploy.sh,  there is a error:"job-server/target/spark-job-server.jar: No such file or directory". I had checked the path of spark-job-server.jar is job-server/target/scala-2.10/spark-job-server.jar.


velvia:
@KaiXinXiaoLei thanks!  I was going to do something like this.

PR #94, opened 2015-03-17
server_start.sh and server_stop.sh should running in any path

KaiXinXiaoLei:
when running server_start.sh, the $PIDFILE should be created in the path of $appdir. For example, running $appdir/server_start.sh in the current directory (eg,  /home), the PIDFILE  should not created in the path of  /home/. Because there is a issue to find PIDFILE  when running  server_stop.sh.


zeitos:
this is hardcoding to 2.10, I'm creating a new PR to use SCALA_VERSION variable

velvia:
@zeitos want to merge this then you can base your changes on these changes too

zeitos:
ok, go ahead and merge this one first.

PR #95, opened 2015-03-17
Adding debian packaging

skate056:


velvia:
@skate056 thanks for the PR.
1) why an extra config file in spark-jobserver/config?  I think it's better to stick to the convention we have, which is templates in the config/ dir at the root, and user copies over and configures/changes however they want.
2) How about using https://github.com/sbt/sbt-native-packager instead?  That can generate more than just debs, but RPMs, and Docker images as well.

skate056:
Agree with your comments. However I wanted to provide a default configuration which would be used when the debian package is installed. If I don't provide the default configuration the service will not start up which I did not want.

I looked at the native package manager but realised that I could not get it working without extensive changes to the existing scripts. I also intended to use the lsb-functions for pid file management etc but all that is done but the existing scripts and it is not possible to change that without refactoring the existing infrastructure.

velvia:
Ah, got it.  Do you think your configuration would be generally useful to
other folks?

On Thu, Apr 9, 2015 at 9:05 AM, Saket <notifications@github.com> wrote:

> Agree with your comments. However I wanted to provide a default
> configuration which would be used when the debian package is installed. If
> I don't provide the default configuration the service will not start up
> which I did not want.
>
> I looked at the native package manager but realised that I could not get
> it working without extensive changes to the existing scripts. I also
> intended to use the lsb-functions for pid file management etc but all that
> is done but the existing scripts and it is not possible to change that
> without refactoring the existing infrastructure.
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/95#issuecomment-91275456>
> .
>



-- 
   --- Toni Morrison

skate056:
I think so. We are using the default config I put in. Ideally in a production deployment it will be controlled by a tool such as puppet or chef.

velvia:
@skate056 I'm going to play around with sbt-native-packager and get back to you.   The only issue I have with this PR is I don't think including a default config works in the project, because everyone will have a different default they want - and you include some commented out lines for example.  Instead let's throw an error in the scripts if a user did not create a default config.

Also I'm going to see if I could kill two birds with one stone, or is better to use your PR  + sbt-docker.  I want to get docker up and running for Mesos/Marathon deploys.

ogirardot:
any updates for sbt-native-packager ?

hntd187:
This pull is over a year old @velvia can we close this?

velvia:
Agreed, let’s close it

PR #96, opened 2015-03-17
Cancel job functionality

petro-rudenko:


job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -217,6 +217,26 @@ class WebApi(system: ActorSystem,
           }
         }
       } ~
+        //  DELETE /jobs/<jobId>
+        //  Stop the current job. All other jobs submited with this spark context
```
zeitos: nit: It stops the job with the specified jobId right?

petro-rudenko: Yes, jobserver job id (that returned after job submitted to jobserver).

velvia:
@petro-rudenko  This looks great so far. 

Can we add a JobKilled status message, to be sent to the statusActor, when a job is killed?   This should also result in a DAO/Database update.  Basically, after a job is killed, if I look up a job using the jobs route, I should see that the status was "killed".

Thanks!

petro-rudenko:
Take a look. The problem is that cancelJobGroup is an asynchronous operation which will cause eventually a SparkException in job thread. So it will seems like a failed job.

job-server/test/spark.jobserver/JobManagerSpec.scala
```diff
@@ -195,5 +196,20 @@ abstract class JobManagerSpec extends JobSpecBase(JobManagerSpec.getNewSystem) {
         case JobResult(_, result: Int) => result should equal (1 + 2 + 3)
       }
     }
+
+    it("should be able to cancel running job") {
```
zeitos: Can you add a test for what happens when the job is not running?

petro-rudenko: It will just call sparkContext.cancelJobGroup(NONEXISTING_GROUP_ID) which is safe (will not raise any exception) and will do actually nothing. In WebAPI it will actually handle this:
https://github.com/petro-rudenko/spark-jobserver/blob/cancel_job3/job-server/src/spark.jobserver/WebApi.scala#L228

velvia:
@petro-rudenko Thanks.  Hmm... so if we merged it now,  then a killed job will show up as a job that errored out, is that right?    Could that be tested in the unit test?

I think this is important enough to merge in soon, and we can deal with the status issues later.

job-server/test/spark.jobserver/JobManagerSpec.scala
zeitos: yes, I'm saying there's no test for that scenario, the user should receive a message that no such job exists right?

job-server/src/spark.jobserver/WebApi.scala
zeitos: yes but, what I'm trying to say is that the documentation says "current job" , there might be several jobs currently working, you just want to cancel one.

job-server/test/spark.jobserver/JobManagerSpec.scala
zeitos: also, if you try to delete a job that already finish what would be the result?

petro-rudenko:
I don't quite understand [this line](https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server/src/spark.jobserver/JobManagerActor.scala#L239). When i send cancelJobGroup command to spark eventually in job future thread there would be sparkException and the logic will come there. Also i cannot access to statusActor from JobManagerActorSpec to test that i'll get this message.

job-server/test/spark.jobserver/JobManagerSpec.scala
velvia: @zeitos It seems from the code that deleting a nonexistant job would get the user the NotFound message.   I think the same thing if the job already finished.

velvia:
@petro-rudenko hm, not sure, I'll have to review the logic.
As far as statusActor, the way it works is that you don't access statusActor directly, but rather subscribe to it.  The parent who created the JobManagerActor automatically subscribes to the StatusActor messages, so all you have to do is check in the test body that you are getting status messages.  If you look at the other tests in `JobManagerActorSpec`, you see tons of expectMsg(....) or expectMsgClass(...) to check status updates.    Hope that helps!  

Basically, in line 209 of your JobManagerSpec, in the test case that you wrote, instead of `expectNoMsg()`, I would think now that you would want something like `expectMsg(JobKilled)`.   Unless you are getting the SparkException and the job dies first.... 

LMK if you'd like a hangout or something to sort stuff out.

petro-rudenko:
Ok got it. JobKilled would be first, because cancelJobGroup is an asynchronous operation and theoretically it's even possible that cancelJobGroup was called, but job was in some millisseconds to finish, so we'll get JobFinished result.

v-gerasimov:
I really need this functionality with my work. What about merging this pull request?

velvia:
@petro-rudenko I think will merge this.   Don't see any reasons not to, just minor lingering issues that can be addressed in the future. Thanks!

noorul:
Will this get back ported to earlier versions?

velvia:
Probably won't have time for it myself, but feel free to submit a PR for
this, it's fairly easy once you locate the original PR.

On Fri, Apr 24, 2015 at 2:09 AM, Noorul Islam K M <notifications@github.com>
wrote:

> Will this get back ported to earlier versions?
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/96#issuecomment-95864128>
> .
>



-- 
   --- Toni Morrison

PR #97, opened 2015-03-17
Fix deployment scripts to use SCALA_VERSION

zeitos:
Leftover from the cross build PR.

### DIFF
--- a/job-server/test/spark.jobserver/JobSpecBase.scala
+++ b/job-server/test/spark.jobserver/JobSpecBase.scala
@@ -28,8 +28,7 @@ trait JobSpecConfig {
       "spark.jobserver.max-jobs-per-context" -> MaxJobsPerContext,
       "akka.log-dead-letters" -> 0,
       "spark.master" -> "local[4]",
-      "context-factory" -> contextFactory,
-      "akka.test.single-expect-default" -> 6000
+      "context-factory" -> contextFactory
     )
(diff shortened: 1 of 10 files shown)


job-server/test/spark.jobserver/JobSpecBase.scala
```diff
@@ -28,7 +28,8 @@ trait JobSpecConfig {
       "spark.jobserver.max-jobs-per-context" -> MaxJobsPerContext,
       "akka.log-dead-letters" -> 0,
       "spark.master" -> "local[4]",
-      "context-factory" -> contextFactory
+      "context-factory" -> contextFactory,
+      "akka.test.single-expect-default" -> 6000
```
velvia: Wait, this doesn't have to do with the script changes....

velvia:
@zeitos thanks.  Might be time to refactor out some of the common script logic into a common.sh...

job-server/test/spark.jobserver/JobSpecBase.scala
zeitos: nope, you're right but I saw a couple of test failing on a slow machine that I have with 2.11

zeitos:
I totally forgot about this, I'll try to catch up this week.

velvia:
@zeitos thanks, LGTM, merging.

PR #98, opened 2015-03-17
Clarify ad-hoc vs. permanent context modes in README

ajmerlob:
I spent at least an hour pulling out my hair when my NamedRdds would disappear at the last moment.  Turns out there's a big difference between ad-hoc queries and a pre-configured context when your queries are related to each other.

This documentation change aims to clarify that difference with the fewest possible changes to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ EXTRA_JAR for adding a jar to the classpath.
 
 ### WordCountExample walk-through
 
+#### Package Jar - Send to Server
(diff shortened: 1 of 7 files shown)


zeitos:
Great!, but you are including not only the README changes but hive code that breaks the build :(

ajmerlob:
I was trying to be so careful with my branches... Oh well. I'll submit it
again.

Aaron

PR #99, opened 2015-03-18
Clarify ad-hoc vs permanent context modes in README

ajmerlob:
I spent at least an hour pulling out my hair when my NamedRdds would disappear at the last moment. Turns out there's a big difference between ad-hoc queries and a pre-configured context when your queries are related to each other.

This documentation change aims to clarify that difference with the fewest possible changes to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ EXTRA_JAR for adding a jar to the classpath.
 
 ### WordCountExample walk-through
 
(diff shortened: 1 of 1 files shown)


zeitos:
I think that when you say "Queries" you are actually referring to "Jobs"

ajmerlob:
I think you are right - Jobs is more accurate.

PR #101, opened 2015-03-20
Removed confused assignment of execution context.

mehmeteking:
Previously it was intended to use "SparkContext's ActorSystem threadpool for the futures, so we don't corrupt our own"; however the assignment
  implicit val executionContext = sparkEnv.actorSystem
does not serve said purpose. It is removed.

### DIFF
--- a/job-server/src/spark.jobserver/JobManagerActor.scala
+++ b/job-server/src/spark.jobserver/JobManagerActor.scala
@@ -179,8 +179,6 @@ class JobManagerActor(dao: JobDAO,
                            jobContext: ContextLike,
                            sparkEnv: SparkEnv,
                            rddManagerActor: ActorRef): Future[Any] = {
-    // Use the SparkContext's ActorSystem threadpool for the futures, so we don't corrupt our own
-    implicit val executionContext = sparkEnv.actorSystem
 
     val jobId = jobInfo.jobId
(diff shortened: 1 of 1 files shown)


mehmeteking:
@velvia, did you have a chance to have a look at this issue?

velvia:
@mehmeteking So I just looked at this, thanks for the PR.
My only worry is, will this mean that `import context.dispatcher` is the one used?  The problem with that is that then the global thread pool for the ActorSystem will get used, and will limit scalability of the number of futures to run within job server.

Anyhow the PR is fine, since Spark may remove Akka altogether.  What do you think about having the JobManagerActor create a separate threadpool for its futures?  Would you be interested in that PR?

mehmeteking:
Hi Evan,

I would definitely be interested. How should we decide the number of threads to assign? We can perhaps assign as many threads as asked for; i.e. No upper limit?

Ekin.

velvia:
Hi Ekin,

I personally don't like the idea of unlimited by default ... Means you can easily run out of memory as a queue for jobs and futures.  I say let it be configurable limit.  

-Evan

mehmeteking:
Evan,

I do not like unbounded resource allocation, neither. A configurable parameter would serve well. I will look into it.

Ekin.

PR #111, opened 2015-04-02
Added ad-hoc ExecutionContext to dispatch Spark jobs.

mehmeteking:
Added ExecutionContext backed by fixed thread-pool to JobManagerActor.
Added spark.jobserver.max-dispatcher-per-context (default 2) to reference.conf; for access from both main and test code.
Removed all implicit ExecutionContext references in JobManagerActor, replaced them with explicit once.

### DIFF
--- a/job-server/src/spark.jobserver/JobManagerActor.scala
+++ b/job-server/src/spark.jobserver/JobManagerActor.scala
@@ -1,5 +1,6 @@
 package spark.jobserver
 
+import java.util.concurrent.Executors._
 import akka.actor.{ActorRef, Props, PoisonPill}
 import com.typesafe.config.Config
 import java.net.{URI, URL}
@@ -7,7 +8,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import ooyala.common.akka.InstrumentedActor
 import org.apache.spark.{ SparkEnv, SparkContext }
 import org.joda.time.DateTime
-import scala.concurrent.Future
+import scala.concurrent.{ Future, ExecutionContext }
 import scala.util.{Failure, Success, Try}
 import spark.jobserver.ContextSupervisor.StopContext
 import spark.jobserver.io.{ JobDAO, JobInfo, JarInfo }
@@ -66,15 +67,15 @@ class JobManagerActor(dao: JobDAO,
   import JobManagerActor._
   import scala.util.control.Breaks._
   import collection.JavaConverters._
-  import context.dispatcher       // for futures to work
 
   val config = context.system.settings.config
+  private val maxRunningJobs = SparkJobUtils.getMaxRunningJobs(config)
+  val executionContext = ExecutionContext.fromExecutorService(newFixedThreadPool(maxRunningJobs))
 
   var jobContext: ContextLike = _
   var sparkEnv: SparkEnv = _
   protected var rddManagerActor: ActorRef = _
 
-  private val maxRunningJobs = SparkJobUtils.getMaxRunningJobs(config)
   private val currentRunningJobs = new AtomicInteger(0)
 
   // When the job cache retrieves a jar from the DAO, it also adds it to the SparkContext for distribution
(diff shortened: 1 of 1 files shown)


job-server/src/main/resources/reference.conf
```diff
@@ -0,0 +1,6 @@
+spark {
+  jobserver {
+    # Maximum number of dispatcher threads for each Spark context
+    max-dispatchers-per-context = 2
```
velvia: Two comments:
1) I think the name `max-threads-per-context` is better description, at least I think of dispatchers as like ExecutionContext.
2) 2 seems really low, and means that by default a user can only have 2 jobs in parallel.  How about a default of 4, at least?

job-server/src/spark.jobserver/JobManagerActor.scala
```diff
@@ -7,6 +8,7 @@ import java.util.concurrent.atomic.AtomicInteger
 import ooyala.common.akka.InstrumentedActor
 import org.apache.spark.{ SparkEnv, SparkContext }
 import org.joda.time.DateTime
+import concurrent.ExecutionContext
```
velvia: scala.concurrent.ExecutionContext

velvia:
462cfa3 looks good to me, except for minor comments.

job-server/src/main/resources/reference.conf
mehmeteking: The reason for both is the same; Spark has its own threads for running tasks and we only use these threads to dispatch them to Spark for running. Hence the name "dispatcher" and low number of threads.

A question we may ask is; since there are two threads but not a unique one, how will Spark react when we call setJobGroup. Then, if our threads are A, and B and the invocation is carried out by A; then only and all jobs dispatched by A would be given the specific id. We should perhaps reduce this to a single thread executor. What do you think?

velvia: Ekin,

I want to make sure there is no misunderstanding here.  The job server does
not simply allow running one "Spark job" in each Spark context, which
traditionally runs single-threaded.  Instead, each spark context, which is
mapped to a single JobManagerActor, may allow many "jobs" to be executed in
parallel, each one in their own thread.   For example, a user creates a
context in job server, which spins up a new JobManagerActor.  This context
contains RDDs to be queried.  Each query is a new Job Server job, executed
against this context.  Using the fair scheduler, we could easily have 16
threads running 16 separate Futures, each one containing a separate query
"job".

Job Server terminology: a "job" is not a single SparkContext, but rather a
single Future tied to a single thread.

As far as setJobGroup, it is the intention for each thread/Future to get
its own Jobgroup, because it maps to each JobServer job.

Hope this clears things up.  A thread pool of 2 or 1 thread would make this
important use case of the job server (multiple jobs within a context) not
work.

You can also see the above because there is a separate Atomic counter which
is meant to count the number of concurrent Futures allowed to run at the
same time, and it defaults to 4 I think.  Plus the design of NamedRdds,
which allows concurrent "jobs" within the same context to safely update
named shared RDDs.

This is the reason why originally we wanted to borrow the SparkContext's
ActorSystem threadpool, because we could execute multiple jobs within that
threadpool.

On Fri, Apr 3, 2015 at 3:35 AM, Ekin Gecikligun <notifications@github.com>
wrote:

> In job-server/src/main/resources/reference.conf
> <https://github.com/spark-jobserver/spark-jobserver/pull/111#discussion_r27723944>
> :
>
> > @@ -0,0 +1,6 @@
> > +spark {
> > +  jobserver {
> > +    # Maximum number of dispatcher threads for each Spark context
> > +    max-dispatchers-per-context = 2
>
> The reason for both is the same; Spark has its own threads for running
> tasks and we only use these threads to dispatch them to Spark for running.
> Hence the name "dispatcher" and low number of threads.
>
> A question we may ask is; since there are two threads but not a unique
> one, how will Spark react when we call setJobGroup. Then, if our threads
> are A, and B and the invocation is carried out by A; then only and all jobs
> dispatched by A would be given the specific id. We should perhaps reduce
> this to a single thread executor. What do you think?
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/111/files#r27723944>
> .
>



-- 
   --- Toni Morrison

mehmeteking:
Used max-jobs-per-context to determine maximum number of threads for JobManagerActor thread-pool.

velvia:
@mehmeteking thanks, looks good.

PR #131, opened 2015-04-24
Removed external dependency on jquery

johnbateman:
On a system without an internet connection jobserver UI does not work


velvia:
Thanks @johnbateman

PR #142, opened 2015-05-06
Streaming support (First Draft, for discussion)

zeitos:
This is a basic implementation for supporting Streaming Jobs (like we have SQL and Hive)

Since Streaming jobs are expected to run for long periods of time we might want to talk about adding some kind of status reporting, what do you guys think?


nickpoorman:
+1 I'm all for status updates / reporting, for any type of long running jobs.

velvia:
Yes, status updates would be great.  In fact there was a PR for
webhook-based status updates, need to look at what happened to it.

We also might need to think about a different or enhanced database schema
for status updates.  I'd prefer to keep a log of all updates somewhere, but
the current table layout is designed just to retain the latest update.

On Wed, May 6, 2015 at 2:19 PM, Nicholas Poorman <notifications@github.com>
wrote:

> +1 I'm all for status updates / reporting, for any type of long running
> jobs.
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/142#issuecomment-99612890>
> .
>



-- 
   --- Toni Morrison

zeitos:
I was thinking of a simple way to see the last status reported, if we want something more complex then I guess we should allow pluggeable notifications ie: send them to kafka, rabbitmq, flume or store it in hdfs.

I'm going to open an Issue to track this feature, meanwhile: anyone interested in taking a look at this PR?

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+class StreamingContextFactory extends SparkContextFactory{
+  import SparkJobUtils._
+
+  type C = StreamingContext with ContextLike
+
+  def makeContext(config: Config, contextConfig: Config, contextName: String): C = {
+    val conf = configToSparkConf(config, contextConfig, contextName)
+    val interval :Int = {
+      if (contextConfig.hasPath("streaming.batch_interval")) {
```
velvia: The usual paradigm for using the config is to define defaults in reference/application.conf.  That way, `config.getInt("streaming.batch_interval")` is guaranteed to succeed, and always grabs the default value as a default.  Then you don't need the nested if/ etc.

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
```diff
@@ -0,0 +1,39 @@
...
+ * Created by ZeoS on 3/18/15.
+ */
+class StreamingContextFactory extends SparkContextFactory{
+  import SparkJobUtils._
+
+  type C = StreamingContext with ContextLike
+
+  def makeContext(config: Config, contextConfig: Config, contextName: String): C = {
+    val conf = configToSparkConf(config, contextConfig, contextName)
+    val interval :Int = {
```
velvia: (nit) space, should be `interval: Int`

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
```diff
@@ -0,0 +1,39 @@
...
+      if (contextConfig.hasPath("streaming.batch_interval")) {
+        contextConfig.getInt("streaming.batch_interval")
+      } else {
+        if (config.hasPath("streaming.batch_interval")) {
```
velvia: Also, you can programmatically try to get the config from "spark.context-settings" in the job config first, then fall back to the contextConfig, like this:

    val contextCfg = config.getConfig("spark.context-settings").withFallback(contextConfig)
    contextCfg.getInt("streaming.batch_interval")

Maybe the first line should be a method in SparkContextFactory, since everyone will need logic like this.

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
```diff
@@ -0,0 +1,39 @@
...
+          config.getInt("streaming.batch_interval")
+        } else {
+          10
+        }
+      }
+    }
+    new StreamingContext(conf, Seconds(interval)) with ContextLike {
+      def isValidJob(job: SparkJobBase): Boolean = job.isInstanceOf[SparkStramingJob]
+      def stop() { this.sparkContext.stop() }
```
velvia: Don't you also need to call some stop method in the streamingcontext?

velvia:
Looks fine, for the most part, some comments.

Let's release 0.5.1 first, and then work on merging this along with stuff for status notifications.

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
zeitos: Yes I surely do! great catch.

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
zeitos: yes, it's better that way.

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
zeitos: done

job-server/src/spark.jobserver/context/StreamingContextFactory.scala
zeitos: Agree, done

velvia:
BTW, as I said on the google group, I'm going to move all the extra contexts to another package to decrease the dependencies on the main jobserver package....  I'll submit it as a PR so you and others can comment on it.

zeitos:
Looks like the tests used an invalid config: JobSpecBase defined a config with context-factory as a top level property and didn't have a context-settings

fixed that and modified the manual instantiations of JobManagerActor to ue a new config that corresponds with the expected one.

zeitos:
About the dependency: sounds good

zeitos:
@velvia  do you want to take another look since I changed how tests instantiate JobManagerActor ?

velvia:
@zeitos: yeah will have a look in a little bit.  Don't merge this yet, I'll have a PR out later today to move these new contexts to a new subproject, then we should release 0.5.1.

zeitos:
@velvia finally found some time to get back to this, Migrated the the new -extras package

zeitos:
@velvia this is interesting... there's something new in master that made the tests run slowly and I had to increase the akka timeout to 15s
```
https://github.com/zeitos/spark-jobserver/commit/dff006ccf7c1bd4bfb9f4dd57ee2b2851b8fa2ce
https://github.com/zeitos/spark-jobserver/commit/e03a75a90216f05a0e173948afc45806d3e474cc
```

none of those merges seem to have anything that will slow down the tests.

velvia:
@zeitos thanks for all the fixes.  This looks good to me, so going to merge.

PR #150, opened 2015-05-13
Split Hive and SQLContext and jobs to new job-server-extras project

velvia:
The goal is to remove all the extra Hive and other dependencies from the core job-server project to keep it clean.

Note that tests including the SqlHiveJob and SQL jobs all got moved to the extras project, and the extras project produces its own test jar which is separate from the original tests jar.

Only change for users would be they would need to pull down job-server-extras JAR instead of the original.

Goal is to merge this before releasing 0.5.1.


velvia:
@zeitos @nemccarthy @ajmerlob have a look.

velvia:
I'm going to update the README before merge btw.

job-server/test/spark.jobserver/TestJarFinder.scala
```diff
@@ -1,11 +1,24 @@
...
+import java.nio.file.Paths
+
 trait TestJarFinder {
   val versionRegex = """(\d\.\d+).*""".r
   val version = scala.util.Properties.versionNumberString match { case versionRegex(d) => d }
-  val testJarDir = "../job-server-tests/target/scala-" + version + "/"
-  val testJar = {
-    val allJars = new java.io.File(testJarDir).listFiles.toSeq.filter { file =>
+
+  def testJarDir: String = "job-server-tests/target/scala-" + version + "/"
```
zeitos: I have a fear that someone might forget to add the version in a future impl of TesJarFinder, ExtrasTestJarFinder does the same thing.

job-server/test/spark.jobserver/TestJarFinder.scala
```diff
@@ -1,11 +1,24 @@
...
+
+  def baseDir: String = {
```
zeitos: val?

zeitos:
I made some minor comments, there are failing tests though.

velvia:
@zeitos thanks, addressed your CRs and increased test timeout.  Frustrating thing is that the tests always passes locally for me.

velvia:
Actually I realized another ensuing complication is that a simple `sbt reStart` of the base job-server project won't include the SQLContext, HiveContext etc., and the same for assemblies.

The simplest fix is to generate and pull the assembly from job-server-extras, and redirect folks to do "sbt reStart" from the job-server-extras project.   A cleaner fix would be to move job-server to a job-server-core project, and have job-server be an aggregate project that pulls in extras.  Or, I can see if I can make the root project do the assembly, which is probably the simplest of all.

velvia:
@zeitos and others:  I've rigged it so that both job-server and job-server-extras projects can do `reStart` and `assembly`.  If users do the old thing of `reStart` from just job-server, then they don't get Hive and SQLContexts.  The package and deploy scripts by default grab the assembly from extras with everything.  Updated the docs.

Once this is merged then I can release 0.5.1.

velvia:
If no more comments, I'll merge this by EOD and release 0.5.1.

zeitos:
Looks good to me

PR #157, opened 2015-05-26
Add web service config

janstenpickle:
- passes through application config to the spray web service
- allows job server configuration to be used by spray

This allows the spray instance for the API to be configured using the job server configuration. This will be useful for modifying parsing configuration for tweaking chunk size and max request size. Full docs on spray config can be found here: http://spray.io/documentation/1.2.2/spray-can/configuration/.

### DIFF
--- a/akka-app/src/ooyala.common.akka/web/WebService.scala
+++ b/akka-app/src/ooyala.common.akka/web/WebService.scala
@@ -1,6 +1,8 @@
 package ooyala.common.akka.web
 
 import akka.actor.ActorSystem
+import com.typesafe.config.Config
+import spray.can.server.ServerSettings
 import spray.routing.{Route, SimpleRoutingApp}
 
 /**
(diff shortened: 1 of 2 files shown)


velvia:
@janstenpickle thanks.  Changes look fine, would you mind updating the PR to include what issues or specific problems this addresses?   Be happy to merge after that.

janstenpickle:
Sure, will do :smile:

janstenpickle:
Just discovered that this knocks out the spray default configuration. I'll close for now and re-raise when I have it working properly! Will see if I can add a test around this too...

velvia:
Should be easy to fix, because sprays default config is in reference.conf and ConfigFactory has a method to retrieve it. 

-Evan

janstenpickle:
Just spent ages looking through the spray code... this isn't necessary at all! It can create the config from the actor system config, my problem was that I was changing the parser config, in the docs it has it under `spray.can.parsing...` in the code it's looking under `spray.can.server.parsing...`. DOH.

velvia:
Hah!  Glad you figured it out  :)

On Wed, May 27, 2015 at 7:40 AM, Chris Jansen <notifications@github.com>
wrote:

> Just spent ages looking through the spray code... this isn't necessary at
> all! It can create the config from the actor system config, my problem was
> that I was changing the parser config, in the docs it has it under
> spray.can.parsing... in the code it's looking under
> spray.can.server.parsing.... DOH.
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/157#issuecomment-105936194>
> .
>



-- 
   --- Toni Morrison

PR #165, opened 2015-06-05
Unhardcode JDBC profile

matthew-sharp:


job-server/src/spark.jobserver/io/JobSqlDAO.scala
```diff
@@ -5,11 +5,20 @@ import java.io.{FileOutputStream, BufferedOutputStream, File}
...
 import org.joda.time.DateTime
 import org.slf4j.LoggerFactory
+import scala.slick.driver.JdbcProfile
 import scala.slick.jdbc.meta.MTable
-
+import scala.reflect.runtime.universe
 
 class JobSqlDAO(config: Config) extends JobDAO {
-  import scala.slick.driver.H2Driver.simple._
+  val slickDriverClass = getOrElse(config.getString("spark.jobserver.sqldao.slick-driver"),
```
velvia: I know we have getOrElse, but I think the pattern of putting defaults in `reference.conf` is more standard now and we should follow it.

job-server/src/spark.jobserver/io/JobSqlDAO.scala
```diff
@@ -5,11 +5,20 @@ import java.io.{FileOutputStream, BufferedOutputStream, File}
...
+    "scala.slick.driver.H2Driver")
+  val jdbcDriverClass = getOrElse(config.getString("spark.jobserver.sqldao.jdbc-driver"),
+    "org.h2.Driver")
+
+  val runtimeMirror = universe.runtimeMirror(getClass.getClassLoader)
+  val profileModule = runtimeMirror.staticModule(slickDriverClass)
```
velvia: Nice!  Is this only for objects?

velvia:
@matthew-sharp Looks good mostly except for minor comment about moving defaults to reference.conf, and the test failure.  I'll try to restart it.

velvia:
Ok, I pushed changes to the git branch, but somehow it's not showing up.   So don't merge this, I'll merge it manually.  Made some config changes and got tests to pass.

PR #168, opened 2015-06-10
make logs in spark-submit write to specified log4j config

addisonj:
This fixes logging in the using spark-submit instead of class, see #164


velvia:
@addisonj sweet thanks!   Does this fix the driver logging as well?  (I'll try the other option for driver logging)

PR #174, opened 2015-06-18
Configurable driver memory

acidghost:
@velvia Here it is :smiley:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,6 @@ test-reports/
 # ignore deployment configs
 config/*.conf
 config/*.sh
+job-server/config/*.conf
+job-server/config/*.sh
 metastore_db/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -11,4 +11,6 @@ test-reports/
 # ignore deployment configs
 config/*.conf
 config/*.sh
+job-server/config/*.conf
```
velvia: Nice!  Should have had that ages ago.

job-server/config/local.sh.template
```diff
@@ -10,6 +10,8 @@ APP_GROUP=spark
 INSTALL_DIR=/home/spark/job-server
 LOG_DIR=/var/log/job-server
 PIDFILE=spark-jobserver.pid
+DRIVER_MEMORY=1G
+SPARK_VERSION=1.3.1
```
velvia: Where is SPARK_VERSION used again?

acidghost: Here https://github.com/spark-jobserver/spark-jobserver/blob/master/project/Dependencies.scala#L30

velvia: @acidghost ah right.  Thanks!  Going to merge.

PR #177, opened 2015-06-21
Make Streaming Context more configurable & Fix Tests

zeitos:
This Thread.sleep should be replaced with some proper status message whenever we have that feature.
Something that would indicate that a streaming batch has been processed

### DIFF
--- a/doc/contexts.md
+++ b/doc/contexts.md
@@ -39,3 +39,10 @@ This can be done easily by extending the `SparkContextFactory` trait, like `SQLC
 ## Jars
 
 If you wish to use the `SQLContext` or `HiveContext`, be sure to pull down the job-server-extras package.
+
+# StreamingContext
+job-server-extras provides a context to run Spark Streaming jobs, there are a couple of configurations you can change in job-server's .conf file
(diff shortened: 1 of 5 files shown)


doc/contexts.md
```diff
@@ -38,4 +38,11 @@ This can be done easily by extending the `SparkContextFactory` trait, like `SQLC
 
 ## Jars
 
-If you wish to use the `SQLContext` or `HiveContext`, be sure to pull down the job-server-extras package.
\ No newline at end of file
+If you wish to use the `SQLContext` or `HiveContext`, be sure to pull down the job-server-extras package.
+
+# StreamingContext
+job-server-extras provides a context to run Spark Streaming jobs, there are a couple of configurations you can change in job-server's .conf file
+streaming.batch_interval: the streaming batch in millis
```
velvia: We might need to reformat this for Markdown using lists.  Can fix it later after merge.

velvia:
Thanks @zeitos.  Going to merge it so tests stop being broken, and will fix formatting and other things afterwards.

PR #187, opened 2015-07-01
spark dependency default 1.4.0

marky-mark:


velvia:
Not sure why this failed, but going to upgrade to 1.4.0 by default very soon.

PR #194, opened 2015-07-27
Docker container for SJS

velvia:
This adds a Docker container for SJS that can be built with "docker" command.  The intention is to publish containers to DockerHub with every release to make it easy for others to try SJS and form a base from which to deploy.  See the included docker.md for more details.

- Easy to change SPARK_MASTER, or follow standard procedure to override the included config
- Database writes to a Docker volume for persistence between container lifecycles
- Logs writes to a Docker volume for ease of inspection

Right now, it uses a Spark container as an underlying base.  Mesos is not included.  The base starts up a bunch of namenode stuff which is not necessary.


PierreVincent:
Looks great! There is no image pushed to Docker Hub yet is there?

Is this meant to be used in production or more like a way to get started quickly to play around with the Job server?

velvia:
@PierreVincent  I'm pushing the image now  :)

As is, the container is for quickly getting started, staging environments.  With some simple config tweaks, ie using MySQL or PG as a database, it could be used in production.

hectcastro:
I had not seen this pull request until after going down the path of containerizing SJS for a [work prototype](https://github.com/azavea/docker-spark-jobserver). There are a few things we did differently that I'll try to list and justify below:

- Instead of using environment variables, we made a [`docker-entrypoint.sh`](https://github.com/azavea/docker-spark-jobserver/blob/0.1.0/bin/docker-entrypoint.sh) (this appears to be a convention among the official Docker images) script that is setup as the `ENTRYPOINT`. This means that we can pass arbitrary command line arguments to the underlying `spark-submit` call via `docker run`.
- By default we copy in base configurations for SJS and log4j, but those can be overridden with volumes or inheriting from this image and adding `COPY` instructions.
- The [default log4j configuration](https://github.com/azavea/docker-spark-jobserver/blob/0.1.0/etc/log4j.properties) is setup to log to the console so that the user can leverage Docker's log driver support to centralize logs.
- `java:7` is the base image, which means we handle installing Scala, `sbt`, and Spark. Right now the `Dockerfile` for SJS has all of those instructions, but we plan to separate them out into several Docker images that inherit from each other.
- Our image is built and hosted by [Quay.io](https://quay.io/repository/azavea/spark-jobserver) vs. Docker Hub.
- As of right now, we plan to version the container separate of the SJS version installed within it. This is more for us to easily determine when bits inside the image change. At some point we may move away from `java:7` as the base image to make this more complete.
- Most of the `docker` commands are wrapped in a `Makefile`. If I knew `sbt` better, I may have gone that route, but it also seems like `make` is simple enough for `sbt` and non-`sbt` people to grok.

All of that said, I like your approach to handle persistence by default with H2 and the volume export. I also hadn't seen the `sbt` Docker plugin. I'll have to look into that.

velvia:
@hectcastro Thanks for the points.

* I think with this current setup you can still pass `spark-submit` args after docker run.  It is true though that SJS memory does not need to be an env var, you could just pass `--driver-memory` as an arg, and I'll document that.
* Agreed with general idea of inheriting containers.  Thus JOBSERVER_CONFIG is not really necessary, just overwrite /app/docker.conf
* Logging.  Do you think the built in logs command is really better?  I think for SJS and Spark which has a ton of logs, just replaying logs is clumsier; I'd rather be able to access the logs using tools like less and vim and be able to search.  However, it does seem console output is the standard pattern, so maybe we should stick with that.
* I'll have to look into quay.io, thanks for the pointer!

hectcastro:
> Logging. Do you think the built in logs command is really better? I think for SJS and Spark which has a ton of logs, just replaying logs is clumsier; I'd rather be able to access the logs using tools like less and vim and be able to search. However, it does seem console output is the standard pattern, so maybe we should stick with that.

Not necessarily `docker logs`, but in preparation for wider support of [logging drivers](https://docs.docker.com/reference/run/#logging-drivers-log-driver). Going to start with the `syslog` driver, then work on routing things to some central aggregation server or service from there.

velvia:
@hectcastro so I tried directing the logging output to stdout, per your reocmmendations. Only hitch is that now the timestamp and log level and thread are missing from the log4j logs output.  Very weird.  Have you seen this?

velvia:
@hectcastro nevermind, figured it out

PR #199, opened 2015-08-04
When submit a job, if sparkcontext not alive, send ContextStopped message to LocalContextSupervisorActor to do the sc clean up job.

litao-buptsse:
When submit a job, if sparkcontext not alive (e.g. yarn application -kill $appId), the context meta is still exist in jobserver. So when I want to recreate a context, the jobserver tell me the context is existed.

So I make this PR, When submit a job, if sparkcontext not alive, send ContextStopped message to LocalContextSupervisorActor to do the sc clean up job.

### DIFF
--- a/job-server/src/spark.jobserver/LocalContextSupervisorActor.scala
+++ b/job-server/src/spark.jobserver/LocalContextSupervisorActor.scala
@@ -6,7 +6,7 @@ import akka.util.Timeout
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigFactory
 import ooyala.common.akka.InstrumentedActor
-import spark.jobserver.JobManagerActor.IsSparkContextAlive
+import spark.jobserver.JobManagerActor.{SparkContextDead, SparkContextAlive, SparkContextStatus}
 import spark.jobserver.io.JobDAO
 import spark.jobserver.util.SparkJobUtils
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


job-server/src/spark.jobserver/LocalContextSupervisorActor.scala
```diff
@@ -126,7 +128,16 @@ class LocalContextSupervisorActor(dao: JobDAO) extends InstrumentedActor {
 
     case GetContext(name) =>
       if (contexts contains name) {
-        sender ! (contexts(name), resultActors(name))
+        val future = (contexts(name) ? SparkContextStatus) (contextTimeout.seconds)
+        Await.result(future, contextTimeout.seconds) match {
```
velvia: Instead of using `Await.result`, which blocks the thread, it's better to map the future:
```
  future.collect {
    case SparkContextAlive =>   sender ! (contexts(name), resultActors(name))
    case SparkContextDead =>
      self ! StopContext(name)
      sender ! NoSuchContext
  }
```

Also:  note that  { } are not needed after a case, and it would be really good to log when the context was found to be dead.

velvia:
@litao-buptsse Thanks again.  Just have a comment about making sure to use nonblocking futures API instead of blocking.  LMK if you can make the change or you'd like me to do it.

litao-buptsse:
@velvia I add a new commit litao-buptsse@363040f to remove the { } after case and log when the context was found to be dead.

But when I change to the nonblocking futures API, job submit timeout exception happens, and I haven't  figure why. So I just 
just make a simple code style refactor commit.

velvia:
@litao-buptsse Try one thing.  With the nonblocking future API, before giving up.  I think the reason why job submit timeout happens is because we had `sender` in the closure inside collect/map, which is a no-no.  Sender is volatile, so we need to capture it before, like this:
```
  val originator = sender   // originator will not change, unlike sender
  future.collect {
    case SparkContextAlive =>   originator ! (contexts(name), resultActors(name))
    case SparkContextDead =>
      self ! StopContext(name)
      originator ! NoSuchContext
  }

```

litao-buptsse:
@velvia Nice! Your solution works fine for me. I add a new commit litao-buptsse@632eb29 following your suggestion.

velvia:
Thanks @litao-buptsse, going to merge.
BTW are you a Lakers fan?  ;-)

litao-buptsse:
@velvia Yeah! I am an old Lakers and Kobe fan. :smile:

PR #202, opened 2015-08-07
Adds Hook For Custom SparkMasterProvider

RussellSpitzer:
Currently the spark.master address passed into the JobServer is used for
all Contexts launched in the future. If the Spark master changes due to
HA Systems the JobServer can no longer make new Contexts. This patch
adds a hook to a new class via spark.master.provider. With this a user
can implement a custom SparkMasterProvider object which will be invoked
on the creation of new SparkConfs. This allows for a dynamic adjustment
of the spark.master property for newley created contexts.

### DIFF
--- a/job-server/src/spark.jobserver/util/SparkMasterProvider.scala
+++ b/job-server/src/spark.jobserver/util/SparkMasterProvider.scala
@@ -30,7 +30,6 @@ object SparkMasterProvider {
 
     try {
       val sparkMasterObjectName = config.getString(SparkMasterProperty)
-      println(s"Using $sparkMasterObjectName to determine Spark Master")
       logger.info(s"Using $sparkMasterObjectName to determine Spark Master")
       val m = ru.runtimeMirror(getClass.getClassLoader)
       val module = m.staticModule(sparkMasterObjectName)
--- a/job-server/src/spark.jobserver/util/SparkMasterProvider.scala
+++ b/job-server/src/spark.jobserver/util/SparkMasterProvider.scala
@@ -1,0 +1,54 @@
+package spark.jobserver.util
+
+import scala.reflect.runtime.{universe => ru}
+import com.typesafe.config.{ConfigException, Config}
+import org.slf4j.LoggerFactory
+
+
+trait SparkMasterProvider {
+
+  /**
+   * Implementing classes will determine what the appropriate SparkMaster is and
+   * return it
+   * @return A Spark Master Address
+   */
+  def getSparkMaster(config: Config): String
+
+}
+
+object SparkMasterProvider {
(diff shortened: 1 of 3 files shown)


job-server/src/spark.jobserver/util/SparkMasterProvider.scala
```diff
@@ -0,0 +1,54 @@
...
+   * Will look for an Object with the name provided in the Config file and return it
+   * or the DefaultSparkMasterProvider if no spark.master.provider was specified
+   * @param config SparkJobserver Config
+   * @return A SparkMasterProvider
+   */
+  def fromConfig(config: Config): SparkMasterProvider = {
+
+    try {
+      val sparkMasterObjectName = config.getString(SparkMasterProperty)
+      println(s"Using $sparkMasterObjectName to determine Spark Master")
```
velvia: Let's not have println's...  actually this should be flagged when you do assembly or test by ScalaStyle...

job-server/src/spark.jobserver/util/SparkMasterProvider.scala
```diff
@@ -0,0 +1,54 @@
...
+      logger.info(s"Using $sparkMasterObjectName to determine Spark Master")
+      val m = ru.runtimeMirror(getClass.getClassLoader)
+      val module = m.staticModule(sparkMasterObjectName)
+      val sparkMasterProviderObject = m.reflectModule(module).instance
+        .asInstanceOf[SparkMasterProvider]
+      sparkMasterProviderObject
+    } catch {
+      case me: ConfigException => DefaultSparkMasterProvider
```
velvia: You could get rid of this line by providing a default in `application.conf`.... I think....

velvia:
LGTM..... thanks @RussellSpitzer !

job-server/src/spark.jobserver/util/SparkMasterProvider.scala
RussellSpitzer: Woops! I'll remove that

job-server/src/spark.jobserver/util/SparkMasterProvider.scala
RussellSpitzer: I didn't want to have to go through the reflection pathway unless necessary, Is it possible to have the default be a SparkMasterProvider Object?

RussellSpitzer:
Removed the errant println and i'm not sure what to do with the Application.conf defaults, Do you prefer having the history squashed?

velvia:
@RussellSpitzer I mean to have the default setting in application.conf, so that `config.getString` cannot throw a ConfigException.  It's not a biggie though.

velvia:
We don't need history squashed, this is good, I'm going to merge it.  Thanks!

PR #209, opened 2015-08-17
Handle DB connection exceptions, log them and exit

noorul:
DB connection related exceptions are not caught currently. Catch them,
log and exit while startup.


velvia:
@noorul thanks!

PR #210, opened 2015-08-18
Add kill-process-tree.sh to server_deploy.sh

jmelching:
The server_deploy.sh script didn't copy up the kill-process-tree.sh script so server_stop.sh failed.


velvia:
@jmelching thanks!

PR #211, opened 2015-08-19
Upgrade to spray-json 1.3.2

v-gerasimov:
Spray Json was upgraded to 1.3.2.
Changelog of 1.3.2: https://github.com/spray/spray-json/blob/master/CHANGELOG


velvia:
@v-gerasimov thanks!

PR #212, opened 2015-08-19
Update to the latest Joda Time

jwalgran:
Attempting to run a job that uses the AWS client raised this exception

```
java.lang.IllegalStateException: Joda-time 2.2 or later version is required, but found version: null
    at com.amazonaws.util.DateUtils.handleException(DateUtils.java:147)
    at com.amazonaws.util.DateUtils.parseRFC822Date(DateUtils.java:195)
    at com.amazonaws.services.s3.internal.ServiceUtils.parseRfc822Date(ServiceUtils.java:78)
    ...
```

Updating the dependency to the latest version of Joda Time avoids the exception and does not break any unit tests.

Some questions related to this change:

1. Is there a better way of handling this conflict? Is there a method by which I can have a job jar with a transitive dependency on a newer version of one of the jars that SJS depends on?

2. Upgrading from 2.1 to 2.8.1 moves through several version. The compatibility section of all the incremental release notes says "Yes" in all cases, but there are some "Yes, except" sections that explain changes in behavior. If the only way to handle the version conflict is to upgrade, is it preferable to only jump from 2.1 to 2.2 rather than the latest stable version?

http://www.joda.org/joda-time/upgradeto220.html
http://www.joda.org/joda-time/upgradeto230.html
http://www.joda.org/joda-time/upgradeto240.html
http://www.joda.org/joda-time/upgradeto250.html
http://www.joda.org/joda-time/upgradeto260.html
http://www.joda.org/joda-time/upgradeto270.html
http://www.joda.org/joda-time/upgradeto280.html
http://www.joda.org/joda-time/upgradeto281.html

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -12,7 +12,7 @@ object Dependencies {
 
   lazy val yodaDeps = Seq(
     "org.joda" % "joda-convert" % "1.2",
-    "joda-time" % "joda-time" % "2.1"
+    "joda-time" % "joda-time" % "2.8.1"
   )
 
   lazy val akkaDeps = Seq(


velvia:
@jwalgran thanks.  Hmm... dealing with conflicts is hard.  You could for example try pushing a newer version of joda without updating SJS by starting server_start.sh with a --package or --jars pointing to the newer joda time.  Hopefully that overrides what comes with Spark/SJS, but it might not ...  but do you want to try that out?  The other way is to bundle the newer version directly into your project using sbt or whatever you use to build.  It's really the same thing.  However, a class that exists already might not get updated correctly due to how the JVM does classloading.

BTW as far as upgrading Joda in SJS I have no issues, so will look at this a bit more.    Joda has very stable APIs so I'm not too worried.

velvia:
@jwalgran BTW do you mind adding Azavea to the README list of job server users?   Thanks  :)

velvia:
@jwalgran what do you think if we just bumped the dep to joda 2.2?  Not particular reason, just want to avoid any possible issues.

jwalgran:
I originally tried to bundle the latest Joda Time with my job jar, but that did not resolve the conflict. The error message reported by the AWS client says that version 2.2 should be good enough, so I will give that a try.

velvia:
If you're OK with it, I will bump the joda time version to 2.2, then close
the PR (or you can do the same)

On Tue, Aug 25, 2015 at 7:55 AM, Justin Walgran <notifications@github.com>
wrote:

> I originally tried to bundle the latest Joda Time with my job jar, but
> that did not resolve the conflict. The error message reported by the AWS
> client says that version 2.2 should be good enough, so I will give that a
> try.
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/212#issuecomment-134612470>
> .
>



-- 
   --- Toni Morrison

jwalgran:
Closed in favor of #223 which just bumps the Joda Time version to 2.2

PR #216, opened 2015-08-21
Support username and password in DB access

ffarozan:
JobSqlDAO to support user and password for security


velvia:
@ffarozan thanks!

PR #217, opened 2015-08-22
Store initial job jars on startup.

lossyrob:
This PR allows the user to supply a `spark.jobserver.job-jar-uris` configuration object that will cause SJS to store jars at the given URIs for the given application names. It behaves as if the first thing you do when SJS starts is post jars to the `localhost:8090/jars/appName` endpoint with a set of local jars.

The idea comes from @hectcastro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ See [Troubleshooting Tips](doc/troubleshooting.md) as well as [Yarn tips](doc/ya
 - Frontline Solvers
 - Aruba Networks
 - [Zed Worldwide](http://www.zed.com)
+- [Azavea](http://azavea.com)
 
 ## Features
 
@@ -85,7 +86,7 @@ Then go ahead and start the job server using the instructions above.
 
 Let's upload the jar:
 
-    curl --data-binary @job-server-tests/target/job-server-tests-$VER.jar localhost:8090/jars/test
+    curl --data-binary @job-server-tests/target/scala-2.10/job-server-tests-$VER.jar localhost:8090/jars/test
     OK⏎
 
 #### Ad-hoc Mode - Single, Unrelated Jobs (Transient Context)
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -19,7 +19,7 @@ spark {
 
     # To load up job jars on startup, place them here,
     # with the app name as the key and the path to the jar as the value
-    # job-jar-uris {
+    # job-jar-paths {
     #   test = ../job-server-tests/target/scala-2.10/job-server-tests_2.10-0.5.3-SNAPSHOT.jar
     # }
 
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -17,6 +17,12 @@ spark {
       rootdir = /tmp/spark-jobserver/filedao/data
     }
 
+    # To load up job jars on startup, place them here,
+    # with the app name as the key and the path to the jar as the value
+    # job-jar-uris {
+    #   test = ../job-server-tests/target/scala-2.10/job-server-tests_2.10-0.5.3-SNAPSHOT.jar
+    # }
+
     sqldao {
       # Slick database driver, full classpath
       slick-driver = scala.slick.driver.H2Driver
--- a/job-server/src/spark.jobserver/JarManager.scala
+++ b/job-server/src/spark.jobserver/JarManager.scala
@@ -5,9 +5,12 @@ import spark.jobserver.io.JobDAO
 import spark.jobserver.util.JarUtils
 import org.joda.time.DateTime
 
+import java.nio.file.{Files, Paths}
+
 // Messages to JarManager actor
 case class StoreJar(appName: String, jarBytes: Array[Byte])
 case object ListJars
+case class StoreInitialJars(initialJars: Map[String, String])
 
 // Responses
 case object InvalidJar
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -85,7 +86,7 @@ Then go ahead and start the job server using the instructions above.
 
 Let's upload the jar:
 
-    curl --data-binary @job-server-tests/target/job-server-tests-$VER.jar localhost:8090/jars/test
+    curl --data-binary @job-server-tests/target/scala-2.10/job-server-tests-$VER.jar localhost:8090/jars/test
```
lossyrob: This was required for my run of the test, is this a difference in `sbt` versions or a typo in the README?

velvia: @lossyrob thanks for the PR!   Yeah it's probably an oversight, a while back the path changed.

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -66,6 +75,44 @@ object JobServer {
...
+
+      val initialJars =
+        initialJarsConfig
+          .asScala
+          .map { case (key, value) => (key, value.unwrapped.toString) }
+          .toMap
+
+      // Ensure that the jars exist
+      for(jarPath <- initialJars.values) {
+        val f = new java.io.File(jarPath)
```
velvia: So, the code here assumes that each of the strings in `job-jar-uris` is actually a local path, is that right?   I think this is an ok place to start, but maybe we should rename the config path as `job-jar-paths`?

job-server/src/main/resources/application.conf
```diff
@@ -17,6 +17,12 @@ spark {
       rootdir = /tmp/spark-jobserver/filedao/data
     }
 
+    # To load up job jars on startup, place them here,
+    # with the app name as the key and the path to the jar as the value
+    # job-jar-uris {
+    #   test = ../job-server-tests/target/scala-2.10/job-server-tests_2.10-0.5.3-SNAPSHOT.jar
```
noorul: Can this be any URI? For example we store job jar in an HTTP server. Will this patch help us to put the url here and upload the jar?

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -66,6 +75,44 @@ object JobServer {
...
+        if(!f.exists) {
+          val msg =
+            if(f.isAbsolute) {
+              s"Inital Jar File $jarPath does not exist"
+            } else {
+              s"Inital Jar File $jarPath (${f.getAbsolutePath}) does not exist"
+            }
+
+          throw new java.io.IOException(msg)
```
noorul: Does this means that it is all or none?

For exampl, say I have four apps and corresponding paths and somehow one of the path is failing. Does this mean none them will be uploaded?

job-server/src/spark.jobserver/JobServer.scala
lossyrob: That makes sense!

job-server/src/main/resources/application.conf
lossyrob: This currently only supports local paths. Adding support for other initial jar locations could be supported, though...is there a distinct need for that currently, or should we start with just local JARs?

job-server/src/spark.jobserver/JobServer.scala
lossyrob: If there is an error with the initial JARs, then SJS won't start, and will inform you about the failing JAR. That behavior was by design, do you think another behavior would be better?

noorul: makes sense

job-server/src/spark.jobserver/JobServer.scala
noorul: Do you think we should add test cases too?

job-server/src/spark.jobserver/JarManager.scala
```diff
@@ -5,9 +5,12 @@ import spark.jobserver.io.JobDAO
 import spark.jobserver.util.JarUtils
 import org.joda.time.DateTime
 
+import java.nio.file.{Files, Paths}
+
 // Messages to JarManager actor
 case class StoreJar(appName: String, jarBytes: Array[Byte])
 case object ListJars
+case class StoreInitialJars(initialJars: Map[String, String])
```
velvia: Very minor nit: from JarManager's perspective it doesn't care if they are initial or not, could call the message `StoreJars`.

job-server/src/spark.jobserver/JarManager.scala
```diff
@@ -18,16 +21,42 @@ case object JarStored
...
+              }
+            } catch {
+              case e: Exception =>
+                  logger.error(e.getMessage)
+                  false
+            }
+          }
+        }
+
+      sender ! (if(success) { JarStored } else { InvalidJar })
```
velvia: minor nit: space after if

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -66,6 +75,45 @@ object JobServer {
...
+          .map { case (key, value) => (key, value.unwrapped.toString) }
+          .toMap
+
+      // Ensure that the jars exist
+      for(jarPath <- initialJars.values) {
+        val f = new java.io.File(jarPath)
+        if(!f.exists) {
+          val msg =
+            if(f.isAbsolute) {
+              s"Inital Jar File $jarPath does not exist"
```
velvia: spelling: Initial

job-server/src/spark.jobserver/JobServer.scala
velvia: I think we can leave it, JobServer.scala is a bit hard to test since it's the outer app.

velvia:
@lossyrob Thanks again;  except for very minor comments I think this is ready for merge.

job-server/src/spark.jobserver/JarManager.scala
noorul: I agree, but I have one question, how do we differentiate both these classes. I think we should somehow document it.

job-server/src/spark.jobserver/JarManager.scala
noorul: Don't we have this enabled in scalastyle?

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -66,6 +75,45 @@ object JobServer {
...
+          .asScala
+          .map { case (key, value) => (key, value.unwrapped.toString) }
+          .toMap
+
+      // Ensure that the jars exist
+      for(jarPath <- initialJars.values) {
+        val f = new java.io.File(jarPath)
+        if(!f.exists) {
+          val msg =
+            if(f.isAbsolute) {
```
noorul: space after if?

job-server/src/spark.jobserver/JarManager.scala
lossyrob: Makes sense, however I wanted to differentiate the difference between the `StoreJar` that took the `jarBytes` and this message, which actually provides the path to be read. Do you think this deserves differentiation in naming? If not I can just change to `StoreJars`.

I'll also add some doc strings on the messages, thanks @noorul

lossyrob:
Sorry I've been out, I'll update this PR with the comments soon.

job-server/src/spark.jobserver/JarManager.scala
lossyrob: How about `StoreLocalJars`?

velvia: @lossyrob `StoreLocalJars` sounds perfect!

lossyrob:
This should be ready for merge now.

noorul:
LGTM

velvia:
@lossyrob Thanks again, everything looks good, going to merge.

PR #226, opened 2015-08-26
Fix typo in SparkStreamingJob

zeitos:
This would break the code of anyone using SparkStramingJob... so should we break that or I can add another trait called SparkStramingJob that extends SparkStreamingJob, and mark it as deprecated and we can delete that one on another release.


noorul:
Will merging this get into a release? I think stable is 0.5.2.

noorul:
LGTM

noorul:
Should we add a note?

velvia:
Yeah @fotero if you could add a note that'd be great.  Welcome back btw!

-Evan

velvia:
Thinking of doing a small point release, a lot of good minor patches recently.  Or we can wait for the SSL changes, but those are much bigger.  

-Evan

zeitos:
https://github.com/spark-jobserver/spark-jobserver/pull/228

PR #227, opened 2015-08-26
SSL secured connection support

koettert:
Hi everyone,
we at KNIME.org are working on a Spark integration into our tool which uses the spark-jobserver. We plan to contribute several new functions to it and the first will be secured connection between client and server based on ssl encryption which is this PR.
We are also currently working on a LDAP authentication based on Apache Shiro (https://github.com/koettert/spark-jobserver/tree/ldap). We will file an additional PR for this later.
Thanks for all your great work!
Bye,
Tobias


noorul:
@koettert Thank you for the patch. Do you mind adding your company name in users list in README.md?

koettert:
@noorul thanks a lot for the valuable feedback. We will have a look into your suggestions and adapt the code accordingly. We will also add KNIME to the company list.

TimMaltGermany:
@noorul thanks, MySSLConfig is indeed an accident. It is not needed. 
If I understand your comment correctly, then we should remove the PasswordRequester as well and only use passwords from the config, correct?

We can move the other hard-coded stuff into the config as well.

noorul:
Yes, that is what I said.

TimMaltGermany:
Pull request is ready for another review. We think that we addressed all comments. We also added a simple test that tests the ssl config factory. 
Thanks.

velvia:
@noorul @TimMaltGermany I think it's fine to have the README change as part of the PR.  Let's not make life unnecessarily hard  :)

README.md
```diff
@@ -234,6 +235,25 @@ def validate(sc:SparkContext, config: Contig): SparkJobValidation = {
 }
 ```
 
+### HTTPS / SSL Configuration
+To activate ssl communication, set these flags in your application.conf file (Section 'spray.can.server'):
```
velvia: Leave a space between ### and the first comment

README.md
```diff
@@ -234,6 +235,25 @@ def validate(sc:SparkContext, config: Contig): SparkJobValidation = {
...
+```
+  ssl-encryption = on
+  keystore = "~/sjs.jks"
```
velvia: Does the `~/` actually work?   It's usually better to stick with absolute paths or at least use environment vars, I think this syntax works `${HOME}"/sjs.ks"`

README.md
```diff
@@ -234,6 +235,25 @@ def validate(sc:SparkContext, config: Contig): SparkJobValidation = {
...
+  keystorePW = "changeit"
+```
+
+You will need a keystore that contains the server certificate. The bare minimum is achieved with this command which creates a self-signed certificate:
```
velvia: Thanks for these instructions.  Awesome.

velvia:
@koettert  @TimMaltGermany Thanks very much for this PR, it will be really valuable for all SJS users.  Reviewed up to a202bf9.  :+1:  for the most part, just really minor comments. 

Does assembly and package still work?  

If no more comments, I think we can merge.

TimMaltGermany:
Great, thanks for merging! We had a minor time zone issue ... I just read your last comments. 
Should I still make the last changes (in a new branch)? Missing are: 
- rename SslContextFactory to SSLContextFactory
- change default value in Readme.md and application.conf of ~/sjs.jks to /some/path/sjs.jks

Cheers,
Tim

noorul:
I raised a PR for renaming SslContextFactory as SSLContextFactory

Thanks and Regards
Noorul

PR #231, opened 2015-08-28
Rename SslContextFactory as SSLContextFactory

noorul:


TimMaltGermany:
Thanks.

noorul:
No problems

TimMaltGermany:
I just noticed a minor glitch. I log the keystore type instead of the encryption type in line 35. Could you please change that as well?
replace:
logger.info(keystoreType + " encryption activated.")
with:
logger.info(config.getString("encryptionType") + " encryption activated.")
(or create a local variable if you prefer)

Thanks, Tim

noorul:
Done!

velvia:
@noorul thanks!

PR #232, opened 2015-08-28
Integration of Apache Shiro for user authentication

koettert:
Implementation of user authentication via LDAP that utilizes the Apache Shiro framework.
Currently, two authentication mechanisms are supported:
1) Basic LDAP authentication with username / password without group membership checking
2) LDAP authentication with username / password and additional group membership checks based on the 'member' attribute of LDAP groups.

It should be relatively straight forward to add further authentication realms, but we chose to restrict the PR to LDAP only for now.

This implementation does not use session tokens and re-authenticates users with every request. Shiro does have a caching mechanism which is activated in the shiro.ini file.

Currently all but two routes require authentication. We chose not to require authentication for "otherRoutes" and "healthzRoutes" in WebApi (these are currently marked with "TODO - authentication required"). Please let us know if that is OK or if another combination of authentications is preferred.

There is a somewhat ugly implementation detail that we would like get your feedback on: The LdapGroupRealm class needs to know the LDAP search base. We currently extract that information from the config in the method JobServer.setSystemPropertiesFromConfig and put it into the java properties so that we can extract it in LdapGroupRealm again from these properties. Is there a better way to do this?

Thanks,
the KNIME team.


noorul:
I think authentication backend should be pluggable.

velvia:
Many thanks guys.  

@koettert re: LdapGroupRealm config.  Is there a config class of some kind?  I can't imagine it only takes java properties.   

This might not necessarily help, but Typesafe Config library actually merges values from java system properties by default.

README.md
```diff
@@ -248,12 +249,53 @@ You will need a keystore that contains the server certificate. The bare minimum
...
+    # an empty list of allowed groups also effectively disables group checking
+	allowedGroups = ["cn=group1,ou=groups", "cn=group2,ou=groups"]
+    # ldap search base
+    searchBase = "dc=xxx,dc=org"
+}
+```
+
+Here is an example of a simple curl command that authenticates a user and uses ssl:
+```
+curl -k --basic --user 'user:pw' https://localhost:8090/contexts
```
velvia: Not a biggie, but I prefer to directly pass in the basic authentication token header using -H so that plain password won't be exposed.

job-server/src/main/resources/shiro.ini
```diff
@@ -0,0 +1,12 @@
+
+# use this for basic ldap authorization, without group checking
+# activeDirectoryRealm = org.apache.shiro.realm.ldap.JndiLdapRealm
+# use this for checking group membership of users based on the 'member' attribute of the groups:
+activeDirectoryRealm = spark.jobserver.auth.LdapGroupRealm
+activeDirectoryRealm.contextFactory.environment[java.naming.security.credentials] = password
+activeDirectoryRealm.contextFactory.url = ldap://xxx.yyy..org:389
```
velvia: So this file won't work as is, right?  Instead of bundling it in src/main/resources, I'd prefer to stick a template in the `config/` subdir, just like we have for the other configs like logging, and then make it part of the deploy process + Docker container.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -51,6 +53,17 @@ class WebApi(system: ActorSystem,
 
   val myRoutes = jarRoutes ~ contextRoutes ~ jobRoutes ~ healthzRoutes ~ otherRoutes
 
+  lazy val authenticator: AuthMagnet[AuthInfo] = {
+    if (config.hasPath("shiro.authentication") &&
+      config.getBoolean("shiro.authentication")) {
+      logger.info("Using authentication.")
+      asShiroAuthenticator(config)
+    } else {
+      logger.trace("No authentication.")
```
velvia: I think it's OK to log this at info, so it's obvious.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -79,33 +92,36 @@ class WebApi(system: ActorSystem,
...
-          entity(as[Array[Byte]]) { jarBytes =>
-            val future = jarManager ? StoreJar(appName, jarBytes)
-            respondWithMediaType(MediaTypes.`application/json`) { ctx =>
-              future.map {
-                case JarStored  => ctx.complete(StatusCodes.OK)
-                case InvalidJar => badRequest(ctx, "Jar is not of the right format")
-              }.recover {
-                case e: Exception => ctx.complete(500, errMap(e, "ERROR"))
+    // user authentication
+    authenticate(authenticator) { authInfo =>
```
velvia: `authInfo` isn't used anywhere right?

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -171,6 +191,7 @@ class WebApi(system: ActorSystem,
    *    GET /healthz              - return OK or error message
    */
   def healthzRoutes: Route = pathPrefix("healthz") {
+    //TODO - authentication required?
```
velvia: I agree `/healthz` and other routes should not need auth.

velvia:
@koettert would you guys kindly add more detail to README and this PR description about the following:
- Are permissions global (ie once authenticated, a user can access any of the routes), or is it possible to tune permissions per route?  IE, some folks might belong to admin group which can delete contexts, but others can only start jobs.   If specific tuning is possible, definitely good to document. 

If not, would be great to file a ticket with details how this could be added....

koettert:
@velvia Thanks a lot for your valuable feedback. We will have a look at it.

koettert:
@noorul Maybe I do not understand correctly what you mean with the authentication backend should be pluggable, but Apache shiro is plugable. To cite its documentation "Shiro provides out-of-the-box realms for popular data sources like LDAP, Active Directory, and JDBC. If needed, you can also create your own realms to support specific functionality not included in the basic realms."

noorul:
@koettert How do I add other realms, for example Basic Auth? Is this documented as part of this patch?

job-server/src/main/resources/application.conf
```diff
@@ -131,3 +131,19 @@ spray.can.server {
   # Increase this in order to upload bigger job jars
   parsing.max-content-length = 30m
 }
+
+shiro {
+    # activate authentication (and authorization)
+    # authentication = on
+    # absolute path to shiro config file, including file name
+    path.ini = "/some/path/shiro.ini"
```
noorul: Isn't config.path a better name?

job-server/src/main/resources/application.conf
```diff
@@ -131,3 +131,19 @@ spray.can.server {
...
+    # note that ssl-encryption should also be on when authentication is on
```
noorul: Why is this so?

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -26,6 +26,13 @@ import org.slf4j.LoggerFactory
 object JobServer {
   val logger = LoggerFactory.getLogger(getClass)
 
+  private def setSystemPropertiesFromConfig(config: Config) {
```
noorul: I think we should avoid adding code to JobServer.scala as this is difficult to test as @velvia said.

noorul:
Also please add tests, which will help to review the code easily.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -51,6 +53,17 @@ class WebApi(system: ActorSystem,
 
   val myRoutes = jarRoutes ~ contextRoutes ~ jobRoutes ~ healthzRoutes ~ otherRoutes
 
+  lazy val authenticator: AuthMagnet[AuthInfo] = {
+    if (config.hasPath("shiro.authentication") &&
+      config.getBoolean("shiro.authentication")) {
```
zeitos: Can we set the fallback value to false and just do config.getBoolean ?

job-server/src/spark.jobserver/auth/AuthInfo.scala
```diff
@@ -0,0 +1,13 @@
+package spark.jobserver.auth
+
+/**
+ * based on example provided by Mario Camou
+ * at
+ * http://www.tecnoguru.com/blog/2014/07/07/implementing-http-basic-authentication-with-spray/
+ */
+class AuthInfo(val user: User) {
+  def hasPermission(permission: String): Boolean = {
+    // Code to verify whether user has the given permission
```
zeitos: I think this comment comes from the example and here 'true' means that by defaule every user has permission

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
```diff
@@ -0,0 +1,126 @@
...
+ * the 'member' attribute values of each group against the given user.
+ *
+ * @note not all LDAP installations use the member property.... we might have to add
+ *   memberOf matches as well and others
+ *
+ * @author dwk (basics in Java stem from different sources by various authors from stackoverflow and such)
+ */
+class LdapGroupRealm extends JndiLdapRealm {
+
+  private val logger = LoggerFactory.getLogger(getClass)
```
zeitos: side note: I think we should have a trait for this... we have Slf4jLogging but that one extends ActorStack

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
```diff
@@ -0,0 +1,126 @@
...
+    val groupMemberFilter = "(member=*)"
+    logger.trace("Retrieving group memberships.")
+
+    val groupSearchAtts: Array[Object] = Array()
+    val groupAnswer = ldapContext.search(searchBase, groupMemberFilter,
+      groupSearchAtts, searchCtls);
+
+    val members = scala.collection.mutable.Map[String, Set[String]]()
+
+    while (groupAnswer.hasMoreElements()) {
```
zeitos: consider using a more functional groupAnswer.map { case (e) => {} }

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
```diff
@@ -0,0 +1,126 @@
...
+  private def getMembers(sr: SearchResult): Set[String] = {
+    val attrs: Attributes = sr.getAttributes();
+
+    val members = scala.collection.mutable.HashSet[String]()
+    if (attrs != null) {
+      val ae: NamingEnumeration[_ <: Attribute] = attrs.getAll();
+      while (ae.hasMore()) {
+        val attr: Attribute = ae.next();
+
+        if (attr.getID().equals("member")) {
```
zeitos: in scala you can compare strings with ==

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
```diff
@@ -0,0 +1,126 @@
...
+        logger.debug("Checking members of group [" + sr2.getName() + "]")
+      }
+
+      members += sr2.getName() -> getMembers(sr2)
+    }
+    members.toMap
+  }
+
+  private def getMembers(sr: SearchResult): Set[String] = {
+    val attrs: Attributes = sr.getAttributes();
```
zeitos: Can you do a filter,a map and a mkString? Something like
```
import scala.collection.JavaConversions._
 ae.filter(_.getID() =="member").map(e => LdapUtils.getAllAttributeValues(e)).flatten.toSet
```

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
```diff
@@ -0,0 +1,120 @@
...
+      } else {
+        None
+      }
+    }
+
+    def isInAllowedGroupOrNoCheckOnGroups(currentUser: Subject): Boolean = {
+      allowedGroups match {
+        case Some(groups) => {
+          val roles: Array[Boolean] = currentUser.hasRoles(groups)
+          roles.foldLeft(false)((r, isInGroup) => isInGroup || r)
```
zeitos: I think this is the same as this is the same as roles.contains(true)

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
```diff
@@ -0,0 +1,120 @@
...
+
+    //TODO - can we read this information from the config instead?
+    val ldapFactory = new IniSecurityManagerFactory(config.getString("shiro.path.ini"))
+    val sManager = ldapFactory.getInstance()
+    SecurityUtils.setSecurityManager(sManager)
+
+    val allowedGroups: Option[java.util.List[String]] = {
+      if (config.hasPath("shiro.ldap.allowedGroups")) {
+        val g = config.getStringList("shiro.ldap.allowedGroups")
+        if (g.size() > 0) { Some(g) } else { None }
```
zeitos: I think !g.isEmpty is more readable, or even better 
```
import scala.collection.JavaConversions._
if(g.nonempty)
```

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
```diff
@@ -0,0 +1,120 @@
...
+
+      //if (!currentUser.isAuthenticated()) {
+      Future {
+        val UserPass(user, pass) = userPass.get
+        val token = new UsernamePasswordToken(user, pass)
+        try {
+          currentUser.login(token)
+          val fullName = currentUser.getPrincipal().toString
+          if (isInAllowedGroupOrNoCheckOnGroups(currentUser)) {
+            logger.trace("ACCESS GRANTED for user [" + fullName + "]")
```
zeitos: same as before, use the template method so you don't have to check if trace is enabled

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
```diff
@@ -0,0 +1,120 @@
...
+            currentUser.logout()
+            Option(new AuthInfo(new User(fullName)))
+          } else {
+            currentUser.logout()
+            logger.info("ACCESS DENIED (GROUP)")
+            None
+          }
+        } catch {
+          case uae: UnknownAccountException => {
+            logger.info("Unknown user")
```
zeitos: can you add the given user to the log so it's easier to find which one failed?

zeitos:
@koettert this is great, thanks for contributing this great functionality! I added a couple of comments!

job-server/src/main/resources/application.conf
TimMaltGermany: sure

job-server/src/main/resources/application.conf
TimMaltGermany: No, not related to LDAP. I modified the comment as follows:
# note that ssl-encryption should also be on when authentication is on to avoid that passwords
# can be sniffed out

job-server/src/spark.jobserver/WebApi.scala
velvia: Yes, I also prefer to just have default as false.

job-server/src/main/resources/shiro.ini
TimMaltGermany: @velvia  Yes, makes a lot of sense. I moved the file. What else needs to be done / documented to make "this part of the deply process - Docker container"?

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
velvia: @zeitos In my newer projects I use `com.typesafe.scalalogging`.  It has traits that does the above.  We should consider adopting it.  The ActorStack stuff we can revisit as Kamon.io and others have implemented really nice Actor monitoring/debugging stuff.

job-server/src/spark.jobserver/JobServer.scala
TimMaltGermany: @noorul Well, this part is easy to test, but anyway: where else should we put it? It is essential that this small piece of code is executed very early during the start-up process. As we wrote in the description for this Pull Request: we would welcome suggestions on how to avoid this copying from the config into properties altogether.

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
```diff
@@ -0,0 +1,126 @@
...
+    val members = scala.collection.mutable.HashSet[String]()
+    if (attrs != null) {
+      val ae: NamingEnumeration[_ <: Attribute] = attrs.getAll();
+      while (ae.hasMore()) {
+        val attr: Attribute = ae.next();
+
+        if (attr.getID().equals("member")) {
+          val javaMembers: java.util.Collection[String] = LdapUtils.getAllAttributeValues(attr)
+
+          val iter: java.util.Iterator[String] = javaMembers.iterator()
```
velvia: You don't need the type ascriptions such as `java.util.Iterator[String]`.  I think you can shorten the above two lines into this, and it is still fairly readable:

    val iter = LdapUtils.getAllAttributeValues(attr).iterator()

Actually, the entire while loop is not necessary, because HashSet supports `++`.  So you can actually replace these 7 lines or so with this one liner:

    members ++= LdapUtils.getAllAttributeValues(attr).iterator().toSet

You might need to import `scala.collection.JavaConversions._` though.

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
```diff
@@ -0,0 +1,120 @@
...
+            None
+          }
+          case ice: IncorrectCredentialsException => {
```
velvia: Braces are not needed for case statements, I think it looks better without them....

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
velvia: Actually, does allowedGroups need to be an Option?  It would be super nice if it was just a Scala List or Seq, because then all the Some/None/Option stuff would not be needed.

job-server/src/spark.jobserver/JobServer.scala
velvia: Yeah, not sure that the code could be located anywhere else....  open to
ideas.

On Sun, Aug 30, 2015 at 11:34 PM, TimMaltGermany <notifications@github.com>
wrote:

> In job-server/src/spark.jobserver/JobServer.scala
> <https://github.com/spark-jobserver/spark-jobserver/pull/232#discussion_r38287266>
> :
>
> > @@ -26,6 +26,13 @@ import org.slf4j.LoggerFactory
> >  object JobServer {
> >    val logger = LoggerFactory.getLogger(getClass)
> >
> > +  private def setSystemPropertiesFromConfig(config: Config) {
>
> @noorul <https://github.com/noorul> Well, this part is easy to test, but
> anyway: where else should we put it? It is essential that this small piece
> of code is executed very early during the start-up process. As we wrote in
> the description for this Pull Request: we would welcome suggestions on how
> to avoid this copying from the config into properties altogether.
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/232/files#r38287266>
> .
>



-- 
   --- Toni Morrison

TimMaltGermany: I move the setter / getter code for this into a new utility object. JobServer now calls: 
Config2Properties.setSystemPropertiesFromConfig(config)
and the LdapGroupRealm reads it with:
private val searchBase: String = Config2Properties.getSystemProperty(Config2Properties.LDAP_SEARCH_BASE_PARAM)

That minimizes changes to the JobServer and quarantines the ugly implementation detail to Config2Properties.

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: sure, didn't know what your general policy was on this.
application.conf now says "shiro.authentication = off", removed hasPath in WebApi

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: done

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: Correct. Do you prefer an _ instead? We thought to leave it in just in case someone might want to use it at some point of time in the future.

job-server/src/spark.jobserver/auth/AuthInfo.scala
TimMaltGermany: yes, I will change the comment to make clear what happens here

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
TimMaltGermany: I have tried to use the JavaConversion before, but your scala style does not allow it:
spark-jobserver\job-server\src\spark.jobserver\auth\LdapGroupRealm.scala:14:0: Import from illegal package

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
TimMaltGermany: see below, without JavaConversions I have to stick to Java more or less

job-server/src/spark.jobserver/auth/LdapGroupRealm.scala
TimMaltGermany: see below, import is not allowed by scala style

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
TimMaltGermany: indeed

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
TimMaltGermany: @zeitos not sure what you mean by this - same as what before? Which template method? 
We don't check whether trace is enabled or not.

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
TimMaltGermany: done

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
TimMaltGermany: @velvia Well, I think yes, it needs to be wrapped in an Option as we need to be able distinguish between the case of an empty group list (meaning that no users are allowed) and the case when no groups are set at all (meaning that we don't want to check on groups). The alternative would be a separate boolean variable that indicates whether group authentication is requested at all.

TimMaltGermany:
@velvia @noorul Regarding the config issue - never mind. I found a much better way that does not require any of these hacks and uses the ini file. I guess it was one of those things that are too obvious to see. Testing it right now.

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
zeitos: I was referring to a previous comment about using logger.xxx(pattern,params) .
and yes, you are not checking if trace is enabled, but you should 
What I was trying to say is consider doing 
```
logger.trace("ACCESS GRANTED for user [%s]", fullName)
```

that will only perform the string concatenation if trace is enabled.

TimMaltGermany:
@velvia @noorul @zeitos @koettert Thanks guys for you very helpful comments. I guess my Scala got a bit rusty... I have changed quite a few things and believe to have simplified everything quite a bit as well. I am also adding a couple of tests. Finishing this will probably take another day or so. 
What is the general procedure for this? Should we create another pull request when done or push the new changes on top of this one? Given the number of changes, i would think it is better to create a new pull request if you don't mind.
Thanks,
Tim

noorul:
IMHO we should squash the into one

job-server/src/spark.jobserver/auth/SJSAuthenticator.scala
velvia: @TimMaltGermany got it thanks.

velvia:
@TimMaltGermany sure create a new PR.  Many thanks!

TimMaltGermany:
@velvia Hi Evan, can you please delete this pull request? I created a new one with cleaned-up history and such. Thanks, Tim

velvia:
Closing as new PR has been created.

PR #239, opened 2015-09-02
JavaSparkJob

velvia:
Add a JavaSparkJob class to make it easy for Java folks to write jobs for SJS, with a JavaHelloWorldJob example.

Missing:
* Adding java collections to the list of types serialized into JSON.


TimMaltGermany:
@velvia This is great. You might also consider adding Java support for NamedRDDs. I believe that 'namedRdds' cannot be directly accessed from Java code.

zeitos:
LGTM

PR #242, opened 2015-09-02
Update troubleshooting.md

eranation:
It took me a while to figure out I need to do this to make SQL/HiveContext work... thought it's worth sharing in the troubleshooting list.


velvia:
@eranation thanks!

PR #244, opened 2015-09-04
Adding Maana to users list

wildwood:
We're getting a lot of good use out of spark-jobserver, to provide a controlled and standardized access to our spark cluster.  I'd like to add Maana to the list of users.

Thanks!


velvia:
@wildwood many thanks!   Love to hear folks finding good use out of it!   :)

Issue #229, opened 2015-08-27
With PostgreSQL backend /jobs page fails to load after 135577 records


noorul:
Our JOBS database has the following number of records

demeter_spark_jobserver=# select count(*) from "JOBS";
 count
--------
 135577

Is this is a known issue with other backends?

velvia:
I think the JobSqlDao doesn't have a limit when retrieving jobs.  That's
probably the problem.

On Thu, Aug 27, 2015 at 5:02 AM, Noorul Islam K M <notifications@github.com>
wrote:

> Our JOBS database has the following number of records
>
> demeter_spark_jobserver=# select count(*) from "JOBS";
> count
>
> 135577
>
> Is this is a known issue with other backends?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/spark-jobserver/spark-jobserver/issues/229>.
>



-- 
If you are free, you need to free somebody else.
If you have some power, then your job is to empower somebody else.
   --- Toni Morrison

noorul:
It looks like, limit is done on the data that is fetched from DB. I think we should restrict data fetch from DB. I will work on a patch for this.

velvia:
@noorul thanks!

On Thu, Sep 3, 2015 at 10:42 AM, Noorul Islam K M <notifications@github.com>
wrote:

> It looks like, limit is done on the data that is fetched from DB. I think
> we should restrict data fetch from DB. I will work on a patch for this.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/spark-jobserver/spark-jobserver/issues/229#issuecomment-137523956>
> .
>



-- 
If you are free, you need to free somebody else.
If you have some power, then your job is to empower somebody else.
   --- Toni Morrison

zeitos:
Closing the issue since PR https://github.com/spark-jobserver/spark-jobserver/pull/248 was merged

PR #248, opened 2015-09-09
Fix issue #229: Always use limit while reading from DB

noorul:
Earlier we were fetching all DB records into memory and then running limit query in memory. For a bigger production system this is causing issues as the records grow beyond certain number. This patch does  the following

1. Modifies the signature of getJobInfos() to take limit parameter so that we always fetch certain number of records from DB.
2. Add new API getJobInfo() to retrieve a information about a specific Job Id. Earlier we were performing a lookup in memory.
3. Adjust tests accordingly.

### DIFF
--- a/job-server-extras/test/spark.jobserver/StreamingJobSpec.scala
+++ b/job-server-extras/test/spark.jobserver/StreamingJobSpec.scala
@@ -46,7 +46,7 @@ class StreamingJobSpec extends JobSpecBase(StreamingJobSpec.getNewSystem) {
         }
       }
       Thread sleep 1000
-      dao.getJobInfos.get(jobId).get match  {
+      dao.getJobInfo(jobId).get match  {
         case JobInfo(_, _, _, _, _, None, _) => {  }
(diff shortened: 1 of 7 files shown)


noorul:
@velvia Does this look okay?

velvia:
Sorry, been extremely busy with Cassandra summit prep.  Maybe @fotero can take a look?

-Evan

noorul:
I see that all checks passed.

velvia:
@noorul this is a good change, but the title really doesn't get at the whole scope of the change.  I thought this was just about adding a limit to the DB query, but you are also changing the API for getting the jobs back.  I think this is a good change, but please add a full description and change the title of the PR to match.  

As for the test see where my comment is ... two lines are commented out.  Let's fix that test .... commented out test code means something we are not testing.

noorul:
Ok, I will incorporate more info in description. Also I will uncomment the tests. I am not sure how I left it out.

noorul:
@velvia Done!

velvia:
@noorul thanks, merging.

PR #249, opened 2015-09-09
Add version in assembly jar file name

noorul:
Also rename spark-job-server to spark-jobserver to be more consistent.


velvia:
@noorul thanks.  Made a suggestion.

PR #250, opened 2015-09-09
Data Routes:

TimMaltGermany:
We need to frequently upload files to the server and reference them from within the JobServer. We therefore found it more convenient to add data routes to the JobServer. The implementation is more or less based on the jar routes and uses a simplified version of the JobFileDAO.

We also had to refactor WebApiSpec and WebApi due to size limitations. I will add comments to the PR at the relevant places.

Please let us know if this makes any sense at all or if there is a better way to handle file upload.

Cheers,
the KNIME team


job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -27,16 +28,63 @@ import org.apache.shiro.config.IniSecurityManagerFactory
 import org.apache.shiro.mgt.SecurityManager
 import org.apache.shiro.SecurityUtils
 
+object WebApi {
+  val StatusKey = "status"
```
TimMaltGermany: moved code from WebApi class into object so that it can be accessed from elsewhere

job-server/test/spark.jobserver/WebApiMainRoutesSpec.scala
```diff
@@ -0,0 +1,304 @@
+package spark.jobserver
+
```
TimMaltGermany: this new class contains all the original test cases from WebApiSpec

job-server/test/spark.jobserver/WebApiSpec.scala
```diff
@@ -111,287 +117,5 @@ with ScalatestRouteTest with HttpService {
       case GetJobConfig(_)          => sender ! config
     }
   }
-
```
TimMaltGermany: actual tests were moved into class WebApiMainRoutesSpec

velvia:
@TimMaltGermany thanks.  I just had a really quick look.  Thoughts/Q's:

* How will the spark job get access to the files?  Please document in the PR description and the EADME.   esp: how will it get access across diff types of jobs?
* Does this integrate with Spark's `addFiles` API to serve data on workers?

TimMaltGermany:
@velvia I just added this to the README.md: 
"These files are uploaded to the server and are stored in a local temporary directory on the server where the JobServer runs. The POST command returns the full pathname and filename of the uploaded file so that later jobs can work with this just the same as with any other server-local file. A job could therefore add this file to HDFS or distribute it to worker nodes via the SparkContext.addFile command.  	
For files that are larger than a few hundred MB, it is recommended to manually upload these files to the server or to directly add them to your HDFS."

So, to answer your questions: 
- the jobs can access these files directly using the full filename returned by the POST command
- it does not integrate with addFiles or import files into HDFS. I suppose we could add two sub-routes for this (POST data/hdfs and POST data/nodes) (maybe in a later PR?), but at least for our use cases that is an overkill. We use these routes to manage models and to import smallish data files.

velvia:
@TimMaltGermany got it, so the idea is to pass the pathname of the file to
the job yourself.

On Thu, Sep 10, 2015 at 1:27 AM, TimMaltGermany <notifications@github.com>
wrote:

> @velvia <https://github.com/velvia> I just added this to the README.md:
> "These files are uploaded to the server and are stored in a local
> temporary directory on the server where the JobServer runs. The POST
> command returns the full pathname and filename of the uploaded file so that
> later jobs can work with this just the same as with any other server-local
> file. A job could therefore add this file to HDFS or distribute it to
> worker nodes via the SparkContext.addFile command.
>
> For files that are larger than a few hundred MB, it is recommended to
> manually upload these files to the server or to directly add them to your
> HDFS."
>
> So, to answer your questions:
>
>    - the jobs can access these files directly using the full filename
>    returned by the POST command
>    - it does not integrate with addFiles or import files into HDFS. I
>    suppose we could add two sub-routes for this (POST data/hdfs and POST
>    data/nodes) (maybe in a later PR?), but at least for our use cases that is
>    an overkill. We use these routes to manage models and to import smallish
>    data files.
>
> —
> <https://github.com/spark-jobserver/spark-jobserver/pull/250#issuecomment-139165926>
> .
>



-- 
   --- Toni Morrison

koettert:
@velvia What is the status of this request? Do you need any further information or are any changes necessary? If not it would be great if the request could be merged soon before we get any new conflicts.
Thanks
Tobias

velvia:
@koettert sorry, been extremely busy.    No significant changes are expected in the near term;  but I agree with you.
@noorul @zeitos do you guys want to have a look?

zeitos:
LGTM!

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -139,6 +188,21 @@ class WebApi(system: ActorSystem,
   }
 
   /**
+   * Routes for listing, deletion of and storing data files
+   *    GET /data                     - lists all currently stored files
```
noorul: Just trying to get clarity. Will this survive server restart? 

1. Start job server
2. POST /data/file1, POST /data/file2
3. GET /data -> Expect two files file1 and file2
4. Restart jobserver
5. GET /data -> Can I expect the two file names to be returned?

job-server/src/spark.jobserver/routes/DataRoutes.scala
```diff
@@ -0,0 +1,76 @@
...
+      // POST /data/<filename>
+      post {
+        path(Segment) { filename =>
+          entity(as[Array[Byte]]) { bytes =>
+            val future = dataManager ? StoreData(filename, bytes)
+            respondWithMediaType(MediaTypes.`application/json`) { ctx =>
+              future.map {
+                case Stored(filename) => {
+                  ctx.complete(Map[String, Any](
+                    StatusKey -> "OK",
```
noorul: Do we really need a status key? I think HTTP status code has this information.

job-server/test/spark.jobserver/WebApiDataRoutesSpec.scala
```diff
@@ -0,0 +1,59 @@
+package spark.jobserver
```
noorul: I think the name of the file should be DataRouteSpec.scala.

There are tools which helps us to jump to corresponding scala source to spec source. And filename helps there.

noorul:
Functionally everything looks alright to me.

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: yes, only the file names will not be 'file1' and 'file', but the full file path (in both cases, before and after the restart)

job-server/src/spark.jobserver/routes/DataRoutes.scala
TimMaltGermany: Well, I just followed the pattern used elsewhere in the JSJ code, but I guess we don't really need it.

job-server/test/spark.jobserver/WebApiDataRoutesSpec.scala
TimMaltGermany: @noorul The class under test is called "DataRoutes.scala", so "DataRoutesSpec.scala" fits, doesn't it? Do you want me to rename both classes?

job-server/src/spark.jobserver/routes/DataRoutes.scala
TimMaltGermany: @noorul 'xjrk58' explicitly added "StatusKey -> "OK" in "Always respond with correct JSON #255", but if I understand correctly, then your suggestion is to remove it here, correct? Do you prefer:
case Stored(filename) => {
                  ctx.complete(StatusCodes.OK, Map[String, Any](
                    ResultKey -> Map("filename" -> filename)))
                }
or: 
case Stored(filename) => {
                  ctx.complete(Map[String, Any](
                    ResultKey -> Map("filename" -> filename)))
                }

?

noorul: I would go with the first one. Even though xjrk58 added #255, I don't think my comment is related to that.

TimMaltGermany: ok, thanks

TimMaltGermany:
@velvia @noorul @zeitos Many thanks for your comments. All tests passed, the latest head is merged in as well. So I believe that we might be ready to merge?

job-server/test/spark.jobserver/WebApiDataRoutesSpec.scala
velvia: Tim, I think DataRoutesSpec.scala is totally fine.

velvia:
I think so.

Going to merge it for now, as I’ll be traveling tomorrow.

PR #252, opened 2015-09-11
Support 1.4.1

zeitos:


zeitos:
@velvia So it looks like the issue was the `default.table` thingy... I changed that to just table and worked... I looked at spark release notes to check if that incompatibility was listed but I couldn't find it.

In any case: that's a spark issue not SJS.

zeitos:
@velvia do you want to merge it?

README.md
```diff
@@ -12,9 +12,9 @@ See [Troubleshooting Tips](doc/troubleshooting.md) as well as [Yarn tips](doc/ya
 
 (Please add yourself to this list!)
 
-- Ooyala
-- Netflix
-- Avenida.com
+- [Ooyala](http://www.ooyala.com)
```
noorul: I think this change is not related. Can we move this to separate patch?

job-server-extras/src/spark.jobserver/HiveTestJob.scala
```diff
@@ -10,15 +10,15 @@ import org.apache.spark.sql.hive.HiveContext
  * (Will create Hive metastore at job-server/metastore_db if Hive isn't configured)
  */
 object HiveLoaderJob extends SparkHiveJob {
-  // The following data is stored at ./test_addresses.txt
+  // The following data is stored at ./hive_test_job_addresses.txt
```
noorul: This change also seems to be not related to the intent of this patch.

noorul:
some nits, other than that LGTM

velvia:
@zeitos LGTM, gonna merge.  We also need to update the table of spark versions on the README, but I'll do that.
@noorul thanks for having a look!

PR #256, opened 2015-09-19
Spark 1.5 dependency

zeitos:
Creating a cleaner PR


README.md
```diff
@@ -49,7 +49,8 @@ See [Troubleshooting Tips](doc/troubleshooting.md) as well as [Yarn tips](doc/ya
 | 0.5.0       | 1.2.0         |
 | 0.5.1       | 1.3.0         |
 | 0.5.2       | 1.3.1         |
-| master      | 1.4.1         |
+| 0.5.3       | 1.4.1         |
```
noorul: Did we release 0.5.3 yet? Does this mean that there will not be a release that supports 1.4.1? Or will master be compatible with 1.4.1 too?

zeitos: we didn't my mistake, I need to remove that line.

zeitos: or we can wait for 0.5.3 to be released and then release this one as 0.6.0

velvia: My proposal was to wait to merge this until we release 0.6.0, which will have 1.4.1, and then master will have this and be on 1.5.0

The reason is that it seems wrong to skip Spark versions, from 1.3.x to 1.5.0, as there will be users who want to use 1.4.x in production.  Since job server is more used for prod, I think we can be slightly slower.

zeitos: I'm not sure why I thought I replied before.

Sounds great

velvia:
@zeitos say can we update the README in this PR to point versions at at 0.6.0/master?

zeitos:
@velvia done

velvia:
@zeitos thanks, going to merge

PR #259, opened 2015-09-22
Remove unnecesary fallback in SparkContextFactory 

zeitos:
There was a fallback in SparkContextFactory that made the default config override job config instead of the other way around.
Since LocalContextSupervisorActor and WebApi do the fallback action there's no need to do it again in the context factory.


velvia:
@zeitos thanks.

PR #260, opened 2015-09-23
Remove unnecessary logging of SQL query

noorul:
This was actually done for debugging. This could pollute the log.


zeitos:
Merged, thanks!

PR #270, opened 2015-10-06
Cleanup /jobs/jobId APIs

noorul:
As of now /jobs/jobId returns status and result which is very minimal.
Since we lack an API which returns more information like the /jobs API
returns as a list, amend /jobs/jobId to return extra information.

This commit makes job status uniform across /jobs and /jobs/jobId. We no
longer have "OK" status instead it will be the same as "FINISHED"

Now GET /jobs/jobId takes different route instead of the current one.

User => WebAPI => GetJobStatus => WebAPI => GetJobResult => WebAPI => User

And DELETE /jobs/jobId takes

User => WebAPI => GetJobStatus => WebAPI => User


velvia:
@noorul thanks.  At first glance this looks great to me.  Will look or merge later.

PR #284, opened 2015-10-14
Adds "context refresh" access point

wildwood:
With this, "GET /refresh-contexts" will remove all contexts on the job server (killing all jobs running in those contexts) and re-build the contexts as specified in config.

This provides a quicker development loop for people developing plugins, who need to clear the classpath each time they upload their jar.

I'm not sure how this figures into a RESTful api - other access points may make more sense.

Thanks!

--David from Maana


noorul:
Well, this is not restful, how about PUT /contexts/context-name ?

velvia:
I think the idea was to refresh ALL contexts.   

David, this is what I suggest:

PATCH /contexts?reset=true

It fits in with the existing /contexts routes, and also is a PATCH which implies it makes a change or updates/refreshes… GET requests should be immutable and effect free.  POST is already taken.

The extra param is just because it’s very invasive and you want to be really sure :)

This also opens the door for PATCH /contexts/<context-name> which would just refresh a single context.

noorul:
PATCH is for updating partially not fully. Here we are destroying the context and bringing it up.

velvia:
Yeah PATCH is not exact match, but can’t think of a better one.   There’s also PUT.

I suppose 

POST /contexts?reset=all   or something similar could do the trick.

noorul:
Also we can combine 

DELETE /contexts
POST /contexts/context-name

velvia:
Nah, separate REST calls are not as convenient.

wildwood:
Okay, I've got the changes made.

How does this process work, at this point?  Should I do a different pull request?  Or will pushing the changes to my fork's master automatically update the pull request?

velvia:
Just push your changes, no need for a different PR.

wildwood:
Hmmm... looks like WebApi is failing scalastyle, because it's over 500 lines.  Not sure how I got past that before.

How to handle?  Does WebApi need to be refactored?  Is a scalastyle exception in order?

velvia:
Eek.  I guess we can extend the limit for now.  It definitely needs to be refactored though.  The Context-Per-JVM PR refactors it, actually…  I need to find time to create that branch, and merge in all our recent changes since then.

wildwood:
Ok, changes are in, moving the access point to "PUT /contexts?reset=all".

Thoughts/Comments?

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -273,6 +273,33 @@ class WebApi(system: ActorSystem,
...
+          parameter("reset") { reset =>
+            respondWithMediaType(MediaTypes.`text/plain`) { ctx =>
+              reset match {
+                case "all" => {
+                  import ContextSupervisor._
+                  import collection.JavaConverters._
+                  logger.warn("refreshing contexts")
+                  val future = (supervisor ? ListContexts).mapTo[Seq[String]]
+                  Await.result(future, 1 second).asInstanceOf[Seq[String]]
+                    .foreach { context => {
```
velvia: Two thoughts: you should parallelize the futures here, and log each one.  Something like this:

    val stopFutures = contexts.map(c => supervisor ? StopContext(c))
    Await.ready(Future.sequence(stopFutures), 15 seconds)

Also I would advise making the timeout configurable.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -273,6 +273,33 @@ class WebApi(system: ActorSystem,
...
+                      val stopFuture = supervisor ? StopContext(context)
+                      Await.ready(stopFuture, 5 seconds)
+                    }
+                  }
+                  Thread.sleep(1000) // we apparently need some sleeping in here, so spark can catch up
+
+                  (supervisor ? AddContextsFromConfig).onFailure {
```
velvia: Note that this won't restart all the contexts from before, just the ones that are predefined in the config.... is that what you want?  Just clarifying....   my understanding of the feature was that it was all the contexts from before.  (We do have a problem we don't save the parameters used to start the contexts....)

velvia:
Added my comments.  Looks fine overall.

job-server/src/spark.jobserver/WebApi.scala
wildwood: Correct, this is the desired behavior - the config has the right sizing for our larger jobs, so re-loading from that makes the most sense.  As you mention, 'ListContexts' doesn't have enough detail in it to re-create the contexts at their preferred sizes.

Would it make sense to go with a different param value for this, like "reset=config" instead of "all"?

velvia: Hmm.  What if you had more contexts open than the ones listed in the config?  Say in config you have “MyContext”, “YourContext” and now I add “EvansContext”.  

I think `reset=config` and only resetting the contexts from the configs would be the most consistent behavior.  What do others think?

wildwood: Good point, the current behavior is more of a "restart", restoring the contexts to what they would look like if you re-started the job server.

scalastyle-config.xml
```diff
@@ -3,7 +3,7 @@
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"></check>
  <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxFileLength"><![CDATA[500]]></parameter>
+   <parameter name="maxFileLength"><![CDATA[525]]></parameter>
```
noorul: Can you increase this so that we don't have to keep doing this if we have to some more?

noorul:
Should we add new API to README.md?

velvia:
+1 to adding new API.

scalastyle-config.xml
velvia: I would vote no, well we just need to split the file, and work on the HA branch.

wildwood:
Sounds good - I added this access point to the 'contexts' part of the API docs on the README, and bumped the max file length to 1000.

velvia:
@wildwood Many thanks.  Going to merge this now, thanks again for the contribution!

PR #286, opened 2015-10-14
Fix regression caused by PR-270

noorul:
After PR-270 DELETE API for following scenarios stopped working.
      DELETE /jobs/non-existing-job-id
      DELETE /jobs/finished-job-id stopped working.


noorul:
@velvia This is a regression fix. Can you take a look?

velvia:
@noorul looks fine to me.  How about pulling in my 1.5.1 update from master so that Travis can pass this PR?

noorul:
@velvia Done!

PR #297, opened 2015-10-30
Add DSE Spark Master Provider

noorul:


noorul:
@RussellSpitzer Can you please review this and provide feedback? Or do you have something similar @DataStax ?

velvia:
Noorul, I think you would not need this if you were using DSE no?  Or is this for an older version of DSE?

-Evan

noorul:
We need this. When DSE elects new master and we need to restart spark jobserver then it will pick the hardcoded master from local.conf.

velvia:
What I mean is that DSE 4.8 already includes the job server, and would surely have something like this.... including it could conflict with future versions / changes ...

-Evan

RussellSpitzer:
No we don't include the "master provider" yet :) It got pushed back

noorul:
We are not using sjs that is coming with dse 4.8.0. Also as Russel said they did not implement this. Can this be merged?

velvia:
Sure.  Lemme have a closer look later.  

-Evan

velvia:
@noorul left you some minor comments, mostly style, but I think would help make code more compact and easier to read.

Now here's the thing - I think overall this is valuable, but the SparkMasterProvider is meant to be a pluggable interface, and not everyone uses DSE.   I think a better place for this is as a separate module, or separate repo, and those that want to include it can then include it, without the need to increase the dependencies of core.  What do you think?   

That could be a spark-jobserver-cassandra module, with not only your provider, but other goodies: a Cassandra DAO, and perhaps a CassandraContext for parsing C* tables.

noorul:
I agree that the code should belong somewhere else. Initially I though of
debating this in the mailing list before raising PR, then I thought any how
you comment on that in the PR.

noorul:
@velvia Shall I create repo named spark-jobserver-cassandra? If so, is there any project template to get the basic structure up?

velvia:
@noorul hmmm.  Yeah we can create that repo.  However for project template...     how about this, let me create that repo, and set up the basic structure.   I have an idea to use it for other things, like to create a cassandra DAO and cassandra specific query jobs.  In the meantime, do you want to address my other CR comments first, and push changes to this PR?

noorul:
Sure! Thanks!

noorul:
We are looking to go live with code from this repository. It will be great we could merge this soon.

velvia:
@noorul I'm going to go ahead and create the other repo now.

velvia:
@noorul repo added, go ahead and put your code their.  Added build files, including ability to build assembly.  Should be good to go.  Can then close this once code is moved over?

velvia:
Going to close this, since we have the other repo for this now.

PR #300, opened 2015-11-01
DataBase migration

v-gerasimov:
DataBase migration by Flyway for H2 and PostgreSQL.

Change column types:
 * Column "ERROR" in table "JOBS" is ```TEXT``` type for PostgreSQL and ```CLOB``` type for H2 now
 * Column "JOB_CONFIG" in table "CONFIGS" is ```TEXT```  type for PostgreSQL and```CLOB``` type for H2 now

New application.conf parameter: ```flyway.locations```


job-server/src/spark.jobserver/io/JobSqlDAO.scala
```diff
@@ -81,6 +82,7 @@ class JobSqlDAO(config: Config) extends JobDAO {
     ds
   }
   val db = Database.forDataSource(dataSource)
+  val migrateLocations = config.getString("flyway.locations")
```
v-gerasimov: Not need, will remove that.

job-server/src/spark.jobserver/io/JobSqlDAO.scala
```diff
@@ -93,38 +95,13 @@ class JobSqlDAO(config: Config) extends JobDAO {
...
-          // Exit the Job Server now.
-          throw new RuntimeException("Some tables in metadata-store are missing. Please recover it first.")
-        }
-
-        // If the cases above are not true, it means all tables exit. No need to initialize the database.
-    }
+    // Flyway migration
+    val flyway = new Flyway()
+    flyway.setDataSource(jdbcUrl, jdbcUser, jdbcPassword)
+    flyway.setLocations(migrateLocations)
```
v-gerasimov: Not need, will remove that.

v-gerasimov: This logical needs for tests. It isn't need if test will have reference.conf or something else.

v-gerasimov:
I have used flyway for fix problem with ```slick 2.1```. In ```slick 2.1``` you cannot create a field with ```VARCHAR``` type without limitation. It was fixed in ```slick 3.0``` by https://github.com/slick/slick/issues/129. When I start spark-jobserver on PostgreSQL I will get ```VARCHAR(254)``` column types and will get ```ERROR: value too long for type character varying(254)```. For solve this problem we can migrate to ```slick 3.0``` (that would be great) but I think that if we use flyway it will be a right choice.

velvia:
@v-gerasimov thanks.  Will have a look a bit later.

job-server/src/main/resources/application.conf
```diff
@@ -162,3 +162,5 @@ shiro {
     # Time out for job server to wait for authentication requests
     authentication-timeout = 10 s
 }
+
+flyway.locations = "db/h2/migration"
```
velvia: I would put this under the spark.jobserver hierarchy, and add a comment that this is for migrations.

job-server/src/spark.jobserver/io/JobSqlDAO.scala
velvia: Ok, can you add a comment about this, with a TODO?

job-server/src/test/resources/db/h2/migration/V0_7_0__init_tables.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE JARS (
```
velvia: Hmm.... is this exactly the same as the one in `src/main/resources`?  Should we have a symlink instead?   Would be very easy to forget to update the tables.

velvia:
@v-gerasimov thanks and sorry for the delay.  LGTM except for minor comments.

job-server/src/test/resources/db/h2/migration/V0_7_0__init_tables.sql
v-gerasimov: Good idea! I have done it.

job-server/src/main/resources/application.conf
v-gerasimov: I put this under spark hierarchy and added a comment.

velvia:
@v-gerasimov sorry I didn't realize you pushed new changes.  Next time just comment on it afterwards.  Unfortunately AFAIK Github doesn't notify for individual commits, just comments.

Going to merge.

PR #301, opened 2015-11-03
Docker image improvements and parameterized Mesos version.

danisla:
This changes the base docker image from ottoops/mesos-java7 to java:7-jre, this alone reduces the image size by about half, ~785MB from ~1.54GB.

If there is a strong preference towards the OracleJDK, we can merge in the [oracle-java7-installer](https://github.com/OttoOps/docker-containers/blob/master/mesos-java7/Dockerfile#L6).

The Mesos library is now installed from apt and configurable via the MESOS_VERSION environment variable.

The docker tag now contains the version for Mesos and Spark, example: 0.6.1-SNAPSHOT.mesos-0.25.0.spark-1.5.1. This should let users discern the correct tag to pull based on their environment.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -102,10 +102,17 @@ object JobServerBuild extends Build {
       val artifact = (AssemblyKeys.outputPath in AssemblyKeys.assembly in jobServerExtras).value
       val artifactTargetPath = s"/app/${artifact.name}"
       new sbtdocker.mutable.Dockerfile {
-        from("ottoops/mesos-java7")
+        from("java:7-jre")
         // Dockerfile best practices: https://docs.docker.com/articles/dockerfile_best-practices/
         expose(8090)
         expose(9999)    // for JMX
+        env("MESOS_VERSION", mesosVersion)
+        runRaw("""echo "deb http://repos.mesosphere.io/ubuntu/ trusty main" > /etc/apt/sources.list.d/mesosphere.list && \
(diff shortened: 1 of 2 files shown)


kiambogo:
One concern I have is that not every user of SJS will be running Spark on Mesos, and with this it will be bundled into every image produced. This would also create some confusion for users who will not be using Mesos and want to pull a Docker image of SJS (seeing the images tagged with mesos).

I like the idea of having this patch included, especially with the ability to produce images tagged with the mesos/spark version/hadoop? version...but that's a lot of combinations. At some point it should probably be on the user to clone/fork, customize to their needs, and publish their own image.

velvia:
As long as it’s easy for users to fork and change the build based on the base docker image, I think it’s fine … that’s the idea of Docker right?   Easy to fork and modify images.

Generating different variants of the base image based on Hadoop version, etc…. haven’t got the time to do this in the near future, but if someone wants to volunteer and start a discussion, please file a Github issue and be glad to support that.

As for including Mesos lib by default, I don’t see anything wrong with that.  It’s very small, doesn’t really add any deps to Spark in terms of the JVM (unlike Hadoop distros).

danisla:
I'm on board with leaving it to the users to build and push their own images. This patch should make that easier for people. 

Should I remove the tag convention that appends the mesos and spark version?

velvia:
@danisla thanks for your contribution!  Going to merge.  Now we have to figure out when to build this, probably soon  :)

PR #319, opened 2015-11-23
Update spark version to 1.5.2

apivovarov:
Spark 1.5.2 was released on Nov 9 https://github.com/apache/spark/releases/tag/v1.5.2


velvia:
@apivovarov thanks.  Just in time, we can upgrade just before we release a new version.

PR #322, opened 2015-11-25
Update yarn related docs.

lant:
As discussed in the mailing list, the start-server.sh does not contain the '--master yarn-client' option because it is not used in all kinds of deployments. The yarn docs don't say anything about it. I just added few lines into the relevant doc page.


velvia:
@lant thanks!

PR #324, opened 2015-11-25
Replace spark 1.5.0 with 1.5.2 in all configs and docs

apivovarov:
It's follow up commit for the previous commit to update spark version to 1.5.2


velvia:
@apivovarov ah thanks, just in time.

apivovarov:
Can it be included to v.0.6.1?

PR #329, opened 2015-11-29
Fix bug - invalid error w no user/pass

slater-ben:
Fixes a bug that causes java.util.NoSuchElementException: None.get error when operation attempted with auth enabled and no user name / password supplied. This bug in turn seemed to stop the UI working the auth enabled.

I'm not much of a scala expert so there may be a better way to do it but this worked in all my testing.


velvia:
@slater-ben thanks, I think this will do.

PR #332, opened 2015-12-02
Improve the SparkJob API to remove code duplication

reactormonk:
Upsides:
- No code duplication of request parsing
- Forces better coding style, no more validation all over the place

Downsides:
- Breaks API
- `type Tmp = ...` required (can't use generics because of how job classes are loaded)

Things to consider:
- Use `util.Try` instead of `scalaz.Validation`

Bugs:
- somehow the Java API doesn't compile with

```
[error] spark-jobserver/job-server-tests/src/main/java/spark.jobserver/JavaHelloWorldJob.java:7: error: JavaHelloWorldJob is not abstract and does not override abstract method validate(Object,Config) in SparkJobBase
[error] public class JavaHelloWorldJob extends JavaSparkJob {
[error]        ^
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,16 +197,16 @@ To create a job that can be submitted through the job server, the job must imple
 Your job will look like:
 ```scala
 object SampleJob  extends SparkJob {
-    override def runJob(sc:SparkContext, jobConfig: Config): Any = ???
-    override def validate(sc:SparkContext, config: Config): SparkJobValidation = ???
+    override def runJob(sc:SparkContext, data: Tmp): Any = ???
+    type Tmp = ...
+    override def validate(sc:SparkContext, config: Config): scalaz.Validation[String, Tmp] = ???
 }
 ```
 
 - `runJob` contains the implementation of the Job. The SparkContext is managed by the JobServer and will be provided to the job through this method.
   This relieves the developer from the boiler-plate configuration management that comes with the creation of a Spark job and allows the Job Server to
 manage and re-use contexts.
-- `validate` allows for an initial validation of the context and any provided configuration. If the context and configuration are OK to run the job, returning `spark.jobserver.SparkJobValid` will let the job execute, otherwise returning `spark.jobserver.SparkJobInvalid(reason)` prevents the job from running and provides means to convey the reason of failure. In this case, the call immediately returns an `HTTP/1.1 400 Bad Request` status code.  
-`validate` helps you preventing running jobs that will eventually fail due to missing or wrong configuration and save both time and resources.  
+- `validate` allows for an initial validation of the context and any provided configuration. If the context and configuration are OK to run the job, returning `scalaz.Success` will let the job execute and pass the returned value to `runJob`. Returning `scalaz.Failure(reason)` prevents the job from running and provides means to convey the reason of failure. In this case, the call immediately returns an `HTTP/1.1 400 Bad Request` status code.  
 
 Let's try running our sample job with an invalid configuration:
 
@@ -239,9 +239,10 @@ Let's try running our sample job with an invalid configuration:
 Named RDDs are a way to easily share RDDs among job. Using this facility, computed RDDs can be cached with a given name and later on retrieved.
 To use this feature, the SparkJob needs to mixin `NamedRddSupport`:
 ```scala
-object SampleNamedRDDJob  extends SparkJob with NamedRddSupport {
-    override def runJob(sc:SparkContext, jobConfig: Config): Any = ???
-    override def validate(sc:SparkContext, config: Config): SparkJobValidation = ???
+object SampleNamedRDDJob extends SparkJob with NamedRddSupport {
+    override def runJob(sc:SparkContext, jobConfig: Tmp): Any = ???
+    type Tmp = ...
+    override def validate(sc:SparkContext, config: Config): scalaz.Validation[String, Tmp] = ???
 }
 ```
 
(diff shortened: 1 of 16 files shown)


velvia:
@reactormonk I like the idea, but I thought of a way to do this without breaking the API.   Could type members be overridden?   Maybe we can write `SparkJobBase` such that it has the Scalaz based validation and new runJob methods, but they are renamed to new methods?   Then, the existing `validate` and `runJob` methods would just call the new methods, and have a default `Tmp` (I think better name might be `JobData`) = `Config` type to maintain compatibility.  That way existing jobs still work, with a  recompilation, but people could choose to implement the new method if we want.  Meantime the examples would be modified as you did to use the new methods so folks can see the benefits.

 What do you think?

Also, could we still pass the job config in to `runJob` or whatever it's new method name would be (maybe `runJobWithData`?)   The reason is some jobs want to use the input system config / parameters, and there's a request to pass in the job ID and other info into a job.

thanks,
Evan

reactormonk:
Yes, I think renaming the methods would be a good idea. Maybe `parse` and `run`? And I approve of `type JobData`. I don't think it's a good idea to access `Config` from `run`, because it's entirely unvalidated, and you'll have to parse it at some point anyway. If `Config` isn't passed to `run`, we force the user to fail early, such that a submitted job can't die because of an invalid `Config`. If there's more information from the jobserver available when running `run`, I'd simply add another parameter to `run`, e.g. a `JobserverContext` class or similar.

reactormonk:
I've tried the `type JobData = Config` default, but it doesn't work because

    type JobData has incompatible type

when you

    override type JobData = Object

because the `= Config` part places a subtype condition on `JobData`. I could try to keep source compatibility with some implicit magic, but that would break binary.

velvia:
Hmmm.   I suppose we could make a more complicated hierarchy.

SparkJobBase - has your new methods

OldSparkJobBase extends SparkJobBase - implements runJob and validate() based on your new methods with type JobData = Config (ignoring comments about why Config is bad — yes its unvalidated)

but need to think about what to do after that.

Is the “override type JobData = Object” for the JavaSparkJob?   Hmmm there’s gotta be a better solution than Object….

@zeitos and others, want to chime in?

reactormonk:
> Is the “override type JobData = Object” for the JavaSparkJob?   Hmmm there’s gotta be a better solution than Object….
No, there is not. It's a path-dependent type, because for each jar there's a different type, so the type is entirely runtime-dependent.

zeitos:
Hey, I agree with not breaking the API, we need a solution that can deprecate the old API before breaking it.
Note that while most runJob implementations will use the config object to get values for running the job, others might also use them to configure the context (ie: streaming jobs)

I don't like having a validate method that does validation AND parsing.

I'll try to take a closer look at the code this weekend.

reactormonk:
How do you do validation without parsing? Also, if you really want to pass along the `Config` object, you still can.

velvia:
For certain applications, I can see that validation may be just checking that certain config keys are there and are the right type, they might not want to create another type.  For other applications it would be more beneficial to parse.   

Overall I’m OK with @reactormonk’s design, but want to see if there’s still any way we can maintain some compatibility — and as I said, I’d like to separate out job config from system / job server config, which is currently fused together.

zeitos:
@reactormonk what I was trying to say is that I don't like the name validation if it does more than a validation, in fact in your code you have a parse method that you call from the validation.

On the other side I took a look at scalaz Validation and really liked the thing that it allows you to collect errors (ValidationNEL)

reactormonk:
@zeitos `parse` is the old name of the method, because the return type signature changed, you can't overload it with the new signature. But if we were to move to a new class/trait name, `parse` would be available again. But I think `validate` is better, because usually first you parse, then you validate, so validating includes parsing, but parsing doesn't include validating.

zeitos:
@reactormonk now I'm not sure what you mean by "parsed structure", I was talking about parsing with the definition you have in your code (converting a Config into a Tmp, so the parsed structure would be a success parsing of config objects and building a Tmp from it).

Anyway as I said: it looks like returning values is the way scalaz works and it might be useful in scala, I might be influenced by years of java.

velvia:
Guys, I’ll hack on this a little bit after the context-per-ivm stuff is merged.  In the meantime, if anybody has more thoughts on any ways to keep compatibility that’d be great.

velvia:
@reactormonk I'd like to revisit this PR.  We definitely need a new API.  It should include:
- Returning a real validation type (but what do you think of using Scalactic instead of Scalaz's ValidationNEL?  I started using Scalactic and really like it.  See http://doc.scalatest.org/2.2.4/index.html#org.scalactic.Or)
- Passing something like a `RuntimeContext` into `runJob` so that things like the job ID, context configuration, named objects, etc. can be passed in and accessed programmatically in a nice way for job writers

You had said that the JavaJob support was the only stumbling block, but I'd like to take this opportunity to rework the Java support and make it more workable.

I'm debating if I should write up a google doc first, or if I should just work on a  PR and have folks comment on the changes.

BTW path dependent types are not friendly for beginners.  I just wrote a blog post about them...

velvia:
Here is a google doc:
https://docs.google.com/document/d/1asrHGjwnMGDCNl_pZxV4RrcMqtfgdjzAOWYlLqRF2sw/edit

reactormonk:
From what I see the scalastic types are isomorphic to the scalaz/cats, except without all the things that make them interesting - the functional typeclasses. I'd prefer one of these two.

You can't get rid of path-dependent types unless you want to start casting. You can't use generics because you're loading the jars in. So path-dependent types are the lesser evil.

zeitos:
About path-dependent types... maybe we can use the jar manifest to optionally define aliases.
@velvia ?

velvia:
Folks, going to close this as I have created a new SparkJob API PR:  
https://github.com/spark-jobserver/spark-jobserver/pull/442

PR #336, opened 2015-12-05
Update Chinese document with a better format and latest content.

zp-j:
Update Chinese documentation.

- Reformated doc with mardown syntax
- Updated doc with latest content


velvia:
@zp-j 謝謝。 Only thing is now we have to update the version table in two places ... u think we can refer to version table in master README?

zp-j:
Hi @velvia, thanks! Absolutely I agree with you. Updated here: [https://github.com/zp-j/spark-jobserver/commit/4d53b3f75f365c5237cb7bcc60846a4ca5c4f734](https://github.com/zp-j/spark-jobserver/commit/4d53b3f75f365c5237cb7bcc60846a4ca5c4f734)

BTW, I will work on the rest documentation when I have some time. :D

velvia:
Seems good to me.  Not sure why the test fails with a purely doc change.

PR #120, opened 2015-04-08
JVM per context

Ankit1010:
Running each context in a separate process for issue #87

PR #345, opened 2015-12-13
Separate JVM per SparkContext

velvia:
This is the long awaited continuation of the original PR #120 to fork off a separate process per SparkContext.   One of the most requested features and makes the job server officially much more scalable.   You can see #120 for the history, but here is a very brief summary:

- If the `spark.jobserver.context-per-jvm` setting is set to `true`, then the JobServer will use `AkkaClusterSupervisorActor`, which forks off a separate JVM to run each `JobManagerActor`, using the `manager_start.sh` script.  If it is false, then it uses the old `LocalContextSupervisorActor`.
- In `application.conf`, the above setting is set to false, to enable `sbt reStart` to run without forking JVM for ease of local development
- In the default deployment configs, this property is set to true, so that by default deployed SJS will use this setting
- The forked JVM talks back to the SJS WebApi via Akka cluster / remote actors.  Right now this is all local, so a new port should not need to be opened.
- The forked JVM does not save directly to the database, but rather sends updates to a `JobDAOActor`.

Currently almost all changes in master, including Spark 1.5.2 compatibility, are merged into this branch, and the tests *should* be all passing.  Please test this out in different environments!


velvia:
@PierreVincent @zeitos here it is!   If you wish to review changes I would start at "Fix shell problem" 3cfafea, most of the ones above that are from original PR #120

job-server/src/main/resources/application.conf
```diff
@@ -124,6 +129,26 @@ spark {
 akka {
   # Use SLF4J/logback for deployed environment logging
   loggers = ["akka.event.slf4j.Slf4jLogger"]
+
+  actor {
+    provider = "akka.cluster.ClusterActorRefProvider"
```
helena: Always spin up a cluster even for those not needed > 1 context? My one thought is it's a lot of chatter for not needing this new feature, unless leveraging things like leader election etc regardless :)

job-server/src/main/resources/application.conf
```diff
@@ -124,6 +129,26 @@ spark {
...
+      port = 0
+      send-buffer-size = 512000b
+      receive-buffer-size = 512000b
+      maximum-frame-size = 256000b
+    }
+  }
+
+  cluster {
+    auto-down-unreachable-after = 20s
+    metrics.enabled = off
```
helena: :( no adaptive load balancing router? To chatty for this I'm sure and a user can override.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  override def preStart(): Unit = {
+    cluster.join(selfAddress)
+    cluster.subscribe(self, initialStateMode = InitialStateAsEvents, classOf[MemberEvent])
+  }
+
+  override def postStop(): Unit = {
+    cluster.unsubscribe(self)
+  }
+
```
helena: Hrm, you have 4 HashMaps in one actor storing data. What if the supervisor crashes or restarts - I'd want some kind of error handling of all this data (have you tested under load of a high cluster and long-running to see how much it stores?) in preRestart perhaps but not sure what.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+          if (initInfoOpt.isDefined && callbackOpt.isDefined) {
+            val (ctxName, ctxConf, isAdHoc) = initInfoOpt.get
+            val (successFunc, failureFunc) = callbackOpt.get
+            initContext(actorName, actorRef, contextInitTimeout)(
+              ctxName, ctxConf, isAdHoc)(successFunc, failureFunc)
+          }
+          else {
+            logger.warn("No initialization or callback found for jobManager actor {}", actorRef.path)
+            actorRef ! PoisonPill
+          }
```
helena: Hrm, you do something similar above. I'd recommend a for comp. like this vs if/else, and lose the '.get' calls, unless you can insure they will never result in a None and throw an error.
      
     (for (a <- opt1; b <- opt2) yield initContext(...)) getOrElse {
         logger.warn("No initialization or callback found for jobManager actor {}", actorRef.path)
         actorRef ! PoisonPill
      }

job-server/src/main/resources/application.conf
velvia: You can't know at startup time how many contexts a user will spin up, but there is a flag to turn this off. 

-Evan

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  private def addContextsFromConfig(config: Config) {
+    for (contexts <- Try(config.getObject("spark.contexts"))) {
+      contexts.keySet().asScala.foreach { contextName =>
+        val contextConfig = config.getConfig("spark.contexts." + contextName)
+          .withFallback(defaultContextConfig)
+        startContext(contextName, contextConfig, false) { ref =>} {
+          e => logger.error("Unable to start context" + contextName, e)
+        }
+        Thread sleep 1000 // Give some spacing so multiple contexts can be created
```
helena: A Thread.sleep in an actor??

job-server/src/spark.jobserver/JarManager.scala
```diff
@@ -28,14 +30,23 @@ case object JarStored
...
   }
 
   override def wrappedReceive: Receive = {
-    case ListJars => sender ! createJarsList()
+    case ListJars =>
+      import akka.pattern.{ask, pipe}
+      import context.dispatcher
+
+      val resp = (jobDao ? JobDAOActor.GetApps)(daoAskTimeout).mapTo[JobDAOActor.Apps]
+      resp.map { msg => msg.apps } pipeTo sender()
```
helena: Capture the sender before executing the Future.

job-server/src/spark.jobserver/JobManager.scala
```diff
@@ -0,0 +1,63 @@
...
+import akka.cluster.Cluster
+import com.typesafe.config.{ConfigValueFactory, ConfigFactory, Config}
+import ooyala.common.akka.actor.ProductionReaper
+import ooyala.common.akka.actor.Reaper.WatchMe
+import org.slf4j.LoggerFactory
+import spark.jobserver.io.{JobDAOActor, JobDAO}
+
+/**
+ * Created by ankits on 4/7/15.
+ */
```
helena: I don't see the rest of the codebase doing this so to keep in style with that I'd remove it. You'd be in the git history ;)

job-server/src/spark.jobserver/JobManagerActor.scala
```diff
@@ -153,7 +166,17 @@ class JobManagerActor(dao: JobDAO,
...
+      import akka.util.Timeout
+      import scala.concurrent.duration._
+      import scala.concurrent.Await
+
+      val daoAskTimeout = Timeout(3 seconds)
+      val resp = Await.result(
+        (daoActor ? JobDAOActor.GetLastUploadTime(appName))(daoAskTimeout).mapTo[JobDAOActor.LastUploadTime],
+        daoAskTimeout.duration)
+
+      val lastUploadTime = resp.lastUploadTime
```
helena: I don't recommend using Await in an actor. It looks like you can do a for comp here with the Future.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  logger.info("AkkaClusterSupervisor initialized on {}", selfAddress)
+
+  override def preStart(): Unit = {
+    cluster.join(selfAddress)
+    cluster.subscribe(self, initialStateMode = InitialStateAsEvents, classOf[MemberEvent])
+  }
+
+  override def postStop(): Unit = {
+    cluster.unsubscribe(self)
```
helena: This is a more graceful way to leave the cluster than just the System.exit you do elsewhere
      cluster leave selfAddress

job-server/src/spark.jobserver/io/JobDAOActor.scala
```diff
@@ -0,0 +1,70 @@
...
+import spark.jobserver.io.JobDAOActor.SaveJar
+import spark.jobserver.util.JarUtils
+
+/**
+ * Created by ankits on 4/6/15.
+ */
+
+object JobDAOActor {
+
+  //Requests
```
helena: I'd create a sealed trait base class for all job actor requests, similarly for the responses, and perhaps a task one for Save.. etc.

job-server/test/spark.jobserver/WebApiSpec.scala
```diff
@@ -42,7 +42,7 @@ with ScalatestRouteTest with HttpService {
   // See http://doc.akka.io/docs/akka/2.2.4/scala/actors.html#Deprecated_Variants;
   // for actors declared as inner classes we need to pass this as first arg
   val dummyActor = system.actorOf(Props(classOf[DummyActor], this))
-  val statusActor = system.actorOf(Props(classOf[JobStatusActor], new InMemoryDAO))
+  val statusActor = system.actorOf(Props(classOf[JobStatusActor], system.actorOf(Props(classOf[JobDAOActor], new InMemoryDAO))))
```
ktoso: Create an `JobStatusActor.props` for it (easier to maintain changes in param lists etc)?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Oh yeah, good catch.  That was old and copied from the days when Spark used global JVM settings to initialize each context and the only way to launch multiple contexts safely was to wait and then clear out the JVM global settings.  Ugh.

velvia:
@helena thanks for the comments.  Will address at some point.

Note to self and others:  `server_stop.sh` right now just sends a SIGTERM to the main web process, which leaves the other contexts hanging with no way to address them.  Need to fix this.

zeitos:
@velvia  this is great! I'll have a look now!

bin/server_start.sh
```diff
@@ -11,13 +11,15 @@ set -e
 
 get_abs_script_path() {
   pushd . >/dev/null
-  cd $(dirname $0)
+  cd "$(dirname "$0")"
```
zeitos: why do we need quotes now?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+import akka.util.Timeout
+import com.typesafe.config.Config
+import ooyala.common.akka.InstrumentedActor
+import spark.jobserver.util.SparkJobUtils
+import scala.collection.mutable
+import scala.util.{Try, Success, Failure}
+import scala.sys.process._
+
+/**
+ * Created by ankits on 4/7/15.
```
zeitos: Should we add some documentation here?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+    case AddContext(name, contextConfig) =>
+      val originator = sender()
+      val mergedConfig = contextConfig.withFallback(defaultContextConfig)
+      if (contexts contains name) {
+        originator ! ContextAlreadyExists
+      } else {
+        startContext(name, mergedConfig, false) { ref =>
+          originator ! ContextInitialized
+        } { err =>
+          originator ! ContextInitError(err)
```
zeitos: The error handling seems the same here and in GetAdHocContext, you could do something like
```
  def handleError(originator: ActorRef): (Throwable => Unit) = { err =>
    originator ! ContextInitError(err)
  }
```
and then 
 ```
(handleError(originator))
```

File: job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala [hunk 15]
```diff
@@ -0,0 +1,217 @@
...
+        }
+      }
+
+    case GetAdHocContext(classPath, contextConfig) =>
+      val originator = sender()
+      val mergedConfig = contextConfig.withFallback(defaultContextConfig)
+
+      var contextName = ""
+      do {
+        contextName = java.util.UUID.randomUUID().toString().substring(0, 8) + "-" + classPath
```
zeitos: you can do take here

job-server/src/main/resources/application.conf
```diff
@@ -124,6 +129,26 @@ spark {
 akka {
   # Use SLF4J/logback for deployed environment logging
```
analytically: Add

`warn-about-java-serializer-usage = off`

job-server/src/main/resources/application.conf
```diff
@@ -124,6 +129,26 @@ spark {
...
+      hostname = "127.0.0.1"
+      port = 0
+      send-buffer-size = 512000b
+      receive-buffer-size = 512000b
+      maximum-frame-size = 256000b
+    }
+  }
+
+  cluster {
+    auto-down-unreachable-after = 20s
```
analytically: Better deal with chatty networks:

```
failure-detector.acceptable-heartbeat-pause = 3s
failure-detector.threshold = 12.0
```

job-server/src/main/resources/application.conf
analytically: Insert here:

`warn-about-java-serializer-usage = off`

project/Dependencies.scala
```diff
@@ -21,6 +21,7 @@ object Dependencies {
     // Akka is provided because Spark already includes it, and Spark's version is shaded so it's not safe
     // to use this one
     "com.typesafe.akka" %% "akka-slf4j" % "2.3.4" % "provided",
+    "com.typesafe.akka" %% "akka-cluster" % "2.3.4",
```
analytically: Should this not be intransitive?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+      } while (contexts contains contextName)
```
zeitos: This will try to check that we don't have 2 contexts with the same name, but we set the value only when the context has been initialized, that leaves a lot of time for other requests to create the same context.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+        resultActors.remove(name)
+        sender ! ContextStopped
+      } else {
+        sender ! NoSuchContext
+      }
+
+    case Terminated(actorRef) =>
+      val name: String = actorRef.path.name
+      logger.info("Actor terminated: {}", name)
+      contexts.foreach { kv => if (kv._2 == actorRef) contexts.remove(kv._1) }
```
zeitos: Can we have the same actorRef with more than one context name?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+  val managerStartCommand = config.getString("deploy.manager-start-cmd")
+  import context.dispatcher
+
+  //actor name -> (context name, context config, context isadhoc)
+  private val contextInitInfos = mutable.HashMap.empty[String, (String, Config, Boolean)]
+  //actor name -> (success callback, failure callback)
+  private val contextCallbacks = mutable.HashMap.empty[String, (ActorRef => Unit, Throwable => Unit)]
+
+  private val contexts = mutable.HashMap.empty[String, ActorRef]
+  private val resultActors = mutable.HashMap.empty[String, ActorRef]
```
zeitos: Maybe we should have a Map[String, (ActorRef, ActorRef)] (contextActor, resultActor) ? they kind of work together

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
zeitos: or maybe check on contextInitInfos that is being set in startContext (still: it's not atomic)

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+        resultActors(ctxName) = resultActor
+        successFunc(ref)
+    }
+
+  }
+
+  private def startContext(name: String, contextConfig: Config, isAdHoc: Boolean)
+                          (successFunc: ActorRef => Unit)(failureFunc: Throwable => Unit) = {
+    require(!(contexts contains name), "There is already a context named " + name)
+    val contextActorName = "jobManager-" + java.util.UUID.randomUUID().toString.substring(16)
```
zeitos: no collision here?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+
+    logger.info("Starting context with actor name {} ", contextActorName)
+
+    val pb = Process(s"$managerStartCommand $contextActorName ${selfAddress.toString}")
```
zeitos: I was expecting s"$managerStartCommand $name ${selfAddress.toString}" 
Is this an error or we are not usiong the name to name the contextActor?

zeitos: Also I think toString is not needed.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,217 @@
...
+        originator ! ContextAlreadyExists
+      } else {
+        startContext(name, mergedConfig, false) { ref =>
+          originator ! ContextInitialized
+        } { err =>
+          originator ! ContextInitError(err)
+        }
+      }
+
+    case GetAdHocContext(classPath, contextConfig) =>
```
zeitos: Should this be rename to StartAdhocContext?

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -57,11 +56,16 @@ object JobServer {
...
-      val jarManager = system.actorOf(Props(classOf[JarManager], jobDAO), "jar-manager")
+      val daoActor = system.actorOf(Props(classOf[JobDAOActor], jobDAO), "dao-manager")
       val dataManager = system.actorOf(Props(classOf[DataManagerActor],
           new DataFileDAO(config)), "data-manager")
-      val supervisor = system.actorOf(Props(classOf[LocalContextSupervisorActor], jobDAO),
-        "context-supervisor")
+      val jarManager = system.actorOf(Props(classOf[JarManager], daoActor), "jar-manager")
+      val contextPerJvm = config.getBoolean("spark.jobserver.context-per-jvm")
+      val supervisor = if (contextPerJvm) {
+        system.actorOf(Props(classOf[AkkaClusterSupervisorActor], daoActor), "context-supervisor")
```
zeitos: you can do 
``` system.actorOf(Props(if (contextPerJvm) classOf[AkkaClusterSupervisorActor] else classOf[AkkaClusterSupervisorActor], daoActor), "context-supervisor")```

File: job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala (hunk 20 above)
Comment (zeitos) [inline job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala]: you can do 
```scala 
contexts.retain((k, v) => v != actorRef)
```

job-server/src/spark.jobserver/JobStatusActor.scala
```diff
@@ -32,6 +33,13 @@ class JobStatusActor(jobDao: JobDAO) extends InstrumentedActor with YammerMetric
   val metricNumJobInfos = gauge("num-running-jobs", infos.size)
   val metricStatusRates = mutable.HashMap.empty[String, Meter]
 
+  override def postStop(): Unit = {
+    val stopTime = DateTime.now()
+    val stoppedInfos = infos.values.map { info =>
+      info.copy(endTime = Some(stopTime), error = Some(new Exception("Context for this job was terminated")))}
```
zeitos: Should we add more info for the user, like:
```scala
 Some(new Exception(s"Context (${info.contextName}) for this job (${info.jobId}}) was terminated")))}
```

I think the job Id might be redundant

job-server/src/spark.jobserver/io/JobDAOActor.scala
zeitos: +1

zeitos:
Great work @velvia lots of folks are waiting for this feature!
I made a couple of comments, the one that I consider more important is the one in StartContext, I don't quite get why we pass a name and then generate another one for the new process.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Oh yeah, probably

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Hmm....  right now I suppose it doesn't matter what the remote actor's name is, I think the local Supervisor will keep track of the remote actor address by context name using a map.  I suppose if we had multiple webApis, is it possible to have multiple contexts named the same thing?  That would be one disadvantage of having the actor name be the same as the context.

I think the remote JobManager knows the context name, so discovery is not an issue.  

@Ankit1010 wrote the original code, what do you think?   Also /cc @helena

job-server/src/main/resources/application.conf
velvia: We should def turn this back on at some point.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Two of the four are used only for callbacks once a forked sparkContext / JobManager process connects back with the parent process that launched it.   I'll see what can be done to improve that.

velvia: The bigger problem is after a parent dies, if it re-launches can it reconnect with one of the workers that dies?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
Ankit1010: I did this way because it lets me identify when a new actor in the cluster is a JobManagerActor [here](https://github.com/spark-jobserver/spark-jobserver/blob/feature/context-per-jvm/job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala#L68) by checking the name prefix. I guess this is not strictly required right now, but in a scenario with multiple masters, or any other kind of actor, we can't assume everyone who responds to the identify is a JobManagerActor

Theres probably a better way of doing that check though - if  there is one then yeah having the same names would be clearer.

If we had multiple masters, contexts could be named the same but just prefixing the master address would be enough to get unique names there i think?

velvia: I was envisioning where multiple masters would actually share a global view of contexts such that they would route to the context named A and know what node that context resides on.   I think this corresponds to real world usage where on one cluster you typically want one context named A, but having multiple masters for HA.

Ankit1010: Oh ok, that makes sense! I wasn't sure what exactly HA means here. If there is a global view of contexts, it should not be possible to have multiple contexts with the same name (atleast after reconciling different masters views) if i understand correctly . 

Anyways, just wanted to add that even without multiple masters, the problem of identifying the JobManager actors still stands because there are other actors in the system. Maybe a more idiomatic way to do it is have the JobManagerActors be children of a "contexts" actor - so their paths would be `<actorSystem>/user/contexts/context1`, etc. and the akka supervisor will then send identify messages only to that subtree `<actorSystem>/user/contexts/*` when a new member joins.

velvia: Hmm I like the actor subtree idea.  I’ll see what I can do, after responding to other generic feedback first and cleaning up logging.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Do you mean than cluster unsubscribe?

bin/server_start.sh
velvia: Not sure, but it's a good practice.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Good call.

job-server/src/main/resources/application.conf
velvia: Thanks!

project/Dependencies.scala
velvia: Oh, hm..  good point.  Will have to investigate that, because I think kaka-cluster brings in a few other things, but maybe we can exclude akka itself.

velvia: I'll exclude akka-remote, but this will only matter for Spark 1.5.x anyways, because Akka has been kicked out of spark 1.6.  :/

velvia: Which of course means that Spark-JobServer can upgrade to Akka 2.4.x and akka-http!  :)  /cc @ktoso

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Hmm, I realize there is a bigger problem.  Not only `StartAdHocContext` but the other `StartContext` also has the same problem.  If I quickly try to start two contexts with the same name then it fails.  Somehow we need to make it atomic, yet account for failure, which is difficult ...   I think I will tag this as an issue to be fixed.   The problem also existed in the `LocalContextSupervisor` before.

velvia: Added https://github.com/spark-jobserver/spark-jobserver/issues/349

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: No, that should not be possible.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Actually, we should really just reuse the context name, which if it was atomically checked, should not exist more than once.  No need for a separate logic I think, and it will be easier to identify.  But there were other ideas from above.

velvia:
Hey guys, most of the CR has been addressed, plus I fixed a logging bug so that now you get logs in the deploy dir.  Two issues remain:

1) I want to move the initialization of JobManagerActor back to passing in most of the things in the constructor.  This will involve passing the config and other objects in the command line via temporary files instead.  The advantage of this approach is that we could also separate out logging for each context into a separate dir.
2) The forked JVMs are not killed right now when job server shuts down.

I feel like I want to address 1) right now, and file 2) as an issue to investigate. 

@zeitos @PierreVincent @Ankit1010 @noorul what do you guys think?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: @Ankit1010 I thought about this some more.  The code is like this:
```scala
        val memberActors = RootActorPath(member.address) / "user" / "*"
        context.actorSelection(memberActors) ! Identify(memberActors)
```

Thus right now actor identity is being invoked on all actors of the forked process that reports MemberUp.  This would include the JobManagerActor, JarManager, JobStatusActor, and any other actor that gets spun up.   Even if we used the subtree idea, any actors spun up by the JobManager itself would appear under the subtree.  Thus, I do believe we need not just the pure context name as the actor name, but something else to differentiate the name.  As for multiple master, we have to sketch that out much further. :)

velvia:
Folks, I've simplified the code slightly and separated out logging into a separate directory for each context.  This should help debugging tremendously.

If I don't hear back from folks, and tests pass, I will add documentation and merge this PR.  Thanks.

Ankit1010:
@velvia 
minor issue - we should handle exceptions when creating context dir/writing config. there is also a small typo there making the build fail.

Overall :+1:.  The new config stuff and per context logging looks awesome. Agree on making the forked JVM a separate issue in the interest of keeping this focused.

zeitos:
I'll take another look today.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
zeitos: Agree

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,256 @@
...
+
+  }
+
+  private def createContextDir(name: String,
+                               contextConfig: Config,
+                               isAdHoc: Boolean,
+                               actorName: String): java.io.File = {
+    // Create a temporary dir, preferably in the LOG_DIR
+    val tmpDir = Option(System.getProperty("LOG_DIR")).map { logDir =>
+      Files.createTempDirectory(Paths.get(logDir), "jobserver")
```
zeitos: Shouldn't this folder have the contextName?

job-server/src/spark.jobserver/JobManager.scala
```diff
@@ -0,0 +1,74 @@
...
+  val logger = LoggerFactory.getLogger(getClass)
+
+  // Allow custom function to create ActorSystem.  An example of why this is useful:
+  // we can have something that stores the ActorSystem so it could be shut down easily later.
+  def start(args: Array[String], makeSystem: Config => ActorSystem) {
+    val clusterAddress = AddressFromURIString.parse(args(1))
+    val workDir = args(0)
+    val contextPath = new java.io.File(workDir + "/context.conf")
+    if (!contextPath.exists()) {
+      println(s"Could not find context configuration file $contextPath")
```
zeitos: logger instead of println?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -0,0 +1,256 @@
...
+    val contextActorName = "jobManager-" + java.util.UUID.randomUUID().toString.substring(16)
+
+    logger.info("Starting context with actor name {}", contextActorName)
+
+    val contextDir = createContextDir(name, contextConfig, isAdHoc, contextActorName)
+    val pb = Process(s"$managerStartCommand $contextDir ${selfAddress.toString}")
+    val pio = new ProcessIO(_ => (),
+                        stdout => scala.io.Source.fromInputStream(stdout)
+                          .getLines.foreach(println),
+                        stderr => scala.io.Source.fromInputStream(stderr).getLines().foreach(println))
```
zeitos: logger instead of println?

```
val pio = new ProcessIO(_ => (),
                        stdout => scala.io.Source.fromInputStream(stdout)
                          .getLines.foreach(logger.info(_)),
                        stderr => scala.io.Source.fromInputStream(stderr).getLines().foreach(logger.error(_)))
```

job-server/src/spark.jobserver/JobManager.scala
zeitos: or System.err.println ?

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: Yep.  Just added it.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
velvia: That's different.  The logging is for the HTTP process's log so we know at what time the context is being forked.

velvia:
@Ankit1010 @zeitos thanks for the extra feedback.   I think it's ready for merge now.   Also generated a TOC for the README which should be quite handy.

velvia:
@Ankit1010 I filed this issue: https://github.com/spark-jobserver/spark-jobserver/issues/353 for tracking the Akka and forked JVM reconnection issues.  It's quite meaty and something we should definitely tackle before release.  Going to merge this now.

PR #350, opened 2015-12-24
Named Objects

TimMaltGermany:
Pull Request: Named Object 

The SJS currently supports NamedRDDs via the NamedRddSupport trait. Only objects of type RDD[T] can be stored and retrieved.
We would like to use the same functionality for DataFrames and possibly future similar implementations and object in general.

Proposed changes:
- introduce new traits NamedObjectSupport and NamedObjects that contain the generalized functionality of NamedRDDSupport and NamedRDDs

This is the initial pull request that is based on some eMail discussions between Tobias, Evan and myself. There are still quite a few TODOs and open questions, but I thought that this pull request might make it easier to discuss things.

The current version only has an implementation for RDDs. In fact, the old trait NamedRddSupport still exsists (but now extends NamedObjectSupport) and ensures backwards compatibility of job implementations.
An implementation that support DataFrames should follow as soon as we finalize the actual API. 
I wil add some inline comments to highlight the open issues that I am aware of.

Thanks,
Tim


job-server-api/src/spark.jobserver/NamedObjectSupport.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.spark.storage.StorageLevel
+import scala.concurrent.duration.Duration
+
+sealed trait NamedObject
+case class NamedRDD[T](rdd: RDD[T], storageLevel: StorageLevel) extends NamedObject
+case class NamedDataFrame(df: DataFrame) extends NamedObject
+
+abstract class NamedObjectPersister[O <: NamedObject] {
+  //TODO what is the context needed for? c: ContextLike,
+  def saveToContext(namedObj: O, name: String): Unit
```
TimMaltGermany: Evan proposed "c: ContextLike" as an argument. I am not sure what we would need it for and it would require access to the context at a place where it is currently not available. So, do we really need this?

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
```diff
@@ -0,0 +1,134 @@
...
+  //TODO - what is this for?
+  //def getFromContext(c: ContextLike, name: String): O
```
TimMaltGermany: I did not quite understand what this method should be good for and left it therefor out for now.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * update reference to named object so that it does not get GCed
+   */
+  def refresh(namedObject: O): O
+  //TODO - what about 'unpersist' ?
```
TimMaltGermany: The old implementation called unpersist after removing a named rdd from the reference list. Is that still something that the cache implementation should handle?

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -0,0 +1,121 @@
...
+
+    val res : Option[O] = getExistingObject(name)
+    res match {
+      case Some(namedObject) =>
+        namedObject
+      case None =>
+        logger.info("Named object [{}] not found, starting creation", name)
+        val f = cachedOp(name, createObject(objGen, name))
+        f.await match {
+          case obj : O => obj
```
TimMaltGermany: this is really awkward - does anybody have a better idea?

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  // we can wrap the operation with caching support
+  // (providing a caching key)
+  private def cachedOp[O <: NamedObject](name: String, f: () => O):
+                           Future[NamedObject] = namesToObjects(name) {
+    f()
+  }
+
+  private def createObject[O <: NamedObject](objGen: => O, name: String)
```
TimMaltGermany: it would be possible to combine cachedOp and createObject into a single method, but I thought this is more readable

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -0,0 +1,121 @@
...
+        namedObj
+      }
+  }
+
+  private def getExistingObject[O](name: String): Option[O] = {
+    //get:  retrieves the future instance that is currently in the cache for the given key.
+    // Returns None if the key has no corresponding cache entry.
+    namesToObjects.get(name) match {
+      case Some(f : Future[O]) =>
+        Some(f.await)
```
TimMaltGermany: The rdd-implementation checked here whether the rdd still exists in the SparkContext.

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -0,0 +1,121 @@
...
+  override def get(name: String)(implicit timeout: Timeout = defaultTimeout): Option[NamedObject] = {
+    getExistingObject(name) match {
+      case objectOpt: Option[NamedObject] @unchecked => objectOpt //TODO  rddOpt.map { rdd=> refresh(rdd) }
+    }
+  }
+
+  //TODO - isn't this identical to getOrElseCreate?
+  // only that the object may already exist and we need to remove it - unless the old
+  // object (rdd) has the same id TODO - is that even possible? The API does not
+  // mention that special case either
```
TimMaltGermany: The original code did remove the old rdd only iff its id was different from the id of the new rdd. The API does not mention this.

project/Dependencies.scala
```diff
@@ -33,6 +34,7 @@ object Dependencies {
   val sparkVersion = sys.env.getOrElse("SPARK_VERSION", "1.5.2")
   lazy val sparkDeps = Seq(
     "org.apache.spark" %% "spark-core" % sparkVersion % "provided" excludeAll(excludeNettyIo, excludeQQ),
+    "org.apache.spark" %% "spark-sql" % sparkVersion % "provided" excludeAll(excludeNettyIo, excludeQQ),
```
TimMaltGermany: moved spark-sql from extra deps as it is now required in job-server-api as well

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
```diff
@@ -0,0 +1,134 @@
...
+ * If two jobs simultaneously tries to create a data object with the same name and in the same namespace,
+ * only one will win and the other will retrieve the same one.
+ *
+ * Note that to take advantage of NamedObjectSupport, a job must mix this in and use the APIs here instead of
+ * the native DataFrame/RDD `cache()`, otherwise we will not know about the names.
+ */
+trait NamedObjects {
+  // Default timeout is 60 seconds. Hopefully that is enough
+  // to let most RDD/DataFrame generator functions finish.
+  val defaultTimeout = akka.util.Timeout(Duration(60, java.util.concurrent.TimeUnit.SECONDS))
```
zeitos: Or we can make it a config / environment variable.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
```diff
@@ -0,0 +1,134 @@
...
+   * to compute it, in the caller's thread, and the result will be cached and returned to the caller.
+   *
+   * If an NObj is requested by thread B while thread A is generating the NObj, thread B will block up to
+   * the duration specified by @timeout. If thread A finishes generating the NObj within that time, then
+   * thread B will get a reference to the newly-created RDD. If thread A does not finish generating the
+   * NObj within that time, then thread B will throw a timeout exception.
+   *
+   * @param name the unique name of the NObj. The uniqueness is scoped to the current SparkContext.
+   * @param rddGen a 0-ary function which will generate the NObj if it doesn't already exist.
+   * @param timeout if the named object isn't created within this timeout, an error will be thrown.
```
zeitos: we need doc for persister

job-server-api/src/spark.jobserver/NamedRddSupport.scala
```diff
@@ -6,115 +6,191 @@ import java.util.concurrent.atomic.AtomicReference
...
-   * @param name the unique name of the RDD. The uniqueness is scoped to the current SparkContext.
-   * @param timeout if the RddManager doesn't respond within this timeout, an error will be thrown.
-   * @tparam T the generic type of the RDD.
-   * @return the RDD with the given name.
-   * @throws java.util.concurrent.TimeoutException if the request to the RddManager times out.
-   */
-  def get[T](name: String)(implicit timeout: Timeout = defaultTimeout): Option[RDD[T]]
+    // Default timeout is 60 seconds. Hopefully that is enough
+    // to let most RDD/DataFrame generator functions finish.
+    val defaultTimeout = akka.util.Timeout(Duration(60, java.util.concurrent.TimeUnit.SECONDS))
```
zeitos: I guess you don't need this one since you have it already in NamedObjectSupport

job-server-api/src/spark.jobserver/NamedRddSupport.scala
```diff
@@ -6,115 +6,191 @@ import java.util.concurrent.atomic.AtomicReference
...
+    }
+
+//    override def getFromContext(c: ContextLike, name: String): NamedRDD[T] = {
+//      c.sparkContext.getPersistentRDDs.filter(_._2.name.equals(name)).map(_._2).headOption match {
+//        case Some(rdd: RDD[T]) => NamedRDD(rdd, rdd.getStorageLevel)
+//      }
+//    }
+
+    /**
+     * Calls rdd.persist(), which updates the RDD's cached timestamp, meaning it won't get
```
zeitos: I *think* this behaviour changed in spark, unless isLocallyCheckpointed is true persist() would not have any effect

job-server/src/spark.jobserver/JobManagerActor.scala
```diff
@@ -230,10 +224,10 @@ class JobManagerActor(dao: JobDAO,
...
       Thread.currentThread.setContextClassLoader(jarLoader)
       val job = constructor()
-      if (job.isInstanceOf[NamedRddSupport]) {
-        val namedRdds = job.asInstanceOf[NamedRddSupport].namedRddsPrivate
-        if (namedRdds.get() == null) {
-          namedRdds.compareAndSet(null, new JobServerNamedRdds(rddManagerActor))
+      if (job.isInstanceOf[NamedObjectSupport]) {
+        val namedObjects = job.asInstanceOf[NamedObjectSupport].namedObjectsPrivate
+        if (namedObjects.get() == null) {
+          namedObjects.compareAndSet(null, new JobServerNamedObjects(context.system))
```
zeitos: you don't need to check for null before doing compareAndSet since that's exactly what compareAndSet will do internally.

job-server/src/spark.jobserver/JobServerNamedObjects.scala
zeitos: I don't see a way out, typeerasure will prevent you from doing some funky thing with O type .

The one thing I don't like (although it's not a great deal) is that you won't be able to have 2 objects with the same name .

job-server/src/spark.jobserver/JobServerNamedObjects.scala
zeitos: Also: there's a problem here if we have 2 jobs that use the same name for different type of objects, one will end up with  not only another object but an instance from another type (it will actually throw an exception in the getOrElseCreate

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -0,0 +1,121 @@
...
+    val res : Option[O] = getExistingObject(name)
+    res match {
+      case Some(namedObject) =>
+        namedObject
+      case None =>
+        logger.info("Named object [{}] not found, starting creation", name)
+        val f = cachedOp(name, createObject(objGen, name))
+        f.await match {
+          case obj : O => obj
+          case NamedRDD(_, _) => throw new IllegalArgumentException("Incorrect type for named object")
```
zeitos: we should log the type of this object.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
velvia: Hmm.  `ContextLike` should be available from a `SparkJobBase`, because this method would be called from the `runJob` method, which is passed the `ContextLike`.  But need to look in detail.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
velvia: Hm, I was thinking for certain classes of objects you could retrieve it directly from the context. For example, for DataFrames, it would be defined like this:
```scala
  def getFromContext(c: ContextLike, name: String): O = {
    val df = c.asInstanceOf[SQLContext].table(name)
    DataFrameObject(df)
  }
```

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
velvia: No, the cache implementation should know nothing about what is cached.  Instead, the RDDPersister (whatever implements NamedObjectPersister for RDD[T]) should handle that.

project/Dependencies.scala
velvia: Why? Is it to include DataFrames?

Instead of pulling in spark-sql, I would say put the DataFrames logic in job-server-extras.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
velvia: +1

job-server/src/spark.jobserver/JobManagerActor.scala
velvia: I'd like to see if we could do a slightly less hacky version of this.  If everyone uses NamedObjectSupport then no need to mix it in.  Maybe put the logic directly in ContextLike.

velvia:
Will have a more in depth look at the code later, I'm sure will have lots of suggestions.

job-server-api/src/spark.jobserver/NamedRddSupport.scala
```diff
@@ -6,115 +6,191 @@ import java.util.concurrent.atomic.AtomicReference
...
+     * that the data returned from this method may be stale and incorrect.
+     *
+     * @param timeout if the RddManager doesn't respond within this timeout, an error will be thrown.
+     * @return a collection of RDD names representing RDDs managed by the RddManager.
+     */
+    def getNames()(implicit timeout: Timeout = defaultTimeout): Iterable[String]
+  }
+
+  private def rddPersister[T](forceComputation: Boolean) = new NamedObjectPersister[NamedRDD[T]] {
+    override def saveToContext(namedObj: NamedRDD[T], name: String) {
```
velvia: Ok, the reason you don't need the SparkContext here is because RDDs internally already contain a reference to the RDD.  Dunno fi that's true of everything else that the user wants persisted.

job-server/src/spark.jobserver/JobManagerActor.scala
```diff
@@ -120,7 +117,7 @@ class JobManagerActor(dao: JobDAO,
       }
 
     case StartJob(appName, classPath, jobConfig, events) =>
-      startJobInternal(appName, classPath, jobConfig, events, jobContext, sparkEnv, rddManagerActor)
```
velvia: Yes, so happy to be getting rid of rddManagerActor.

job-server/src/spark.jobserver/JobServerNamedObjects.scala
velvia: Could we have `cachedOp` just return `NamedObject` instead of `O`?

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -0,0 +1,121 @@
...
+  // we must store a reference to each NamedObject even though only its ID is used here
+  // this reference prevents the object from being GCed and cleaned by sparks ContextCleaner
+  // or some other GC for other types of objects
+  private val namesToObjects: Cache[NamedObject] = LruCache()
+
+  override def getOrElseCreate[O <: NamedObject](name: String, objGen: => O)
+                                (implicit timeout: Timeout = defaultTimeout,
+                                    persister: NamedObjectPersister[O]): O = {
+
+    val res : Option[O] = getExistingObject(name)
```
velvia: I think all this logic is unnecessary.  You are trying to see if the object exists in the cache, then creating it, but your logic is not atomic, whereas the LRUCache already has an API which does exactly this, and is atomic.  Basically you could replace the body of this whole function with:

    val fut = namesToObjects(name) {
      logger.info("Named object [{}] not found, starting creation", name)
      Future(objGen)
    }
    Await.result(fut, timeout)

velvia:
@TimMaltGermany I love that you have tests.  Looked through the code.  A good start, I think the JobServerNamedObjects implementation could be much simpler, and made some suggestions.

TimMaltGermany:
Thanks guys for the comments. I will work on it on Thursday and Friday this week.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
TimMaltGermany: Hmm, the NamedObjectPersister is currently not informed when an object is destroyed. I moved the unpersist into the destroy method of the _NamedRdds implementation, but I think it might be better if we added a destroy (or removeFromContext) method to NameObjectPersister?

job-server/src/spark.jobserver/JobServerNamedObjects.scala
TimMaltGermany: Well, the problem is that chachedOp does return `NamedObject` here, but getOrElseCreate needs to return an `O`. (In fact, `get` also needs to return `O`).

job-server/src/spark.jobserver/JobServerNamedObjects.scala
TimMaltGermany: Good point. We could use name prefixes. I can easily add this to the new implementation of NamedRDDSupport, but it would be better if we could somehow enforce this implicitly. Ideas?

project/Dependencies.scala
TimMaltGermany: ok, done

job-server/src/spark.jobserver/JobServerNamedObjects.scala
zeitos: I wouldn't change the name for the nameobject since it's currently used to store the rdd, what we can do is convert namedObject to a mutable.Map (I'm not sure what's the best map in scala for concurrency, there's a ConcurrentMap but has been deprecated in favor of mutable.Map since 2.10) that maps each type to an LRU

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
TimMaltGermany: ok, introduced `spark.jobserver.named-object-creation-timeout`
However, to get this to work I had to change the timeout type to `Option[Timeout]` as in: 
    def getOrElseCreate[O <: NamedObject](name: String,
                         objGen: => O)(implicit timeout: Option[Timeout] = None,
                                       persister: NamedObjectPersister[O]): O

job-server-api/src/spark.jobserver/NamedRddSupport.scala
TimMaltGermany: done, but again, to make this work, I had to change the type of the timeout parameter to `Option[Timeout]`

job-server-api/src/spark.jobserver/NamedRddSupport.scala
TimMaltGermany: hmm, can we ignore this in this PR? After all, this is the current SJS behavior.

job-server/src/spark.jobserver/JobServerNamedObjects.scala
TimMaltGermany: tricky - how would I do that given that the type of O is erased?

job-server/src/spark.jobserver/JobServerNamedObjects.scala
TimMaltGermany: I simplified it to:

    cachedOp(name, createObject(objGen, name)).await(timeout) match {
      case obj: O @unchecked => obj
      case NamedRDD(_, _)    => throw new IllegalArgumentException("Incorrect type for named object")
    }

TimMaltGermany:
Tomorrow, I will implement a NamedObjectPersister for DataFrames to see how far we get with the current API. A bit of documentation is also still missing and an update of the Readme.md.

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
```diff
@@ -0,0 +1,150 @@
...
+  //def getFromContext(c: ContextLike, name: String): O
+
+  /**
+   * update reference to named object so that it does not get GCed
+   */
+  def refresh(namedObject: O): O
+
+  //TODO - what about 'unpersist' / removeFromContext / destroy?
+  //proposal:
+  // def removeFromContext(c: ContextLike, name: String): Unit
```
TimMaltGermany: I do believe that we need a `destroy` method so that the persister knowns when to unpersist the named object.

TimMaltGermany:
Hi, I have added an implementation for DataFrames. Please check out the tests in: `job-server-extras/test/spark.jobserver/NamedObjectsSpec.scala`

There are still a few loose ends: 
- When would we need the getFromContext method? I still don't quite see the use case.
- None of the current two persister implementations require the ContextLike argument. 
- Documentation needs an update.

I would now welcome further feedback or should we schedule a Hangout?
Cheers.

job-server/src/spark.jobserver/JobServerNamedObjects.scala
zeitos: yes, O is erased but an instance belongs to a type and you can always get the type of an instance with classOf[] of course you'll have to deal with nulls.
Or you can add an implicit manifest and do (manifest.erasure.getName)

TimMaltGermany: true, but in that case we can log it only after the object was created, I was hoping we could simply add the type to the existing log message

velvia:
@TimMaltGermany will probably be early next week when I can have a look.  open to hangout.  cheers.

README.md
```diff
@@ -1,582 +1,629 @@
-[![Build Status](https://travis-ci.org/spark-jobserver/spark-jobserver.svg?branch=master)](https://travis-ci.org/spark-jobserver/spark-jobserver)
```
TimMaltGermany: sorry about the diff, I tried both the Windows and Unix version of the newlines, but it seems that both are different from Master....

job-server-api/src/spark.jobserver/NamedObjectSupport.scala
TimMaltGermany: done

job-server-api/src/spark.jobserver/NamedRddSupport.scala
```diff
@@ -1,120 +1,206 @@
...
    */
-  def destroy(name: String): Unit
+  trait _NamedRdds {
 
-  /**
-   * Returns the names of all named RDDs that are managed by the RddManager.
-   *
-   * Note: this returns a snapshot of RDD names at one point in time. The caller should always expect
-   * that the data returned from this method may be stale and incorrect.
+    val defaultTimeout = Timeout(Duration(60, java.util.concurrent.TimeUnit.SECONDS))
```
TimMaltGermany: Still using the hard-coded timeout here for backwards compatibility and since the trait cannot access the config.

job-server/src/spark.jobserver/JobManagerActor.scala
TimMaltGermany: I think we should handle this in a different PR.

TimMaltGermany:
Hi,
I have updated the PR as discussed this morning with @velvia with a couple of additional changes:     
- the NamedObjects trait now has a a `destroy` and a `forget` method. `destroy` removes the object from the cache and calls the persister to unpersist the object, while `forget` only removes it from the cache. 
- the default timeout remains unchanged as long as people stick to the old NamedRddSupport, it is only read from the config when then new NamedObjectsSupport is used.

Cheers,
Tim

job-server-api/src/spark.jobserver/NamedRddSupport.scala
velvia: You can just define something like `def config` as an abstract variable in the Trait  to be able to access the config, then in the implementation pass it in the constructor.

TimMaltGermany:
Hi, I have removed the hard-coded timeout setting from the old NamedRddSupport as well and merge the last master into this branch.
Are we good to go now?
Cheers,
Tim

velvia:
I’ll have a look shortly.  Think we’re really really close.

velvia:
@TimMaltGermany many thanks.  Just a few cosmetic changes.  Think we're ready to go.

zeitos:
Looks good to me! Great job @TimMaltGermany !

TimMaltGermany:
Thanks for the feedback. I fixed the style issues. Ready to merge....

velvia:
@TimMaltGermany awesome job!  Merging.

PR #356, opened 2016-01-08
Build for  Spark 1.6.0

RaafatAkkad:
Compiles for Spark 1.6.0. Changed SparkContext instantiation for some classes to remove the following warning: "constructor SparkContext in class SparkContext is deprecated: Passing in preferred locations has no effect at all, see SPARK-10921"


velvia:
@RaafatAkkad thanks, this overall looks pretty good to me.  Let me think about timing.

LHaferkamp:
You should also change the netty version in Dependencies.scala
from
"io.netty" % "netty-all" % "4.0.23.Final" to 4.0.29.Final
as the new netty version is used in Spark 1.6 and the old one causes memory leaks

RaafatAkkad:
@LHaferkamp Thanks for the feedback.

Netty 4.0.29.Final has been a dependency since Spark 1.5.0 (https://github.com/apache/spark/commit/26ed22aec8af42c6dc161e0a2827a4235a49a9a4). As this is not specific to Spark 1.6.0, I think it's worth making the change and submitting a separate pull request.

velvia:
@RaafatAkkad if you don't mind you can go ahead and just make the change here for convenience.

Issue #360, opened 2016-01-15
job id on UI should link to status endpoint


h0ke:
On our fork at Preact, I've linked the job id to the status endpoint. Also, I've added a small link the config endpoint (see screenshot). Is this something you would be interested in? I can get the PR prepped pretty quickly.

![screen shot 2016-01-15 at 8 42 16 am](https://cloud.githubusercontent.com/assets/181554/12358860/2f1b755c-bb64-11e5-9809-b7818911120a.png)

velvia:
Yes definitely!

-Evan
"Never doubt that a small group of thoughtful, committed citizens can change the world" - M. Mead

PR #361, opened 2016-01-15
add links to UI to status and config endpoints for each job. 

h0ke:
add link to jobid on ui that points to status endpoint
add link after jobid on ui that points to config endpoint
issue #360

### DIFF
--- a/job-server/src/main/resources/html/js/spark-jobserver-ui.js
+++ b/job-server/src/main/resources/html/js/spark-jobserver-ui.js
@@ -10,7 +10,7 @@ function getJobs() {
       $.each(jobs, function(key, job) {
         var items = [];
         items.push("<tr>");
-        items.push("<td>" + job.jobId + "</td>");
+        items.push("<td><a href='./jobs/" + job.jobId + "'>" + job.jobId + "</a> (<a href='./jobs/" + job.jobId + "/config'>C</a>)</td>");
         items.push("<td>" + job.classPath + "</td>");
         items.push("<td>" + job.context + "</td>");
         items.push("<td>" + job.startTime + "</td>");


job-server/src/main/resources/html/js/spark-jobserver-ui.js
```diff
@@ -10,7 +10,7 @@ function getJobs() {
       $.each(jobs, function(key, job) {
         var items = [];
         items.push("<tr>");
-        items.push("<td>" + job.jobId + "</td>");
+        items.push("<td><a href='./jobs/" + job.jobId + "'>" + job.jobId + "</a> (<a href='./jobs/" + job.jobId + "/config'>C</a>)</td>");
```
noorul: C? Can we use better name like Config?

noorul: Can you attach a screenshot?

velvia:
@h0ke LGTM.  I think changing to Config is slightly preferable, but not that huge of a deal.

job-server/src/main/resources/html/js/spark-jobserver-ui.js
h0ke: I defaulted to `C` since a wall of `Config` seemed overwhelming to me. But I can change it, please let me know either way.

screen shot 2016-01-15 at 8 42 16 am

velvia: I think C is OK with me.  @noorul if you have no objections will merge this soon.

noorul:
@h0ke Merged! Thank you!

h0ke:
Glad to help!

PR #373, opened 2016-01-26
Enable per job dependencies management

amarouni:
This is WIP.

This PR would allow the following : 
````
curl 'localhost:8090/jobs?appName=j01&classPath=Job01&sync=true&context=ctx01' -d '{
dependent-jar-uris = ["file:////deps.jar", "file:///deps2.jar"]}'
````
We'll be able to load deps.jar and deps2.jar for this particular job and use the dependencies both in the spark driver and the workers at job's runtime. 

TODO/discuss :
1- Once the jars are added they cannot be removed (as per the Spark context sparkContext.addjar("file"))
2- If the job (SparkJob as implemented by classpath value to be precise) that requires runtime dependencies is executed without these dependencies we'll get a ClassNotFound and that's expected. However if we submit the job (same classpath) again (on the same context) with the correct dependencies the job will still fail with the same ClassNotFound, as if there's some cached version of the ClassLoader that doesn't see the new dependencies.

### CODE REVIEW

**Comment by velvia** [comment]
@amarouni Thanks!  Some thoughts:

For #2 - you are saying that this basically doesn't work yet?

- It'd be nice to write a test, but won't be easy.
- Please add some docs to the README (when it all works).

Thanks again!

**Comment by amarouni** [comment]
@velvia Thanks for the quick review. I'll fix that and I'll make sure that the final version includes some tests and some docs.

Now for #2 I was hoping to get some help from you as you're more familiar with the code base.
From what I saw it looks like once the class (classPath) is loaded (from cache or a new instance) its class loader is created and cannot be changed anymore. Even if I disabled the cache the class's classloader won't be able to find the runtime dependecies.

Thanks !

**Comment by velvia** [comment]
Hmmm yeah that’s really weird.   The class loader is supposed to let you add more URLs dynamically.  However some things can hinder class discovery:

1) If the class is already loaded in the JVM somewhere else, then it will use that
2) Maybe the URL already loaded

Try modifying or adding a method to the class loader to print out the URLs, it might already have one, for debugging.

As long as you’re not creating a new instance of the classloader, I think addURL should dynamically change it.

-Evan

**Comment by velvia** [comment]
@amarouni looking at your error above, it seems in the second case it is looking for `TalendUtils/Utils1$`, which is not the same as `Utils/Utils1$`.  Did something change in your code?

**Comment by amarouni** [comment]
@velvia It's a typo ! Don't worry about that.
I'll look at what you proposed in your previous comment about the class loader and the cached classes.

Thanks!

**Comment by amarouni** [comment]
@velvia, 
I looked further at the issue and did a couple of tests :
I created a job with :
````scala
val lib = getClass.getClassLoader.loadClass(config.getString("libName"))
println(lib.getName)
````
And executed the following jobs :
````bash
curl 'localhost:8090/jobs?appName=j01&classPath=Job03&context=ctx01&sync=true' -d '{
dependent-jar-uris = "file:////depsUtils1.jar",
 	    libName = "TalendUtils.Utils1"}'

curl 'localhost:8090/jobs?appName=j01&classPath=Job03&context=ctx01&sync=true' -d '{
dependent-jar-uris = "file:////depsUtils2.jar",
 	    libName = "TalendUtils.Utils2"}'
}'
````

The jobs were executed successfully and I got the correct class name printed in the logs, so I was sure that between consecutive executions the ClassLoader is available for additions even after the class was first instantiated.

What I also tested is printing the URLs in the ClassLoader to see if there's something wrong. I was able to see all of the required jars urls so that would eliminate the possibility of missing URL from the class loader.

What I saw also was that the failed jobs (missing dependencies) were stuck in the Running Jobs on the SJS UI. As far as I know the jobs should be cleaned when they fail, so maybe there's something missing and they get stuck and are impossible to run again.

Hope this helps

**Comment by velvia** [comment]
@amarouni, thanks for all the investigation.   It seems unclear whether the failures had to do with the class loader / new jars or not.  They seem similar to some other observed hangs.

**Comment by amarouni** [comment]
@velvia, I did some more tests on an unmodified clone of the project and I found out that if the "dependent-jar-uris" defined in the config file used to start SJS in dev mode is missing some jars needed to run jobs then the jobs would hang forever with 
````java
java.lang.NoClassDefFoundError
````
and they become impossible to run again. So I think that "Class Not Found Exception" is another issue that's not related to the PR.

So I'll clean the PR add some tests and documentation and open an issue for the jobs hanging with missing dependencies.

**Comment by velvia** [comment]
Thanks.  Looking forward to it.

**Comment by amarouni** [comment]
@velvia I added some documentation and a test case. I'm also aware that the test case uses a hardcoded path, but I'm hoping to get some help or a workaround.

**Comment by velvia** [comment]
Thanks.  Probably won’t be able to have a look until later this week…. @zeitos?

**Comment by amarouni** [comment]
Yup,  as I expected the hardcoded path to the extras jar had the scala version in it so it failed the scala-2.11 test

File: README.md [hunk 1]
```diff
@@ -290,6 +290,19 @@ You have a couple options to package and upload dependency jars.
     - Use the `dependent-jar-uris` context configuration param.  Then the jar gets loaded for every job
     - Use the `--package` option with Maven coordinates with `server_start.sh`.
     - Put the extra jars in the SPARK_CLASSPATH
+    - Use the `dependent-jar-uris` job configuration param when submitting a job. On an ad-hoc context this has the same effect as `dependent-jar-uris` context configuration param. On a persistent context the jars will be loaded for the current job and then for every job that will be executed on the persistent context.
```
zeitos: can you group this together with what's in line 290?

job-server/test/spark.jobserver/JobManagerSpec.scala
```diff
@@ -208,5 +210,26 @@ abstract class JobManagerSpec extends JobSpecBase(JobManagerSpec.getNewSystem) {
       }
     }
 
+    it("should be able to start a job with job jar dependencies and return results"){
+      manager ! JobManagerActor.Initialize(daoActor, None)
+      expectMsgClass(initMsgWait, classOf[JobManagerActor.Initialized])
+
+      uploadTestJar()
+      val extrasJarPath = Paths.get(".").toAbsolutePath().normalize().toString() +
```
zeitos: why not doing Paths.get("..") ?

job-server/test/spark.jobserver/JobManagerSpec.scala
```diff
@@ -208,5 +210,26 @@ abstract class JobManagerSpec extends JobSpecBase(JobManagerSpec.getNewSystem) {
...
+        "/../job-server-extras/target/scala-2.10/job-server-extras_2.10-0.6.2-SNAPSHOT.jar"
```
zeitos: this won't work with scala 2.11 tests :D consider using scala.util.Properties.versionString (you'll need to remove "version ")

zeitos:
there's a minor problem with the test that make the code fail.

job-server/src/spark.jobserver/JobManagerActor.scala
```diff
@@ -124,8 +124,14 @@ class JobManagerActor(contextConfig: Config) extends InstrumentedActor {
           self ! PoisonPill
       }
 
-    case StartJob(appName, classPath, jobConfig, events) =>
+    case StartJob(appName, classPath, jobConfig, events) => {
+      getSideJars(jobConfig).foreach { jarUri =>
+        logger.info("Adding {} to Current Job Class path", jarUri)
+        jarLoader.addURL(new URL(convertJarUriSparkToJava(jarUri)))
+        jobContext.sparkContext.addJar(jarUri)
```
zeitos: is there any way we can verify that the jar is not already loaded?

job-server/test/spark.jobserver/JobManagerSpec.scala
velvia: @amarouni can you try using the `TestJarFinder` class?  It is designed to specifically work around this absolute path problem.  Pretty sure it (or a slight modification thereof) should help get around this problem.

amarouni: @velvia thanks,  I'll take a look at that !

README.md
amarouni: @zeitos No the jars should be present on the SJS node only.

zeitos:
@amarouni  can you sync your code with master so tests will hopefully pass?

amarouni:
@zeitos my code is in sync with master. The "routes with timeout" test is failing on scala 2.11.6. I don't have this issue myself.
Maybe @velvia can restart the build.

zeitos:
@amarouni I restarted the build. @velvia merged a PR yesterday that seems to fix the build that's why I mentioned to sync with master

zeitos:
Test Passed, merging, thanks @amarouni !!

PR #380, opened 2016-02-03
Fix config collision in spark master provider

noorul:
Currently spark master provider configuration property is
spark.master.provider, which is colliding with spark.master which is
used for spark master URL.


noorul:
@velvia Not sure why travis is failing

velvia:
Not sure either.  May need to increase some timeouts.  If it’s restarted then it will pass.

noorul:
I restated several times

noorul:
At last

velvia:
@noorul looks fine.  I wonder if `spark.master-provider` would be more obvious than `ha.provider` though, and still won't collide.

PR #384, opened 2016-02-09
Update Dependencies to use scala-library 2.10.5

apivovarov:
Spark-1.6.0 uses Scala 2.10.5
https://github.com/apache/spark/blob/branch-1.6/pom.xml#L166
Dependencies should be in sync with Spark


velvia:
@apivovarov thanks

PR #385, opened 2016-02-10
Redirect /dev/null to stdin in server_start.sh

apivovarov:
Currently server_start.sh hang at the end if run it inside another script via ssh
if we run cmd on background we need to redirect /dev/null to stdin

Redirecting /dev/null to stdin will give an immediate EOF to any read call from that process. This is typically useful to detach a process from a tty (such a process is called a daemon). For example, when starting a background process remotely over ssh, you must redirect stdin to prevent the process waiting for local input.

Another reason to redirect to /dev/null is to prevent an unused file descriptor being created for stdin. This can minimize the total open file handles when you have many long running processes.


apivovarov:
CI says

apivovarov:
I checked how spark internal scripts call spark-submit. Yes, they redirect /dev/null to stdin.
https://github.com/apache/spark/blob/master/sbin/spark-daemon.sh#L154
so, this PR is doing the same as spark

velvia:
@apivovarov thanks.  Yeah something is up with Travis CI lately.   :/   Will go ahead and merge.

Issue #147, opened 2015-05-08
Fail to run two jobs simultaneously in two contexts


ningjun:
I use spark-jobserver 0.4.1 with spark1.3.1  stand alone cluster. I precreate two contexts and then submit two jobs (each job to a different contexts) to the jobserver. It always failed with the following error. Note that I can run each job alone successfully.

Please advise

RN TaskSetManager: Lost task 0.0 in stage 0.0 (TID 0, LAB4-WIN01.mycompany.com): java.io.IOException: org.apache.spark.SparkException: Failed to get broadcast_1_piece0 of broadcast_1
	at org.apache.spark.util.Utils$.tryOrIOException(Utils.scala:1156)
	at org.apache.spark.broadcast.TorrentBroadcast.readBroadcastBlock(TorrentBroadcast.scala:164)
	at org.apache.spark.broadcast.TorrentBroadcast._value$lzycompute(TorrentBroadcast.scala:64)
	at org.apache.spark.broadcast.TorrentBroadcast._value(TorrentBroadcast.scala:64)
	at org.apache.spark.broadcast.TorrentBroadcast.getValue(TorrentBroadcast.scala:87)
	at org.apache.spark.broadcast.Broadcast.value(Broadcast.scala:70)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:58)
	at org.apache.spark.scheduler.Task.run(Task.scala:64)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:203)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.apache.spark.SparkException: Failed to get broadcast_1_piece0 of broadcast_1
	at org.apache.spark.broadcast.TorrentBroadcast$$anonfun$org$apache$spark$broadcast$TorrentBroadcast$$readBlocks$1$$anonfun$2.apply(TorrentBroadcast.scala:137)
	at org.apache.spark.broadcast.TorrentBroadcast$$anonfun$org$apache$spark$broadcast$TorrentBroadcast$$readBlocks$1$$anonfun$2.apply(TorrentBroadcast.scala:137)
	at scala.Option.getOrElse(Option.scala:120)
[...]

PR #390, opened 2016-02-22
Allow larger job results to be returned by default

velvia:
See #147 for more details.  Basically this allows larger job results to be returned by default.  It is a short term solution;  the longer term solution would be to chunk and pre-serialize the result.   pre-serializing might be a good solution for storing results in a database anyhow.

### DIFF
--- a/doc/troubleshooting.md
+++ b/doc/troubleshooting.md
@@ -29,6 +29,12 @@ Then simply restart the job server.
 
 Note that the idle-timeout must be higher than request-timeout, or Spray and the job server won't start.
 
+## Timeout getting large job results
+
+If your job returns a large job result, it may exceed Akka's maximum network message frame size, in which case the result is dropped and you may get a network timeout.  Change the following configuration, which defaults to 10 MB:
+
+    akka.remote.netty.tcp.maximum-frame-size = 100 MiB
+
 ## AskTimeout when starting job server or contexts
 
 If you are loading large jars or dependent jars, either at startup or when creating a large context, the database such as H2 may take a really long time to write those bytes to disk.  You need to adjust the context timeout setting:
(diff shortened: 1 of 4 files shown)


velvia:
@noorul @zeitos?

doc/troubleshooting.md
```diff
@@ -29,6 +29,12 @@ Then simply restart the job server.
 
 Note that the idle-timeout must be higher than request-timeout, or Spray and the job server won't start.
 
+## Timeout getting large job results
+
+If your job returns a large job result, it may exceed Akka's maximum network message frame size, in which case the result is dropped and you may get a network timeout.  Change the following configuration, which defaults to 10 MB:
+
+    akka.remote.netty.tcp.maximum-frame-size = 100 MiB
```
noorul: Is MiB == MB?

velvia: MB = 1024 * 1024;  MiB = 1000000

zeitos: Actually... it's the other way around, 1 MiB is 2^20
https://en.wikipedia.org/wiki/Mebibyte

zeitos: nit: it should say 10 MiB

zeitos:
looks good, 10MiB sounds reasonable

doc/troubleshooting.md
noorul: Thank you for the explanation

noorul:
LGTM

doc/troubleshooting.md
velvia: @zeitos no I wanted to make a point how to make the setting bigger than the default.

zeitos: what I'm saying is that it says "which defaults to 10 MB:" instead of "which defaults to 10 MiB:"

velvia: Got it.  Will add the i and merge.  :)

PR #392, opened 2016-03-05
NamedObjects: cache with references to named objects must be a singleton

TimMaltGermany:
Fixes problem that references to named objects were lost inbetween jobs.

Sorry about the delay infixing this !

### DIFF
--- a/job-server-extras/test/spark.jobserver/NamedObjectsJobSpec.scala
+++ b/job-server-extras/test/spark.jobserver/NamedObjectsJobSpec.scala
@@ -10,42 +10,50 @@ import collection.JavaConversions._
 
 class NamedObjectsJobSpec extends JobSpecBase(JobManagerSpec.getNewSystem) {
 
-  private val emptyConfig = ConfigFactory.parseString("spark.jobserver.named-object-creation-timeout = 60 s")
+  //private val emptyConfig = ConfigFactory.parseString("spark.jobserver.named-object-creation-timeout = 60 s")
 
   before {
     dao = new InMemoryDAO
     daoActor = system.actorOf(JobDAOActor.props(dao))
     manager = system.actorOf(JobManagerActor.props(JobManagerSpec.getContextConfig(adhoc = false)))
     supervisor = TestProbe().ref
+
+    manager ! JobManagerActor.Initialize(daoActor, None)
+    
+    expectMsgClass(classOf[JobManagerActor.Initialized])
+
+    uploadTestJar()
+
   }
 
   val jobName = "spark.jobserver.NamedObjectsTestJob"
 
+  private def getCreateConfig(createDF: Boolean, createRDD: Boolean) : Config = {
+    ConfigFactory.parseString("spark.jobserver.named-object-creation-timeout = 60 s, " + 
+        NamedObjectsTestJobConfig.CREATE_DF + " = " + createDF + ", " +
+        NamedObjectsTestJobConfig.CREATE_RDD + " = " + createRDD)
+  }
+  
+  private def getDeleteConfig(names: List[String]) : Config = {
+    ConfigFactory.parseString("spark.jobserver.named-object-creation-timeout = 60 s, " + 
+        NamedObjectsTestJobConfig.DELETE+" = [" + names.mkString(", ") + "]")
+  }
+  
   describe("NamedObjects (RDD)") {
     it("should survive from one job to another one") {
 
-      manager ! JobManagerActor.Initialize(daoActor, None)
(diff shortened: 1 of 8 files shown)


TimMaltGermany:
hmm, the test failure is  a bit odd
WebApiWithAuthenticationSpec:
> - jobs should not allow user with valid authorization when timeout *** FAILED ***
>  200 OK was not equal to 500 Internal Server Error (WebApiWithAuthenticationSpec.scala:164)
I don't think that this is related to my changes. Could it be that this test is flaky?

TimMaltGermany:
This time it only failed for scala 2.10.4.

job-server-extras/test/spark.jobserver/NamedObjectsJobSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+  val jobName = "spark.jobserver.NamedObjectsTestJob"
+
+  describe("NamedObjects (RDD)") {
+    it("should survive from one job to another one") {
+
+      manager ! JobManagerActor.Initialize(daoActor, None)
+      expectMsgClass(classOf[JobManagerActor.Initialized])
+
+      uploadTestJar()
+      manager ! JobManagerActor.StartJob("demo", jobName, emptyConfig.withValue(NamedObjectsTestJobConfig.CREATE_DF, ConfigValueFactory.fromAnyRef(false))
```
velvia: Replace the `emptyConfig.withValue(...)` stuff with `ConfigFactory.parseString(....)`, which is much more readable, and you could avoid the long line as well.

job-server/src/spark.jobserver/JobServerNamedObjects.scala
```diff
@@ -96,5 +93,15 @@ class JobServerNamedObjects(system: ActorSystem) extends NamedObjects {
...
 
+/**
+ * companion object that hold reference to cache with named object so that we can reference
+ * named objects across jobs
+ */
+object JobServerNamedObjects {
+  // we must store a reference to each NamedObject even though only its ID is used here
+  // this reference prevents the object from being GCed and cleaned by sparks ContextCleaner
+  // or some other GC for other types of objects
+  val namesToObjects: Cache[NamedObject] = LruCache()
```
velvia: This would work for contexts that are forked, but would not work when `context-per-ivm` is false.

velvia:
@TimMaltGermany if context-per-jvm is false, then putting the cache in the object would cause all named objects to be shared amongst everyone in the JVM, which could include multiple contexts, right?  This is probably not what we want.  I realize this is an edge case, but I think it would be slightly better to leave the LRU cache in the JobServerNamedObjects class, but instantiate the class only once, at the start of the `JobManagerActor`, instead of at the start of every job.  WDYT?

TimMaltGermany:
Thanks for the comments. 
 - simplified the test class  
- created single instance of `JobServerNamedObjects ` instead of the singleton object
I create the instance as an instance val when the `JobManagerActor ` is created, Should this be done inside the `initialize ` method instead?

CBribiescas:
I can confirm its working with RDD persister and this Persister as well
```

package spark.jobserver

import org.apache.spark.broadcast.Broadcast

/**
 * wrapper for named objects of type Broadcast
 */
case class NamedBroadcast[T](broadcast: Broadcast[T]) extends NamedObject

/**
 * implementation of a NamedObjectPersister for Broadcast objects
 *
 */
class BroadcastPersister[T] extends NamedObjectPersister[NamedBroadcast[T]] {
  override def persist(namedObj: NamedBroadcast[T], name: String) {
    namedObj match {
      case NamedBroadcast(broadcast) =>{}
    }
  }

  override def unpersist(namedObj: NamedBroadcast[T]) {
    namedObj match {
      case NamedBroadcast(broadcast) =>
        broadcast.unpersist(blocking = false)
    }
  }

  /**
    *
    *
    * @param namedBroadcast the NamedBroadcast to refresh
    */
  override def refresh(namedBroadcast: NamedBroadcast[T]): NamedBroadcast[T] = namedBroadcast match {
    case NamedBroadcast(broadcast) =>
      namedBroadcast
  }

}
```

velvia:
@TimMaltGermany hey thanks!  And all the tests are passing!  Going to merge.

@CBribiescas just two comments:

- `override def persist` : this seems to be a NOP, so you might as well just make the body empty, e.g.
    `override def persist(namedObj: NamedBroadcast[T], name: String) {}`
- for refresh.  I don't think anything has to be done either....

Would you be interested in submitting a PR for NamedBroadcast?

CBribiescas:
Yeah, I'll submit a PR today for That.

PR #398, opened 2016-03-11
add in an option to debug

addisonj:
This allows for `DEBUG_PORT` to be set as env var, when set, this turns on remote debugging for SJS


velvia:
@addisonj thanks

PR #404, opened 2016-03-16
Adding NamedBroadcasts with persister

CBribiescas:
I also tested with another set of classes from our repo.  I had a generic test which I wanted to copy/paste in here however I can see that its out of date.  The real test I did does work however it contains company IP and I can't post.

### DIFF
--- a/doc/yarn.md
+++ b/doc/yarn.md
@@ -4,6 +4,8 @@
 
 (I would like to thank Jon Buffington for sharing the config tips below.... @velvia)
 
+Note:  This is for yarn with docker.  If you are looking to deploy on a yarn cluster via EMR, then this link would be more useful [EMR](https://github.com/spark-jobserver/spark-jobserver/blob/master/doc/EMR.md)
+
 ### Configuring the Spark-Jobserver Docker package to run in Yarn-Client Mode
 
 To run the Spark-Jobserver in yarn-client mode you have to do a little bit extra of configuration.
--- a/job-server-extras/src/spark.jobserver/NamedBroadcast.scala
+++ b/job-server-extras/src/spark.jobserver/NamedBroadcast.scala
@@ -12,9 +12,6 @@ case class NamedBroadcast[T](broadcast: Broadcast[T]) extends NamedObject
   */
 class BroadcastPersister[T] extends NamedObjectPersister[NamedBroadcast[T]] {
   override def persist(namedObj: NamedBroadcast[T], name: String) {
-    namedObj match {
-      case NamedBroadcast(broadcast) =>{}
-    }
   }
   override def unpersist(namedObj: NamedBroadcast[T]) {
     namedObj match {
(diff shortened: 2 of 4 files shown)


job-server-extras/src/spark.jobserver/NamedBroadcast.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+case class NamedBroadcast[T](broadcast: Broadcast[T]) extends NamedObject
+
+/**
+  * implementation of a NamedObjectPersister for Broadcast objects
+  */
+class BroadcastPersister[T] extends NamedObjectPersister[NamedBroadcast[T]] {
+  override def persist(namedObj: NamedBroadcast[T], name: String) {
+    namedObj match {
+      case NamedBroadcast(broadcast) =>{}
```
velvia: You could leave out the `{}`.  Actually, you could just make the entire body of this method empty, the match and case isn't doing anything.

job-server-extras/test/spark.jobserver/NamedObjectsJobSpec.scala
```diff
@@ -28,10 +28,11 @@ class NamedObjectsJobSpec extends JobSpecBase(JobManagerSpec.getNewSystem) {
 
   val jobName = "spark.jobserver.NamedObjectsTestJob"
 
-  private def getCreateConfig(createDF: Boolean, createRDD: Boolean) : Config = {
+  private def getCreateConfig(createDF: Boolean, createRDD: Boolean, createBroadcast: Boolean) : Config = {
```
velvia: You could avoid the need to change most of the test by adding a default parameter, i.e. `createBroadcast: Boolean = false`

velvia:
@CBribiescas thanks for the PR!   Just some minor comments, otherwise LGTM.
Were you going to add to the docs btw for EMR?  I guess that could be a separate PR.

CBribiescas:
@velvia I've made the changes you suggested.  For the EMR docs I was mentioning on Gitter, I was only saying that the page named "Yarn tips" was a bit confusing because I was on Yarn via EMR so I went there when really I needed to go to the EMR page.  I've added one line in hopes to prevent someone else from running into that issue in the future.

That being said, I'm happy to contribute to the EMR page after my proof of concept with anything that may help.

-Carlos

doc/yarn.md
```diff
@@ -4,6 +4,8 @@
 
 (I would like to thank Jon Buffington for sharing the config tips below.... @velvia)
 
+Note:  This is for yarn with docker.  If you are looking to deploy on a yarn cluster via EMR, then this link would be more useful [EMR](https://github.com/spark-jobserver/spark-jobserver/blob/master/doc/EMR.md)
+
```
velvia: Thanks!

velvia:
@CBribiescas thanks! Going to merge.

PR #408, opened 2016-03-19
Wraps exceptions into RuntimeException objects as a workaround for the case of custom exceptions.

ishassan:


velvia:
@ishassan thanks!  Would you mind fixing the failing test in JobManagerSpec?  Changing the return type from Illegal*Exception to RuntimeException should do the trick.

ishassan:
Pushed the fix. It now passes the tests for Scala 2.11.6, but not for Scala 2.10.4, producing this error message
```
[info] - jobs should not allow user with valid authorization when timeout *** FAILED ***
[info]   200 OK was not equal to 500 Internal Server Error (WebApiWithAuthenticationSpec.scala:164)
```
I'm not really sure how that is related to this change!

velvia:
Islam, thanks.   Would you mind marking that test with ignore?    It’s been failing intermittently for a while and needs to be investigated separately. 

thanks!

ishassan:
Done. Hopefully everything is ok now. Thanks a lot Evan for your guidance and for giving me the chance to play around and try all these things :).

velvia:
@ishassan thanks so much.  Merging, welcome to the club of contributors!

v-gerasimov:
@ishassan Could you explain what exactly is the problem that you are trying to solve by this PR? I catch some exceptions and I am wrapping it in another exception, so I want that spark-jobserver returns it to me. For example, I catch every sql parsing problem (it's different types of exception) and wrapping it into my SqlParsingException. But with your PR I can't do it, because it always returns root exception.

PR #410, opened 2016-03-25
expose MaxDirectMemorySize via env var

addisonj:
With newer versions of Netty, it appears `MaxDirectMemorySize` needs to be
increased for certain uses cases

This change exposes an env var, `MAX_DIRECT_MEMORY`, which allows for setting
this value


velvia:
@addisonj thanks!

PR #418, opened 2016-04-07
Added support for chunked encoding when job result is a Stream[Byte]

leone17:
If a job result is scala Stream[Byte] it is returned in body using chunk encoding

### DIFF
--- a/job-server/src/spark.jobserver/WebApi.scala
+++ b/job-server/src/spark.jobserver/WebApi.scala
@@ -1,38 +1,41 @@
 package spark.jobserver
 
+import java.util.NoSuchElementException
 import java.util.concurrent.TimeUnit
+import javax.net.ssl.SSLContext
 
-import akka.actor.{ ActorSystem, ActorRef }
+import akka.actor.{Actor, ActorLogging, ActorRef, ActorSystem, Props}
 import akka.pattern.ask
 import akka.util.Timeout
-import com.typesafe.config.{ Config, ConfigFactory, ConfigException, ConfigRenderOptions }
-import java.util.NoSuchElementException
-import javax.net.ssl.SSLContext
-import ooyala.common.akka.web.{ WebService, CommonRoutes }
+import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigRenderOptions}
+import ooyala.common.akka.web.JsonUtils.AnyJsonFormat
+import ooyala.common.akka.web.{CommonRoutes, WebService}
+import org.apache.shiro.SecurityUtils
+import org.apache.shiro.config.IniSecurityManagerFactory
 import org.joda.time.DateTime
 import org.slf4j.LoggerFactory
-import spark.jobserver.util.SparkJobUtils
-import spark.jobserver.util.SSLContextFactory
+import spark.jobserver.auth._
+import spark.jobserver.io.JobInfo
 import spark.jobserver.routes.DataRoutes
+import spark.jobserver.util.{SSLContextFactory, SparkJobUtils}
+import spray.can.Http
+import spray.http._
+import spray.httpx.SprayJsonSupport.sprayJsonMarshaller
+import spray.io.ServerSSLEngineProvider
+import spray.json.DefaultJsonProtocol._
+import spray.routing.directives.AuthMagnet
+import spray.routing.{HttpService, RequestContext, Route}
+
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.util.Try
-import spark.jobserver.io.JobInfo
-import spark.jobserver.auth._
-import spray.http.HttpResponse
-import spray.http.MediaTypes
-import spray.http.StatusCodes
-import spray.httpx.SprayJsonSupport.sprayJsonMarshaller
-import spray.json.DefaultJsonProtocol._
-import spray.routing.{ HttpService, Route, RequestContext }
-import spray.routing.directives.AuthMagnet
-import spray.io.ServerSSLEngineProvider
-import org.apache.shiro.config.IniSecurityManagerFactory
-import org.apache.shiro.mgt.SecurityManager
-import org.apache.shiro.SecurityUtils
+
 
 object WebApi {
   val StatusKey = "status"
   val ResultKey = "result"
+  val ResultKeyStartBytes = "{\n".getBytes
+  val ResultKeyEndBytes = "}".getBytes
+  val ResultKeyBytes = ("\"" + ResultKey + "\":").getBytes
 
   def badRequest(ctx: RequestContext, msg: String) {
     ctx.complete(StatusCodes.BadRequest, errMap(msg))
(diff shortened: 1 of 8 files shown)


job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -61,6 +64,14 @@ object WebApi {
     Map(ResultKey -> result)
   }
 
+  def resultToByteIterator(jobReport: Map[String, Any], result: Iterator[_]): Iterator[_] = {
+    ResultKeyStartBytes.toIterator ++
+      (jobReport.map(t => Seq(AnyJsonFormat.write(t._1).toString(),
```
velvia: Hmmm... why not just format the jobReport directly into a string, instead of doing it by hand?  WebApi elsewhere translates maps to strings. `AnyJsonFormat.write(jobReport)` should do it I think.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -112,6 +124,12 @@ class WebApi(system: ActorSystem,
   val DefaultJobLimit = 50
   val StatusKey = "status"
   val ResultKey = "result"
+  val ResultChunkSize = if (config.hasPath("spark.jobserver.result-chunk-size")) {
```
velvia: `Try(config.getBytes("spark.jobserver.result-chunk-size").toInt).getOrElse(100 * 1024)`

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -211,7 +229,8 @@ class WebApi(system: ActorSystem,
    *    DELETE /data/<filename>       - deletes given file, no-op if file does not exist
    *    POST /data/<filename-prefix>  - upload a new data file, using the given prefix,
    *                                      a time stamp is appended to ensure uniqueness
-   * @author TimMaltGermany
+    *
```
velvia: Minor, but why indent part of the comment?

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -416,7 +441,8 @@ class WebApi(system: ActorSystem,
          * [
          *   {jobId: "word-count-2013-04-22", status: "RUNNING"}
          * ]
-         * @optional @param limit Int - optional limit to number of jobs to display, defaults to 50
+          *
+          * @optional @param limit Int - optional limit to number of jobs to display, defaults to 50
```
velvia: Again, why reformat just part of a comment?

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -466,7 +492,13 @@ class WebApi(system: ActorSystem,
                       JobManagerActor.StartJob(appName, classPath, jobConfig, events))(timeout)
                     respondWithMediaType(MediaTypes.`application/json`) { ctx =>
                       future.map {
-                        case JobResult(_, res)       => ctx.complete(resultToTable(res))
+                        case JobResult(_, res) => {
```
velvia: No need for braces around the case code block.  This style is not used elsewhere in the code.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -384,7 +404,12 @@ class WebApi(system: ActorSystem,
                 val resultFuture = jobInfo ? GetJobResult(jobId)
                 resultFuture.map {
                   case JobResult(_, result) =>
-                    ctx.complete(jobReport ++ resultToTable(result))
+                    result match {
+                      case s: Stream[_] =>
```
velvia: Have you tested this?    GET /jobs/<jobId> could be used to retrieve a streams multiple times, want to make sure this works.

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -504,6 +536,43 @@ class WebApi(system: ActorSystem,
     }
   }
 
+  private def sendStreamingResponse(ctx: RequestContext,
+                                    chunkSize: Int,
+                                    byteIterator: Iterator[_]): Unit = {
+    // simple case class whose instances we use as send confirmation message for streaming chunks
+    case class Ok(remaining: Iterator[_])
```
velvia: I think this entire Actor definition should be moved out to another file/Class by itself, instead of being defined inline.   Would be more readable.  Perhaps sendStreamingResponse should be moved out to a trait as well.

velvia:
@leone17 overall, a very welcome and awesome addition.  Hopefully helpful comments.

One thought/doubt - have you tested this with context-per-jvm?   Most contexts deployed from now on will be forked in a separate process.   Furthermore down the line we might try to persist job results.  Both of these things point to the importance of serializing the results.  Would a Stream() be fully serializable, and what are the issues if any?

job-server/src/spark.jobserver/WebApi.scala
leone17: Calling AnyJsonFormat.write(jobReport) will turn it into JsObject, whereas I want to have the whole things wrapped into a JsObject including chunk encoded result, but do it in a lazy evaluated way, which is why I'm using iterators here.

job-server/src/spark.jobserver/WebApi.scala
leone17: I wouldn't want to use exceptions for this, besides Try would swallow any exception and return a default including exception I don't want to mask.
How about
`Option("spark.jobserver.result-chunk-size").filter(config.hasPath).fold(100 * 1024)(config.getBytes(_).toInt)
`

job-server/src/spark.jobserver/WebApi.scala
leone17: Yep, I'll fix this.

job-server/src/spark.jobserver/WebApi.scala
leone17: Will fix this

job-server/src/spark.jobserver/WebApi.scala
leone17: ok, will fix this up

job-server/src/spark.jobserver/WebApi.scala
leone17: Streams are memorized, so this can be called multiple times. It wouldn't be desirable to keep a large Stream in memory, though. Once persistence of results is implemented it would be better to stream it out of storage again.

job-server/src/spark.jobserver/WebApi.scala
leone17: Ok, I'll refactor and update the PR.

leone17:
@velvia thanks for the comments.
Regarding your context-per-jvm question, in scala 2.11 Streams are now serializable (it may be necessary to call .force on the Stream instance), but it wouldn't be desirable to serialize the whole stream as a blob between processes, that would defeat the purpose. In that scenario it would be better to enhance this mechanism to chunk between spray and JobManager processes and directly chunk stream the response back. Once job result persistence is implemented it may be simpler to persist the Stream returned from runJob() and then in the other process read it and stream it out of InputStream.
Either way this PR does not deal with either of those scenarios. It can be revisited and enhanced later.
Thoughts?

job-server/src/spark.jobserver/WebApi.scala
velvia: Lazy is good, but in this case, your code is simply manually creating the json map bytes, then turning those into iterator anyways.... what I'm saying is that doing `AnyJsonFormat.write(jobReport)`, converting to string/bytes, then `toIterator`, is no less lazy than what you currently have (since the main lazy part is the stream output).

velvia:
@leone17 I'm OK with merging this, would you mind adding a note in the README that this streams support will not work for when context-per-jvm=true (my understanding of your earlier comments, right?), along with a TODO somewhere.  We can address the context-per-ivm changes later.

Thanks!

leone17:
@velvia Thanks, updated README and added a comment

job-server/src/spark.jobserver/WebApi.scala
leone17: But I'm trying to avoid converting the lot including result into JsObject tokens, since the whole thing including result in the middle of jobReport JsObject needs to be chunked in the response. Unless I'm missing something I can't see how I can do it directly with `AnyJsonFormat.write(jobReport)` without performing conversion of the whole thing in strict way. I agree the things looks a bit ugly, but unless I modify AnyJsonFormat in some way, I'm not sure how to do it as it stands.

velvia:
@leone17 many thanks!

PR #422, opened 2016-04-12
Add support of java8

stannie42:


velvia:
This looks good, many thanks!

PR #425, opened 2016-04-12
Upgraded to sbt 0.13.11, upgraded dependencies to latest minor verson…

hntd187:
I fixed tests to pass for new versions, the only things that were wrong was some lists came out of order. But since I don't believe this alters the overall intent of the job server this is probably fine, but if you want i can investigate this some more and change the test back.

I also fixed some warnings along the way, but I have no loyalty or opinion of them so from a style perspective if you want we can back any of them out.

### DIFF
--- a/job-server/test/spark.jobserver/DataManagerActorSpec.scala
+++ b/job-server/test/spark.jobserver/DataManagerActorSpec.scala
@@ -15,12 +15,12 @@ class DataManagerActorSpec extends TestKit(DataManagerActorSpec.system) with Imp
     with FunSpecLike with Matchers with BeforeAndAfter with BeforeAndAfterAll {
 
   import com.typesafe.config._
-  import CommonMessages.NoSuchJobId
   import DataManagerActor._
 
   private val bytes = Array[Byte](0, 1, 2)
   private val tmpDir = Files.createTempDirectory("ut")
-  private val config = ConfigFactory.empty().withValue("spark.jobserver.datadao.rootdir", ConfigValueFactory.fromAnyRef(tmpDir.toString))
+  private val config = ConfigFactory.empty().withValue("spark.jobserver.datadao.rootdir",
+    ConfigValueFactory.fromAnyRef(tmpDir.toString))
 
   override def afterAll() {
     dao.shutdown()
--- a/job-server/test/spark.jobserver/JobSpecBase.scala
+++ b/job-server/test/spark.jobserver/JobSpecBase.scala
@@ -17,10 +17,12 @@ trait JobSpecConfig {
   import collection.JavaConverters._
 
   val JobResultCacheSize = Integer.valueOf(30)
-  val NumCpuCores = Integer.valueOf(Runtime.getRuntime.availableProcessors())  // number of cores to allocate. Required.
-  val MemoryPerNode = "512m"  // Executor memory per node, -Xmx style eg 512m, 1G, etc.
+  // number of cores to allocate. Required.
+  val NumCpuCores = Integer.valueOf(Runtime.getRuntime.availableProcessors())
+  // Executor memory per node, -Xmx style eg 512m, 1G, etc.
+  val MemoryPerNode = "512m"
   val MaxJobsPerContext = Integer.valueOf(2)
-  def contextFactory = classOf[DefaultSparkContextFactory].getName
+  def contextFactory: String = classOf[DefaultSparkContextFactory].getName
   lazy val config = {
     val ConfigMap = Map(
       "spark.jobserver.job-result-cache-size" -> JobResultCacheSize,
@@ -29,7 +31,7 @@ trait JobSpecConfig {
       "spark.jobserver.max-jobs-per-context" -> MaxJobsPerContext,
       "spark.jobserver.named-object-creation-timeout" -> "60 s",
       "akka.log-dead-letters" -> Integer.valueOf(0),
-      "spark.master" -> "local[4]",
+      "spark.master" -> "local[*]",
       "context-factory" -> contextFactory,
       "spark.context-settings.test" -> ""
     )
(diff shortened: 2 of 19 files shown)


hntd187:
@velvia hmmm, not sure what is going on with this travis build, any insight?

job-server/test/spark.jobserver/DataManagerActorSpec.scala
```diff
@@ -15,12 +15,12 @@ class DataManagerActorSpec extends TestKit(DataManagerActorSpec.system) with Imp
     with FunSpecLike with Matchers with BeforeAndAfter with BeforeAndAfterAll {
 
   import com.typesafe.config._
-  import CommonMessages.NoSuchJobId
   import DataManagerActor._
 
   private val bytes = Array[Byte](0, 1, 2)
   private val tmpDir = Files.createTempDirectory("ut")
-  private val config = ConfigFactory.empty().withValue("spark.jobserver.datadao.rootdir", ConfigValueFactory.fromAnyRef(tmpDir.toString))
+  private val config = ConfigFactory.empty().withValue("spark.jobserver.datadao.rootdir",
```
velvia: Really moniro, but you can also do `ConfigFactory.fromMap(Map("spark.jobserver.datadao.rootdir", tmpDir.toString).asJava)`  ... though you need `collection.JavaConverters._`

job-server/test/spark.jobserver/JobSpecBase.scala
```diff
@@ -29,7 +31,7 @@ trait JobSpecConfig {
       "spark.jobserver.max-jobs-per-context" -> MaxJobsPerContext,
       "spark.jobserver.named-object-creation-timeout" -> "60 s",
       "akka.log-dead-letters" -> Integer.valueOf(0),
-      "spark.master" -> "local[4]",
+      "spark.master" -> "local[*]",
```
velvia: Does this mean the # of cores you have?

job-server/test/spark.jobserver/JobSpecBase.scala
```diff
@@ -44,14 +46,14 @@ trait JobSpecConfig {
...
-      "streaming.stopSparkContext" -> true
+      "streaming.batch_interval" -> Integer.valueOf(40),
+      "streaming.stopGracefully" -> Boolean.box(false),
+      "streaming.stopSparkContext" -> Boolean.box(true)
     )
     ConfigFactory.parseMap(ConfigMap.asJava).withFallback(ConfigFactory.defaultOverrides())
   }
 
-  def getNewSystem = ActorSystem("test", config)
+  def getNewSystem: ActorSystem = ActorSystem("test", config)
```
velvia: This is pretty obvious right?  Not biggie either way

job-server/test/spark.jobserver/io/JobSqlDAOSpec.scala
```diff
@@ -228,36 +228,36 @@ class JobSqlDAOSpec extends TestJarFinder with FunSpecLike with Matchers with Be
...
       // Cannot compare JobInfos directly if error is a Some(Throwable) because
       // Throwable uses referential equality
       dao.saveJobInfo(jobInfoNoEndSomeErr)
       val jobs2 = dao.getJobInfos(2)
       jobs2.size should equal (2)
-      jobs2.head.endTime should equal (None)
-      jobs2.head.error.isDefined should equal (true)
-      intercept[Throwable] { jobs2.head.error.map(throw _) }
-      jobs2.head.error.get.getMessage should equal (throwable.getMessage)
+      jobs2.last.endTime should equal (None)
```
velvia: Hmmm.... it seems the output of this DAO has gotten flipped....  not sure how that happened

velvia:
Looks mostly fine, I gave lots of small minor comments.  As for the test failures I'm not sure, that error message about major.minor is like a JDK error.   There is another PR that I may merge that may fix this, by upgrading to JDK8.  So hold on for a sec.

velvia:
Some other possibilities:
The travis.yml file might need to be changed for SBT 0.13.11?
Also did any of the dep upgrades also upgrade type safe config to a version that doesn't support Java 7 anymore?  If so that's potentially problematic.... though I personally can't check because I'm on JDK8 locally.

job-server/test/spark.jobserver/JobSpecBase.scala
hntd187: Yes * = number of available cores, so if you have say an i7, it's equivalent to local[8]

job-server/test/spark.jobserver/io/JobSqlDAOSpec.scala
hntd187: I put some effort into trying to figure it out and for the life of me it's a mystery. I ran a bunch of jobs through the server to see if it mattered and everything was fine, but I'm speaking purely empirically. I can't think of a time when job order like this would matter, but I tried my best to verify such was the case.

hntd187:
You know I think that's it. Typesafe's config 1.3.0 is for Java8, 1.2.1 is for Java6+, so probably just an amateur mistake on my part. I'm running JDK8 locally, so probably why I didn't catch it.

hntd187:
Okay, so I went and installed and ran it on the JDK7, which yea, fails. It's because in the typesafe config 1.3.0 they use java.time.Duration, which obviously is a java8 feature, where we still use Joda, so that's why it fails. Okay, lemme update this to be correct based on that.

velvia:
Yeah or just force Config to be 1.2.0

-Evan

job-server/test/spark.jobserver/DataManagerActorSpec.scala
hntd187: This was because it violated the style guide for line length.

velvia:
@hntd187 thanks, I think this LGTM, but can you merge from master and update so stuff can be merged?  Coolio.

hntd187:
@velvia Evan, I rebased off it after some frustration, hopefully it still works, I think my rebase was a bit screwed up so hopefully nothing bad happened

velvia:
@hntd187 many thanks!

PR #430, opened 2016-04-15
add new feathure for web ui 

gaoshui87:
the web ui show more message for the job message,
and i am finish patch when the job fail.the fail count number is wrong


bin/server_start.sh
```diff
@@ -28,7 +28,7 @@ GC_OPTS="-XX:+UseConcMarkSweepGC
 # To truly enable JMX in AWS and other containerized environments, also need to set
 # -Djava.rmi.server.hostname equal to the hostname in that environment.  This is specific
 # depending on AWS vs GCE etc.
-JAVA_OPTS="-XX:MaxDirectMemorySize=$MAX_DIRECT_MEMORY \
+JAVA_OPTS="-XX:MaxDirectMemorySize=512M \
```
velvia: Hmm.. this change seems to take out a PR that we merged recently.... we want this $MAX_DIRECT_MEMORY there so users can configure it, can we revert this change? Thanks.

velvia:
@gaoshui87 Thanks very much for the PR.  If I was to summarize your change, it is like this:
- Add a script to kill the forked processes via the PID
- Track the PIDs of forked processes in AkkaClusterSupervisor
- Some new messages to the actors
- Some other random changes, config changes etc

I think the first three are valuable, but we just need to clean this up.  The last one, we want to be careful not to revert existing changes.

velvia:
Also, let's tackle this change after we release 0.6.2.  Thanks!

gaoshui87:
@velvia  yor are right
1,first i kill the processes 
2,add a messge to get the jobs ,how many number is runnig
3,it is rich the web ui to see more message

velvia:
@gaoshui87 do you mind to close this PR and submit a new one based on latest master?  There are huge number of changes between this PR branch and master, so we can't merge it.  Thanks.

hntd187:
@gaoshui87 Also could you not include your target directory in the commit :-)

hntd187:
@gaoshui87 it's been about 20 days or so, could you please rebase from master and resubmit?

PR #435, opened 2016-04-20
Some Build refactoring and updating...

hntd187:
Okay this seems huge, but it's not.

1. Added a shell script to test both JDKs
2. Did all the build plugins to latest versions and corrected problems
3. Refactored versions and exclusion rules into their own files so they can be included and managed easier than the dependency files.
4. Fixed some warnings
5. Because a scalariform style default was deprecated the plugin chose to reformat literally the entire code base, so that's why we have this. I can honestly understand not wanting this at all, so I can back it, just lemme know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,5 @@ env:
   global:
     _JAVA_OPTIONS="-Xmx1500m -XX:MaxPermSize=512m -Dakka.test.timefactor=3"
 scala:
-   - 2.10.4
-   - 2.11.6
+   - 2.10.6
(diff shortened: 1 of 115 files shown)


velvia:
Hmm… most of these commits are from master.  I think it must have to do with how you are merging in changes, that somehow they appear again….

hntd187:
Oh god damn it, you're right. Okay lemme fix this, sorry.

hntd187:
Okay, I have no idea what is happening here. Let me sort this out and I'll resubmit this.

PR #433, opened 2016-04-19
Upgrade sbt and sbt plugins: sbt-revolver, sbt-assembly, and sbt-docker

jaceklaskowski:
The title says it all. Please review @velvia and merge at your convenience. Thanks.

PR #438, opened 2016-04-21
Plugins and Build Refactoring...

hntd187:
Okay, this should be just normal commits and not 900 extra commits I somehow put in there. 

This request looks huge, but it's really not...

1. Added a shell script to test both JDKs
2. Did all the build plugins to latest versions and corrected problems
3. Refactored versions and exclusion rules into their own files so they can be included and managed easier than the dependency files.
4. Fixed some warnings
5. Because a scalariform style default was deprecated the plugin chose to reformat literally the entire code base, so that's why we have this. I can honestly understand not wanting this at all, so I can back it, just lemme know, but it is doing what the plugin was told to do, so it's not unexpected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ Spark Job Server is now included in Datastax Enterprise 4.8!
 | 0.5.2       | 1.3.1         |
 | 0.6.0       | 1.4.1         |
 | 0.6.1       | 1.5.2         |
-| master      | 1.6.0         |
+| 0.6.2       | 1.6.1         |
+| master      | 1.6.1         |
 
 For release notes, look in the `notes/` directory.  They should also be up on [notes.implicit.ly](http://notes.implicit.ly/search/spark-jobserver).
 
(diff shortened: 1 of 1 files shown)


hntd187:
@velvia I'm pretty sure the one test flaked out, can you do a retest?

velvia:
@hntd187 thanks, flaky test restarted.   Wonder if Travis has some option to auto retest failed ones.

hntd187:
This also encompasses the changes in #433

akka-app/src/ooyala.common.akka/web/CommonRoutes.scala
```diff
@@ -81,33 +86,37 @@ object MetricsSerializer {
...
   }
 
   private def meterToMap(m: Metered) =
-    Map("units" -> m.rateUnit.toString.toLowerCase,
-        "count" -> m.count,
-        "mean"  -> m.meanRate,
-        "m1"    -> m.oneMinuteRate,
-        "m5"    -> m.fiveMinuteRate,
-        "m15"   -> m.fifteenMinuteRate)
+    Map(
```
velvia: This is a case where the formatting doesn't look as good as before

velvia:
@hntd187 sorry bud, but this is pretty hard to review due to the sheer reformatting ... would prefer to back that out so we can look at the other changes.

velvia:
@hntd187 I'd like to release without this PR....

hntd187:
Yea, don't let this hold the release back. I'll roll back the formatting changes.

velvia:
I guess this is not needed due to the other PRs?

hntd187:
I rolled back the formatting changes, this is still needed as it upgraded the final out of date plugins. @velvia

PR #440, opened 2016-04-23
Upgrade Mesos to 0.28.1

jaceklaskowski:


velvia:
@jaceklaskowski thanks, don't see why not  :)

PR #441, opened 2016-04-25
Use passed `spark.driver.memory` when creating a context

buryat:
I'll change the code so it could be auto-mergeable.

Not sure if we need to pass other driver properties


velvia:
@buryat thanks.  Do you want to merge with master latest?

buryat:
@velvia Merged with the latest master.

job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -205,10 +205,15 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
           return
       }
 
+    var driverMemory = "0"
+    if (contextConfig.hasPath("spark.driver") && contextConfig.hasPath("spark.driver.memory")) {
```
noorul: is it not enough to check contextConfig.hasPath("spark.driver.memory") ?

hntd187: Yes, this is redundant should only be spark.driver.memory

buryat: Not enough, the `driver` section might not exist completely.
Similar check you have for `spark.proxy` in here https://github.com/spark-jobserver/spark-jobserver/pull/441/files#diff-8a327aff8a0090a45c45ca1544981684R216

velvia: I think you can eliminate the check entirely by doing something like

    Try(config.getString("spark.driver.memory")).getOrElse(....)

or use match etc.   Lately I've been using another library called Ficus which can return Options.  Either way is much DRYer than having to check beforehand.

hntd187: Ficus is fantastic btw, takes out a lot of boilerplate code for checking the existence of keys first and retrieving them when you don't know their types ahead of time.

velvia:
@buryat sorry do you mind rebasing from latest master?  There are now conflicts.

buryat:
Unfortunately, it's a wrong approach, I'm closing the PR.

PR #442, opened 2016-04-27
New SparkJob API

velvia:
Since the beginning of Spark Job Server, the SparkJob API has pretty much remained unchanged.  Over time, however, the API has shown its age.  Some of its issues include:
- Conflation of the job input with the context configuration
- Need to mix in NamedRDDs / NamedObjects
- Unavailability of things like job ID and other environmental things
- Possible need to parse Config twice: once in Validate method, and once in runJob

This PR represents a new SparkJob API which is designed to solve all of the above issues.  
Existing jobs can still be compiled, but will not be binary compatible with this new SJS.
The new API appears under a new namespace, spark.jobserver.api.SparkJob

In addition, SparkContextFactory will be extended to support loading of non-api.SparkJobBase jars as well as different types of apps such as Python.  Basically the app loading logic will be delegated to each Factory, with the requirement that it returns something called a JobContainer, which ultimately returns a SparkJobBase.  For example, a Java-specific (or Python) API could be supported so long as it is wrapped in a SparkJobBase.

Still TODO: Java support is currently broken.


reactormonk:
I'd go with `Validated` instead of `Or` - seems better from a semantic side imo. http://eed3si9n.com/herding-cats/Validated.html

On the Java side - just add a version with lots of `asInstanceOf`, `Object` and add an "abandon all types ye who enter here" shield.

hntd187:
Is there any easy way to paramaterize the runJob return type so we can maintain some type safety surround it? Should try and give it a concrete type instead of Any.

velvia:
Did you have a look at the details in the change?  there is now a JobOutput type so you can have the safety !

-Evan

hntd187:
@velvia Oh! Not sure how I missed that, but awesome!

hntd187:
@velvia Looks great Evan, also I was wondering if in this we could consider some refactoring that would make the server-extras dependency not rely on the entire jobserver and thus the api becomes dependency-less. One thing that I noticed was the server extras pulls in the entire jobserver, which I think unless you had some specific reason to do that should probably be refactored to not do that. I already worked out what would be mostly necessary to free that dependency in this commit 

https://github.com/hntd187/spark-jobserver/commit/7659b4a6b9343070f0b95494ee2766d0d0136049

is there anyway we could cherry pick this or also implement this refactor so api and extras are truly independent of the server dependency itself? I believe I don't actually write any code for this, I just move files around for locality.

velvia:
@hntd187 thanks, actually if you look one of my commits is to separate out the job server extras, so it should be done already  :)   I didn’t realize you had done it already though so feel free to comment on that

hntd187:
@velvia Awesome! Comparing the two commits 
https://github.com/hntd187/spark-jobserver/compare/extras...spark-jobserver:758e5c8e77fef5b47098557c8b040071828df04f?expand=1 it seems like we basically had roughly the same idea, the only differences to me seem to be the new API stuff so that is fantastic this LGTM

README.md
```diff
@@ -258,6 +259,29 @@ manage and re-use contexts.
 - `validate` allows for an initial validation of the context and any provided configuration. If the context and configuration are OK to run the job, returning `spark.jobserver.SparkJobValid` will let the job execute, otherwise returning `spark.jobserver.SparkJobInvalid(reason)` prevents the job from running and provides means to convey the reason of failure. In this case, the call immediately returns an `HTTP/1.1 400 Bad Request` status code.  
 `validate` helps you preventing running jobs that will eventually fail due to missing or wrong configuration and save both time and resources.  
 
+### NEW SparkJob API
+
+Note: As of version 0.7.0, a new SparkJob API that is significantly better than the old SparkJob API will take over.  Existing jobs should continue to compile against the old `spark.jobserver.SparkJob` API, but this will be deprecated in the future.  Note that jobs before 0.7.0 will need to be recompiled, older jobs may not work with the current SJS example.  The new API looks like this:
```
zeitos: wouldn't a change like this deserve a new major version?  1.0?

velvia: Good Q.  Well that’s why it will be 0.7, and not 0.6.3, but maybe you’re right that it’s a big enough change for 1.0.   Well, at the very least, if we extend the SparkJobFactory to support Python jobs, that will definitely be worth a 1.0. 

My only thoughts is that we should make a 1.0 more stable and free of bugs than today, and we need to think about Spark versioning.  Spark 2 will be out soon, and some will clamor for that, but lots of folks (most of us I imagine) will have systems on Spark 1.x for a while in production, and not ready to move up.

job-server-tests/src/main/java/spark.jobserver/JavaHelloWorldJob.java
```diff
@@ -1,12 +1,12 @@
-package spark.jobserver;
+// package spark.jobserver;
```
zeitos: are we removing this whole class?

README.md
zeitos: yeah the "more stable" makes sense... we can always call it 1.0.0-alpha but I guess we can stick to 0.7 since version 0 means that you are in a sense: unstable.

zeitos:
@velvia I like the new API :D looks good, I couldn't try it for real yet but the new validation makes a lot of sense.

job-server-tests/src/main/java/spark.jobserver/JavaHelloWorldJob.java
velvia: It doesn’t compile  :(   I can’t figure out why.     I’m hoping to create a new Java-specific API that doesn’t inherit from SparkJobBase, it will be much easier.

zeitos: I don't think spark.jobserver is a valid folder for java, you need spark/jobserver.

Did you try that?

hntd187: @velvia @zeitos It should be something like this, but the jobserver itself needs to account for the different class. The ```type C``` stuff as well as the ```case object``` do not translate well to java, so they don't compile. But also take note of the path I changed as also Zeitos suggested...

https://github.com/hntd187/spark-jobserver/commit/825d02e7051dd73ed5c2acf77a6836e6efb9950a

velvia: Hmm, I think you may be on the right track.  

@hntd187 would you have time to try pushing your idea of fix to this branch?

hntd187: Yes, but I think it should be a separate pull request since the java api is gonna need some project restructuring since java is picky about what is located where.

hntd187:
@velvia I rebased off master for you to get the latest changed in. Unless someone objects I'll merge this after the tests run.

hntd187:
Okay, passed 2 runs of tests, let's merge. @velvia

velvia:
@hntd187 many thanks.  Looking forward to your Java API stuff, the current java support on master is now broken.

PR #449, opened 2016-04-30
Build refactoring and updating final plugins...

hntd187:
I reset my master so it's the same request as before for #438 

Okay, this should be just normal commits and not 900 extra commits I somehow put in there.

This request looks huge, but it's really not...

Added a shell script to test both JDKs
Did all the build plugins to latest versions and corrected problems
Refactored versions and exclusion rules into their own files so they can be included and managed easier than the dependency files.
Fixed some warnings
Because a scalariform style default was deprecated the plugin chose to reformat literally the entire code base, so that's why we have this. I can honestly understand not wanting this at all, so I can back it, just lemme know, but it is doing what the plugin was told to do, so it's not unexpected.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -1,45 +1,35 @@
 import sbt._
-
+import Versions._
+import ExclusionRules._
 object Dependencies {
-  val excludeCglib = ExclusionRule(organization = "org.sonatype.sisu.inject")
-  val excludeJackson = ExclusionRule(organization = "org.codehaus.jackson")
-  val excludeScalaTest = ExclusionRule(organization = "org.scalatest")
-  val excludeScala= ExclusionRule(organization = "org.scala-lang")
-  val excludeNettyIo = ExclusionRule(organization = "io.netty", artifact= "netty-all")
-  val excludeAsm = ExclusionRule(organization = "asm")
-  val excludeQQ = ExclusionRule(organization = "org.scalamacros")
-  val jdkVersion = scala.util.Properties.isJavaAtLeast("1.8")
 
-  lazy val typeSafeConfigDeps = if(jdkVersion){
-    "com.typesafe" % "config" % "1.3.0"
-  } else {
-    "com.typesafe" % "config" % "1.2.1"
-  }
-  lazy val yammerDeps = "com.yammer.metrics" % "metrics-core" % "2.2.0"
+  lazy val typeSafeConfigDeps = "com.typesafe" % "config" % typeSafeConfig
+
+  lazy val yammerDeps = "com.yammer.metrics" % "metrics-core" % metrics
 
   lazy val jodaDeps = Seq(
-    "org.joda" % "joda-convert" % "1.8.1",
-    "joda-time" % "joda-time" % "2.9.3"
+    "org.joda" % "joda-convert" % jodaConvert,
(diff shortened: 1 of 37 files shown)


hntd187:
Flaky test is flaky, can we retest @velvia ?

noorul:
Restarted

velvia:
Thanks for all your help @noorul

noorul:
Pleasure

noorul:
I see that you are not pinning most of the dependencies. Is that a good idea?

hntd187:
@noorul what do you mean? All the versions are being locked  at specific versions,  some of them are ancient which lead to this upgrade.

noorul:
@hntd187 My bad, I overlooked those variables to which version is assigned.

project/Dependencies.scala
```diff
@@ -1,45 +1,35 @@
...
-    "com.typesafe.akka" %% "akka-slf4j" % "2.3.15" % "provided",
-    "com.typesafe.akka" %% "akka-cluster" % "2.3.15" exclude("com.typesafe.akka", "akka-remote"),
-    "io.spray" %% "spray-json" % "1.3.2",
-    "io.spray" %% "spray-can" % "1.3.3",
-    "io.spray" %% "spray-caching" % "1.3.3",
-    "io.spray" %% "spray-routing" % "1.3.3",
-    "io.spray" %% "spray-client" % "1.3.3",
+    "com.typesafe.akka" %% "akka-slf4j" % akka % "provided",
+    "com.typesafe.akka" %% "akka-cluster" % akka exclude("com.typesafe.akka", "akka-remote"),
+    "io.spray" %% "spray-json" % sprayJson,
```
velvia: Not a biggie, but I wonder if `Version.sprayJson` etc, would be more readable.  Probably not.

scalastyle-config.xml
```diff
@@ -19,7 +19,7 @@
  <check level="error" class="org.scalastyle.scalariform.SpacesBeforePlusChecker" enabled="true"></check>
  <check level="error" class="org.scalastyle.file.FileLineLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxLineLength"><![CDATA[110]]></parameter>
+   <parameter name="maxLineLength"><![CDATA[180]]></parameter>
```
velvia: Okay, I disagree with this one.  I hate hate long runaway lines!   Code is far more readable when lines are shorter.  Though I'm sure you know this and there's some reason why this is being increased.

velvia:
@hntd187 looked at first three commits, minor comments, well ok one about line length I'd like changed.

hntd187:
Well personally, I use some large wide screen monitors so short lines drive me kinda nuts, I read code very left to right and as long as it's not like 250 characters long I prefer a longer line, but that's just me personally, I'm sure there would be a lot of disagreement over that sentiment :-) Either way, I will change it back. @velvia

noorul:
LGTM

hntd187:
@noorul @velvia please take note with this that I disabled scalariform for the build because the new version will reformat all the code, why I am not sure because the settings are the same from version to version, so it seems more on scalariform changing something. I'll submit a pull request with it turned back on so we can see the differences.

noorul:
ack

Issue #456, opened 2016-05-11
Extend NamedObjectSupport in order to work with SparkSqlJob


PowerToThePeople111:
Currently only a SparkJob can have the trait NamedObjectSupport. It would be great if this trait could also be added to an SparkSqlJob.

velvia:
Merged so closing.

PR #457, opened 2016-05-11
self-reference is now of supertype SparkJobBase

PowerToThePeople111:
See #456  
The change was simple. Just exchange for the supertype of SparkJob and SparkSqlJob which is SparkJobBase.

I tested it locally and it seems to work.


velvia:
This LGTM, except for flaky tests, so going to merge.

PR #469, opened 2016-05-22
Per-user authenticated contexts

TimMaltGermany:
Added option to require that proxy users must be identical to the authenticated user.

In order to support secured clusters it is necessary to ensure that the user running a spark context is identical to the authenticated user. We added a boolean configuration option (`shiro.use-as-proxy-user`). When authentication and this flag are activated, then spark.proxy.user is always set to the name of the authenticated user. Users will ever only see and be able to manipulate their own contexts. An internal prefix is used to ensure that no naming conflicts arise when different users use the same name for their respective contexts.

Note that in this case the `spark.proxy.user` parameter in the following call is always ignored!
POST /contexts/my-new-context?spark.proxy.user=<IGNORED>

### DIFF
--- a/job-server/src/spark.jobserver/WebApi.scala
+++ b/job-server/src/spark.jobserver/WebApi.scala
@@ -7,7 +7,7 @@ import javax.net.ssl.SSLContext
 import akka.actor.{ActorRef, ActorSystem}
 import akka.pattern.ask
 import akka.util.Timeout
-import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigRenderOptions}
+import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigValueFactory, ConfigRenderOptions}
 import ooyala.common.akka.web.JsonUtils.AnyJsonFormat
 import ooyala.common.akka.web.{CommonRoutes, WebService}
 import org.apache.shiro.SecurityUtils
@@ -247,7 +247,8 @@ class WebApi(system: ActorSystem,
     authenticate(authenticator) { authInfo =>
       get { ctx =>
         (supervisor ? ListContexts).mapTo[Seq[String]]
-          .map { contexts => ctx.complete(contexts) }
+          .map { contexts =>
+            ctx.complete(removeProxyUserPrefix(authInfo, contexts)) }
       } ~
         post {
           /**
@@ -266,8 +267,9 @@ class WebApi(system: ActorSystem,
               complete(StatusCodes.BadRequest, errMap("context name must start with letters"))
             } else {
               parameterMap { (params) =>
-                val config = ConfigFactory.parseMap(params.asJava).resolve()
-                val future = (supervisor ? AddContext(contextName, config))(contextTimeout.seconds)
+                val (cName, config) = determineProxyUser(ConfigFactory.parseMap(params.asJava).resolve(),
+                                                            authInfo, contextName)
+                val future = (supervisor ? AddContext(cName, config))(contextTimeout.seconds)
                 respondWithMediaType(MediaTypes.`application/json`) { ctx =>
                   future.map {
                     case ContextInitialized   => ctx.complete(StatusCodes.OK)
@@ -284,7 +286,8 @@ class WebApi(system: ActorSystem,
           //  Stop the context with the given name.  Executors will be shut down and all cached RDDs
           //  and currently running jobs will be lost.  Use with care!
           path(Segment) { (contextName) =>
-            val future = supervisor ? StopContext(contextName)
+            val (cName, _) = determineProxyUser(config, authInfo, contextName)
+            val future = supervisor ? StopContext(cName)
             respondWithMediaType(MediaTypes.`text/plain`) { ctx =>
               future.map {
                 case ContextStopped => ctx.complete(StatusCodes.OK)
@@ -530,6 +533,36 @@ class WebApi(system: ActorSystem,
 
   override def timeoutRoute: Route =
     complete(500, errMap("Request timed out. Try using the /jobs/<jobID>, /jobs APIs to get status/results"))
+
+  def determineProxyUser(aConfig: Config,
+                         authInfo: AuthInfo,
+                         contextName: String): (String, Config) = {
+    if (config.getBoolean("shiro.authentication") && config.getBoolean("shiro.use-as-proxy-user")) {
+      //proxy-user-param is ignored and the authenticated user name is used
+      val config = aConfig.withValue(SparkJobUtils.SPARK_PROXY_USER_PARAM,
+        ConfigValueFactory.fromAnyRef(authInfo.toString))
+      (authInfo.toString + "_" + contextName, config)
+    } else {
+      (contextName, aConfig)
+    }
+  }
+
+  def removeProxyUserPrefix(authInfo: AuthInfo, contextNames: Seq[String]): Seq[String] = {
+    if (config.getBoolean("shiro.authentication") && config.getBoolean("shiro.use-as-proxy-user")) {
+      val userNamePrefix = authInfo.toString + "_"
+      contextNames map { cName =>
+        {
+          if (cName.startsWith(userNamePrefix)) {
+            Some(cName.substring(userNamePrefix.length))
+          } else {
+            None
+          }
+        }
+      } filter (_.isDefined) map (_.get)
+    } else {
+      contextNames
+    }
+  }
 
   private def getJobManagerForContext(context: Option[String],
                                       contextConfig: Config,
--- a/job-server/src/spark.jobserver/WebApi.scala
+++ b/job-server/src/spark.jobserver/WebApi.scala
@@ -36,7 +36,7 @@ object WebApi {
   val ResultKeyEndBytes = "}".getBytes
   val ResultKeyBytes = ("\"" + ResultKey + "\":").getBytes
 
-  val NameContextDelimiter = "_"
+  val NameContextDelimiter = "~"
 
   def badRequest(ctx: RequestContext, msg: String) {
     ctx.complete(StatusCodes.BadRequest, errMap(msg))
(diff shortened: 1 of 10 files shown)


velvia:
@TimMaltGermany great change.  I think "authenticated proxy users" is not the best title, right?  Isn't this more like "Per-user authenticated contexts"?  Sorry I can't think of a better name.

koettert:
@velvia You are right. It is more of a dedicated context for a authenticated user so why not call it "Per-user authenticated contexts". Can you please rename the push request. Thanks

velvia:
Done

TimMaltGermany:
@velvia @zeitos Any chance that either one of you reviews this? Or propose someone else?

zeitos:
@TimMaltGermany sure!

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -531,6 +534,36 @@ class WebApi(system: ActorSystem,
...
+      (authInfo.toString + "_" + contextName, config)
+    } else {
+      (contextName, aConfig)
+    }
+  }
+
+  def removeProxyUserPrefix(authInfo: AuthInfo, contextNames: Seq[String]): Seq[String] = {
+    if (config.getBoolean("shiro.authentication") && config.getBoolean("shiro.use-as-proxy-user")) {
+      val userNamePrefix = authInfo.toString + "_"
+      contextNames map { cName =>
```
zeitos: Can you use collect instead of map & filter?

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -531,6 +534,36 @@ class WebApi(system: ActorSystem,
...
+        {
+          if (cName.startsWith(userNamePrefix)) {
```
zeitos: mmmm _ is not a valid character for user names right?... I can't have fernando_otero as my username?

I think this restriction has to be documented or we might need to use a less likely separator than '_'

job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala
```diff
@@ -71,42 +72,62 @@ class WebApiWithAuthenticationSpec extends FunSpec with Matchers with BeforeAndA
...
+          sender ! ContextSupervisor.ContextInitialized
+        }
+      case ContextSupervisor.StopContext(name) =>
+        addedContexts.remove(name)
+        sender ! ContextSupervisor.ContextStopped
+      case ContextSupervisor.AddContextsFromConfig =>
+        sender ! "OK"
+      case m =>
+        //dev support when adding new test cases:
+        println("Unhandled message: " + m)
```
zeitos: can you log error?

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -531,6 +534,36 @@ class WebApi(system: ActorSystem,
...
     complete(500, errMap("Request timed out. Try using the /jobs/<jobID>, /jobs APIs to get status/results"))
 
+  def determineProxyUser(aConfig: Config,
+                         authInfo: AuthInfo,
+                         contextName: String): (String, Config) = {
+    if (config.getBoolean("shiro.authentication") && config.getBoolean("shiro.use-as-proxy-user")) {
+      //proxy-user-param is ignored and the authenticated user name is used
+      val config = aConfig.withValue(SparkJobUtils.SPARK_PROXY_USER_PARAM,
+        ConfigValueFactory.fromAnyRef(authInfo.toString))
+      (authInfo.toString + "_" + contextName, config)
```
zeitos: Consider moving "_" to a constant val with a declarative name since it's an important (NameContextDelimiter maybe?)

zeitos:
@TimMaltGermany great job, I added some comments I feel that using _ might end up with name collisions like since user "tim" when calling removeProxyUserPrefix will get all his contexts + some invalid contexts from "tim_germany"  (all prefixed with germany)

how about changing _ to something like _--_ or any other pattern that is not commonly used in user names?

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: No, there is no such restriction. Check out the test 'it("should allow user name prefixes in contexts")' in WebApiWithAuthenticationSpec 
The point is that the context is always prefixed with *some* user's name. So, '_' should be just fine.

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: sure

job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala
TimMaltGermany: you mean 'sys.error("Unhandled message: " + m)' ? 
This is a test class after all. I could also just remove that case.

job-server/src/spark.jobserver/WebApi.scala
TimMaltGermany: Like this?
`contextNames collect {`
        `case cName if (cName.startsWith(userNamePrefix)) =>    `
`            cName.substring(userNamePrefix.length)    `
`      }`

job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala
zeitos: I meant a logger but it's a test, you are right... let it print...

job-server/src/spark.jobserver/WebApi.scala
zeitos: yes, it should be better in performance and readability.
You could even use the prefix a a regex and do something like
```
  val UserNamePrefix = (authInfo.toString + "_(.*)").r
      contextNames collect {
         case UserNamePrefix(cName) => cName
      }

```

TimMaltGermany:
Hmm, I guess you are right. If we have a user 'tim' and a user 'tim_germany' then 'tim' will indeed see bogus contexts of user 'tim_germany' (bogus since they are prefixed with 'germany_' and will not exist). 
How about this: if we actually have a user name that contains our separator, then we replace that separator in the user name by some other character. So, the prefix for 'tim_germany' will become 'tim-germany'? That should always work, right?

job-server/src/spark.jobserver/WebApi.scala
zeitos: yes, but if _ is part of the user name then you can have collisions
In your tests if you add a context for another user called: presidentskroob_newUser the tests start to fail
```
diff --git a/job-server/test/spark.jobserver/auth/SJSAuthenticatorSpec.scala b/job-server/test/spark.jobserver/auth/SJSAuthenticatorSpec.scala
index 9277348..161cf1e 100644
--- a/job-server/test/spark.jobserver/auth/SJSAuthenticatorSpec.scala
+++ b/job-server/test/spark.jobserver/auth/SJSAuthenticatorSpec.scala
@@ -54,6 +54,7 @@ securityManager.cacheManager = $cacheManager
 root = secret, admin
 guest = guest, guest
 presidentskroob = 12345, president
+presidentskroob_newUser = 12345, president
 darkhelmet = ludicrousspeed, darklord, schwartz
 lonestarr = vespa, goodguy, schwartz
 
diff --git a/job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala b/job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala
index fd80e6f..eae3948 100644
--- a/job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala
+++ b/job-server/test/spark.jobserver/auth/WebApiWithAuthenticationSpec.scala
@@ -76,9 +76,13 @@ class WebApiWithAuthenticationSpec extends FunSpec with Matchers with BeforeAndA
   private val routesWithoutProxyUser = routesWithTimeout(false, "1 s")
 
   private val USER_NAME = "presidentskroob"
+  private val USER_NAME2 = "presidentskroob_newUser"
   // set to some valid user
   private val authorization = new Authorization(new BasicHttpCredentials(USER_NAME, "12345"))
   private val authorizationInvalidPassword = new Authorization(new BasicHttpCredentials(USER_NAME, "xxx"))
+
+  private val authorization2 = new Authorization(new BasicHttpCredentials(USER_NAME2, "12345"))
+  private val authorizationInvalidPassword2 = new Authorization(new BasicHttpCredentials(USER_NAME2, "xxx"))
   private val authorizationUnknownUser = new Authorization(new BasicHttpCredentials("whoami", "xxx"))
   private val dt = DateTime.parse("2013-05-29T00Z")
   private val jobInfo = JobInfo("foo-1", "context", JarInfo("demo", dt), "com.abc.meme", dt, None, None)
@@ -206,6 +210,10 @@ class WebApiWithAuthenticationSpec extends FunSpec with Matchers with BeforeAndA
       //clear
       addedContexts.clear
       //add a context that has the user name as a prefix
+      Post("/contexts/" + USER_NAME2 + "_c").withHeaders(authorization2) ~>
+        sealRoute(testRoutes) ~> check {
+          status should be(OK)
+        }
       Post("/contexts/" + USER_NAME + "_c").withHeaders(authorization) ~>
         sealRoute(testRoutes) ~> check {
           status should be(OK)
```

zeitos:
@TimMaltGermany I'm on gitter if you prefer to talk there.

I see two options:
1- you need to use a reserved token that can't be used in user names :S
2- Ensure that there's no way to have _ in the user name ie:  replacing _ by __ so then tim_germany becomes tim__germay and tim__germany will become tim____germany_c1 and so on.
and there's no username with only one _ in its name, and in that way you have a delimiter...

TimMaltGermany:
Either one is fine with me. Option 1 might be more elegant, but I can only think of a \ that is probably never allowed in user names, but that might cause problems elsewhere. Option 2 is not as pretty, but should always work. 
Option 2 then?

zeitos:
well ... option 2 will work if you check that there's only one _ 
or else tim__germany will try to find tim__germany_* prefixes that included the ones from tim__germany__new  :D (an incorrect _new_context)

So instead of substring you'll need to use a regex like:
"^" + userName + "_([^_].+)"

what do you think?

TimMaltGermany:
I think I have addressed all comments. Many thank @zeitos !

job-server/src/spark.jobserver/WebApi.scala
```diff
@@ -531,6 +536,36 @@ class WebApi(system: ActorSystem,
   override def timeoutRoute: Route =
     complete(500, errMap("Request timed out. Try using the /jobs/<jobID>, /jobs APIs to get status/results"))
 
+  def userNamePrefix(authInfo: AuthInfo) : String = {
```
zeitos: One last comment, can you please document this explaining what we talked in the issue (maybe add a link to the issue)

zeitos:
@TimMaltGermany I added a small comment but I think it will be important to whoever reads that line.

velvia:
Just a random thought.  You could also use a much less likely occurring character - such as the tilde (~), or some unicode character.

velvia:
Thanks @zeitos for the review and @TimMaltGermany looking good

TimMaltGermany:
Thanks, I added some comments and changed the separator to ~.

TimMaltGermany:
@velvia Ready to merge then?

zeitos:
Closing, thanks @TimMaltGermany

PR #476, opened 2016-05-30
replace map with foreach

souo:
Expression with unit return type，replace map with foreach。


velvia:
LGTM, going to merge.

PR #496, opened 2016-06-10
fix for flaky-tests

TimMaltGermany:
@noorul @hntd187 I just set up this branch to address the flacky test.

The latest failure on master is caused by too few job slots: 
java.lang.AssertionError: assertion failed: expected class spark.jobserver.CommonMessages$JobResult, found class spark.jobserver.CommonMessages$NoJobSlotsAvailable

I think that can be easily fixed. 

The other issues reported by @hntd187 are:
The tests in spark.jobserver.auth.WebApiWithAuthenticationSpec fail sometimes for no obvious reason. Given how they are written it seems really strange they fail, but one is ignored and the other 2 are not. Could we make a decision on whether or not these tests should remain or set to ignore for the time being? They seem to cover a rather edge case situation.

### DIFF
--- a/job-server-extras/test/spark.jobserver/HiveJobSpec.scala
+++ b/job-server-extras/test/spark.jobserver/HiveJobSpec.scala
@@ -1,5 +1,6 @@
 package spark.jobserver
 
+import scala.util.{ Try, Success, Failure }
 import com.typesafe.config.ConfigFactory
 import org.apache.spark.sql.Row
 import org.apache.spark.{SparkContext, SparkConf}
@@ -9,8 +10,15 @@ import spark.jobserver.context.{HiveContextLike, HiveContextFactory}
 import spark.jobserver.io.{JobDAO, JobDAOActor}
 
 class TestHiveContextFactory extends HiveContextFactory {
-  override protected def contextFactory(conf: SparkConf): C =
-    new TestHiveContext(new SparkContext(conf)) with HiveContextLike
+  override protected def contextFactory(conf: SparkConf): C = {
+    val sc = new SparkContext(conf)
+    Try(new TestHiveContext(sc) with HiveContextLike) match {
+      case Success(hc) => hc
+      case Failure(e) =>
+        sc.stop
+        throw e
+    }
+  }
 }
 
 object HiveJobSpec extends JobSpecConfig {
--- a/README.md
+++ b/README.md
@@ -498,6 +498,8 @@ database created with necessary rights granted to user.
       }
     }
 
+    flyway.locations="db/postgresql/migration"
+
 It is also important that any dependent jars are to be added to Job Server class path.
 
 ### Chef
@@ -606,7 +608,9 @@ User impersonation for an already Kerberos authenticated user is supported via `
 
   POST /contexts/my-new-context?spark.proxy.user=<user-to-impersonate>
   
-However, whenever the flag `shiro.use-as-proxy-user` is set to `on` (and authentication is `on`) then this parameter is ignored and the name of the authenticated is *always* used as the value of the `spark.proxy.user` parameter when creating contexts. 
+However, whenever the flag `shiro.use-as-proxy-user` is set to `on` (and authentication is `on`) then this parameter 
+is ignored and the name of the authenticated user is *always* used as the value of the `spark.proxy.user` 
+parameter when creating contexts. 
 
 To pass settings directly to the sparkConf that do not use the "spark." prefix "as-is", use the "passthrough" section.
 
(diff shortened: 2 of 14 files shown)


TimMaltGermany:
@velvia Do you have any idea why there might not be any job slots left? This appears to be some leftover from earlier tests.

velvia:
Not too sure.  Which test is this again?

-Evan

TimMaltGermany:
It is NamedObjectsJobSpec.describe("NamedObjects (DataFrame + RDD)")
Interesting is that the test fails in line 101 which is the 9th job that is started by this test, but maybe that is just a coincidence. Could it be that the previous SQL jobs do not properly finish?

TimMaltGermany:
I increased the max number of jobs per context to 8 for this test. Of course it might be just luck that it did not fail there this time. 
Now it fails at:
[info] - jars should not allow user with valid authorization when timeout *** FAILED ***

[info]   200 OK was not equal to 500 Internal Server Error (WebApiWithAuthenticationSpec.scala:400)
which is the issue reported by @hntd187

TimMaltGermany:
uh, now it fails in JobManagerSpec at:
- should be able to start a job with job jar dependencies and return results *** FAILED ***
...
[info]   at spark.jobserver.JobManagerSpec$$anonfun$2$$anonfun$apply$mcV$sp$22.apply$mcV$sp(JobManagerSpec.scala:220)

hntd187:
@TimMaltGermany I believe the JobManagerSpec test is something that @velvia is disabling in prep for the new Job API, so it's probably okay if you set it to ignore as well.

TimMaltGermany:
Is there a way to manually trigger another build? 
I would like to run a few more tests to make sure that these changes really had an effect.

velvia:
@TimMaltGermany I can restart the build again I think.   However another thing to try is to temporary disable the other tests to see if running fewer tests in that suite really affects it.

Also would be great if you could explain what the authentication test is trying to do underneath the hood, as I don't understand it, hopefully that could give more insight as to why it might fail?

velvia:
Restarted.  Maybe it's a good idea to collapse all the commits and rebase / push -f afterwards.

TimMaltGermany:
The NamedObjectsJobSpec test still fails occasionally due to 'NoJobSlotsAvailable'. Turns out that I didn't change the parameter at all. I will have another look. These tests really shouldn't influence each other.

TimMaltGermany:
@velvia I will consolidate the check-ins once this is done

I have changed a few things around just to see whether that has any effect:

1.  HiveContextFactory in HiveJobSpec needs to stop the SparkContext if it fails to create the Hive context. We might want to add this check to 'spark.jobserver.HiveContextFactory' as well, just in case.

2. NamedObjectsSpec only creates its SparkContext when run and only once.

3. JobSpecBaseBase.afterAll uses TestKit.shutdownActorSystem(system) instead of the ooyala version. It also only shuts down the manager in afterAll (as opposed to after each method).

job-server-extras/test/spark.jobserver/HiveJobSpec.scala
```diff
@@ -9,8 +10,15 @@ import spark.jobserver.context.{HiveContextLike, HiveContextFactory}
...
-    new TestHiveContext(new SparkContext(conf)) with HiveContextLike
+  override protected def contextFactory(conf: SparkConf): C = {
+    val sc = new SparkContext(conf)
+    Try(new TestHiveContext(sc) with HiveContextLike) match {
+      case Success(hc) => hc
+      case Failure(e) =>
+        sc.stop
+        throw e
+    }
+  }
```
TimMaltGermany: This was required as this test fails on my Windows machine and it did not clean-up after itself.

TimMaltGermany:
I give up. Now it fails for JobManagerSpec when trying to retrieve the extra jar path ('getExtrasJarPath').
Is this new or have any of you seen this before as well?

velvia:
Is this the same test that I disabled in the new job API PR?
That one fails due to the extras jar path.  If you run the job-server-extras/test first, then it works.

TimMaltGermany:
@velvia I guess so. Once again, this failure only happens on travis. It only failed in one of 4 tests - so this also seems to be a timing problem. Could you please trigger the build again so that we see what happens? Thanks.

hntd187:
@TimMaltGermany Restarted

hntd187:
LGTM @TimMaltGermany any chance you can squash some of the commits down?

TimMaltGermany:
Done - seems like I also incorporated @noorul 's check-ins - hope that is OK?

TimMaltGermany:
It might be cleaner and easier to just create a new Branch/PR.

noorul:
I think it will be better to rebase and squash

TimMaltGermany:
Well, I did rebase but unfortunately including your last check-ins. Is there a way to undo that?

noorul:
Maybe it will be ideal to have separate branch and pr

TimMaltGermany:
ok, sorry about the confusion. Will do it tomorrow.

TimMaltGermany:
created PR #505 instead

PR #498, opened 2016-06-11
Make dbcp optional

noorul:
Allow user to enable/disable DBCP. By default it is not required as it
is only useful when there is huge number of database connections
involved. Also in some scenarios like when pgpool is running in front of
PostgreSQL DBCP is creating issues. There is no need for client side
pool in such cases.

### DIFF
--- a/job-server/test/spark.jobserver/io/JobSqlDAOSpec.scala
+++ b/job-server/test/spark.jobserver/io/JobSqlDAOSpec.scala
@@ -7,8 +7,12 @@ import spark.jobserver.TestJarFinder
 import com.google.common.io.Files
 import java.io.File
 
-class JobSqlDAOSpec extends TestJarFinder with FunSpecLike with Matchers with BeforeAndAfter {
-  var config = ConfigFactory.load("local.test.jobsqldao.conf")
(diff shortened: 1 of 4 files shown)


job-server/test/spark.jobserver/io/JobSqlDAOSpec.scala
```diff
@@ -8,7 +8,7 @@ import com.google.common.io.Files
 import java.io.File
 
 class JobSqlDAOSpec extends TestJarFinder with FunSpecLike with Matchers with BeforeAndAfter {
-  private val config = ConfigFactory.load("local.test.jobsqldao.conf")
+  var config = ConfigFactory.load("local.test.jobsqldao.conf")
```
velvia: Instead of making this config a var, I think a better practice would be to make a base trait or abstract class called `JobSqlDAOSpecBase`, which defines a config method (`def config: Config`).  Then, have JobSqlDAOSpec extend JobSqlDAOSpecBase and define the config with a val.  Same for JobSqlDAODBCPSpec.  Now you don't need to make the config mutable, which is probably not a good idea.

velvia:
LGTM other than use of var in the test.

Issue #475, opened 2016-05-27
No Logs are generated


jomach:
Hi Guys, 

when I run this in Development mode no log file are generated as say in the docs. 
My Steps: 
sbt
job-server-extras/reStart or job-server/reStart

I get this on the command line : 
job-server-extras[ERROR] log4j:WARN No appenders could be found for logger (spark.jobserver.JobServer$).
job-server-extras[ERROR] log4j:WARN Please initialize the log4j system properly.
job-server-extras[ERROR] log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
job-server-extras[ERROR] Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties

noorul:
It will looks for config/log4j-local.properties

See https://github.com/spark-jobserver/spark-jobserver/blob/master/project/Build.scala#L238

zeitos:
@jomach did noorul's  answer worked? can I close this issue?

jomach:
@noorul: For me the config file is there but still I get no file with name : job-server-local.log

noorul:
@jomach You are right. This is an issue on master branch.

noorul:
Fixed in master

PR #504, opened 2016-06-14
Fix for issue #475

noorul:
It looks like without file protocol override does not happen.


velvia:
Thanks @noorul !    Add yourself to the notes/0.7.0.markdown

PR #505, opened 2016-06-14
flaky-tests:

TimMaltGermany:
Fixes for flaky tests reported by @hntd187 and others. 

1. Timeout test in WebApiWithAuthenticationSpec uses a TimeoutException to ensure that an Exception is always thrown.

2.  NamedObjectsSpec only creates its SparkContext when run and only once.

3. NamedObjectsTestJob sleeps 1 ms so that other threads can do their work and release the job resources.

4. HiveContextFactory in HiveJobSpec needs to stop the SparkContext if it fails to create the Hive context. We might want to add this check to 'spark.jobserver.HiveContextFactory' as well, just in case.

5. JobSpecBaseBase.afterAll uses TestKit.shutdownActorSystem(system) instead of the ooyala version. It also only shuts down the manager in afterAll (as opposed to after each method).

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,7 +608,9 @@ User impersonation for an already Kerberos authenticated user is supported via `
 
   POST /contexts/my-new-context?spark.proxy.user=<user-to-impersonate>
   
-However, whenever the flag `shiro.use-as-proxy-user` is set to `on` (and authentication is `on`) then this parameter is ignored and the name of the authenticated is *always* used as the value of the `spark.proxy.user` parameter when creating contexts. 
+However, whenever the flag `shiro.use-as-proxy-user` is set to `on` (and authentication is `on`) then this parameter 
+is ignored and the name of the authenticated user is *always* used as the value of the `spark.proxy.user` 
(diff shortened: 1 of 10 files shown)


hntd187:
Thanks! @TimMaltGermany pending all the checks passing I'll merge this.

velvia:
@TimMaltGermany thanks so much for the fixes!

noorul:
Awesome!

TimMaltGermany:
@velvia Could you please trigger another build? I would like to see this go through one more time...

hntd187:
@TimMaltGermany Done, one more run then we'll go

hntd187:
Cool, looks good! Thanks again Tim.

noorul:
I think we still have failures. Master is failing.

velvia:
Really?  I checked and master seems to be passing

noorul:
Hmm! I got an e-mail with failure. Now when I look at the master, it looks like passing. Somehow I deleted that mail too :(

PR #506, opened 2016-06-15
Add code coverage to .travis.yml

hntd187:
Add some code coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,5 @@ scala:
 jdk:
    - oraclejdk8
    - oraclejdk7
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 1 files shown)


velvia:
👍 
Maybe these three changes can be squashed down?

hntd187:
I believe I can sqash it down to one commit when it gets merged.

noorul:
I think we should remove job-server-tests folder from this

zeitos:
Nice! 🌟🌟🌟🌟🌟

hntd187:
@noorul why? They are code that is used by people to verify the server and we should verify they work as intended.

noorul:
Agree that whatever code we write should have test cases. But we do not have them in job-server-tests. Why to run coverage on folder in which we have no tests?

hntd187:
@noorul I am also suggesting we write tests for them. 😄 Since they are integration tests they should have unit tests covering them to ensure they don't regress in a new release and can be a reliable verification of the entire system.

noorul:
Agreed!

Issue #517, opened 2016-06-23
Publish Docker with Scala 2.11


mariussoutier:
I want to use Spark Jobserver using Docker, however only the Scala 2.10 version is available. Please create a tag with Scala 2.11.
- [x] Add Scala version tag to docker image
- [x] Read Scala version from environment var
- [x] Use correct spark-job-server binaries
- [x] Instead of fetching pre-built binaries, build Spark with according Scala version
- [ ] Update release process to publish two versions

Thanks!

mariussoutier:
One issue I'm seeing is how to fetch the right Spark version; Spark + Scala 2.11 must be built from source.

janschultecom:
+1

mariussoutier:
See #530

noorul:
Closing

PR #518, opened 2016-06-23
Add Scala version to docker tag

mariussoutier:
Re: #517 - Add Scala version to docker tag in order to build images for Scala 2.10 and Scala 2.11.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -145,7 +145,7 @@ object JobServerBuild extends Build {
     imageNames in docker := Seq(
       sbtdocker.ImageName(namespace = Some("velvia"),
                           repository = "spark-jobserver",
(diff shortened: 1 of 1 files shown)


hntd187:
I don't know how we would verify this won't regress anything. Any suggestions?

velvia:
Well changing the docker tag won’t regress anything so long as we don’t publish any Docker images until the next release.
  
The only consideration is I’m not sure how adding a scala version tag will affect the “latest” tag in Docker.

noorul:
I don't see an issue with pushing a new image with different tag.

mariussoutier:
Right now there is no "latest" tag in Docker.

velvia:
@mariussoutier thanks.  Going to merge this.  However we still need to automate the release process so we can publish two versions.  :-p

mariussoutier:
Sure, there's all some work to be done so it will actually use Spark with Scala 2.11.

Issue #522, opened 2016-06-25
UI and WebAPI not working with latest commit


noorul:
From jobserver mailing list, Will asked

Anyone getting the same issue? Built the latest code on master and noticed 
the following:

Job Server UI no longer shows any of the jobs.
Trying to get job information by sending a get request to /jobs/<jobid> 
leads to:

{
  "result": "No such job ID <jobid>", 
  "status": "ERROR"
}

when the job clearly exists.

noorul:
Fixed!

PR #523, opened 2016-06-25
Fix for issue #522

noorul:
Fix regression caused by f121807

### DIFF
--- a/job-server/src/spark.jobserver/JobInfoActor.scala
+++ b/job-server/src/spark.jobserver/JobInfoActor.scala
@@ -33,38 +33,34 @@ class JobInfoActor(jobDao: JobDAO, contextSupervisor: ActorRef) extends Instrume
 
   override def wrappedReceive: Receive = {
     case GetJobStatuses(limit) =>
-      sender ! jobDao.getJobInfos(limit.get)
+      val originator = sender
+      jobDao.getJobInfos(limit.get).foreach(originator ! _)
 
     case GetJobStatus(jobId) =>
-      val jobInfo = Await.result(jobDao.getJobInfo(jobId), 60 seconds)
-      val resp = if (jobInfo.nonEmpty) NoSuchJobId else jobInfo.get
-      sender ! resp
+      val originator = sender
+
+      jobDao.getJobInfo(jobId).collect {
+        case Some(jobInfo) => originator ! jobInfo
+        case None          => originator ! NoSuchJobId
+      }
 
     case GetJobResult(jobId) =>
-      breakable {
-        val jobInfo: Option[JobInfo] = Await.result(jobDao.getJobInfo(jobId), 60 seconds)
+      val originator = sender
 
-        if (jobInfo.nonEmpty) {
-          sender ! NoSuchJobId
-          break
-        }
-
-        jobInfo.filter { job => job.isRunning || job.isErroredOut }
-          .foreach { jobInfo =>
-            sender ! jobInfo
-            break
+      jobDao.getJobInfo(jobId).collect {
+        case Some(jobInfo) =>
+          if (jobInfo.isRunning || jobInfo.isErroredOut) {
+            originator ! jobInfo
+          } else {
+            // get the context from jobInfo
+            val context = jobInfo.contextName
+            for {
+              resultActor <- (contextSupervisor ? ContextSupervisor.GetResultActor(context)).mapTo[ActorRef]
+              result <- resultActor ? GetJobResult(jobId) } {
+              originator ! result   // a JobResult(jobId, result) object is sent
+            }
           }
-
-        // get the context from jobInfo
-        val context = jobInfo.get.contextName
-
-        val future = (contextSupervisor ? ContextSupervisor.GetResultActor(context)).mapTo[ActorRef]
-        val resultActor = Await.result(future, 3 seconds)
-
-        val receiver = sender // must capture the sender since callbacks are run in a different thread
-        for (result <- resultActor ? GetJobResult(jobId)) {
-          receiver ! result // a JobResult(jobId, result) object is sent
-        }
+        case None => originator ! NoSuchJobId
       }
 
     case GetJobConfig(jobId) =>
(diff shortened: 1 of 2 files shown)


job-server/src/spark.jobserver/JobInfoActor.scala
```diff
@@ -33,18 +33,19 @@ class JobInfoActor(jobDao: JobDAO, contextSupervisor: ActorRef) extends Instrume
 
   override def wrappedReceive: Receive = {
     case GetJobStatuses(limit) =>
-      sender ! jobDao.getJobInfos(limit.get)
+      val jobInfos = Await.result(jobDao.getJobInfos(limit.get), 60 seconds)
```
velvia: This Await is not necessary and just adds more synchronous code with timeouts that will result in a TimeoutError later.  Instead, let's make it reactive:

    val originator = sender
    jobDao.getJobInfos(limit.get).foreach(originator ! _)

This sends the result to the sender whenever it is ready, no need for a timeout, and no need to block the thread of an actor which is a bad design pattern.

velvia:
@noorul how does this fix the issue?  All this does is make a timeout.

Also, we should really avoid using Await whenever we can.  This is a hard-coded timeout which the user cannot control.   The timeout should be controlled in the database driver level, not at this level....  for example if it was a Cassandra backend it should be a read timeout in the Cass driver.

noorul:
This pattern is already used 

https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server/src/spark.jobserver/JobInfoActor.scala#L39

job-server/src/spark.jobserver/JobInfoActor.scala
velvia: Please do not use Await, and please do not introduce another timeout which the user needs to consider.  Use callbacks:

    val originator = sender
    jobDao.getJobInfos(limit.get).foreach(originator ! _)

This gets the job done with NO timeouts and does not block the thread of the JobInfoActor, which is a terrible idea and probably the reason why the bug happened in the first place.

job-server/src/spark.jobserver/JobInfoActor.scala
```diff
@@ -33,18 +33,19 @@ class JobInfoActor(jobDao: JobDAO, contextSupervisor: ActorRef) extends Instrume
...
+      sender ! jobInfos
 
     case GetJobStatus(jobId) =>
       val jobInfo = Await.result(jobDao.getJobInfo(jobId), 60 seconds)
```
velvia: Sorry to put this on you @noorul but please get rid of all other Await.result from this class, and every other class that uses jobDao.  Everything should be Future callbacks, no timeouts.

    jobDao.getJobInfo(jobId).collect {
      case Some(jobId) => originator ! jobId
      case None            => originator ! NoSuchJobId
    }

velvia: Note I didn't test the above out .... I'm assuming Future has a collect method but it might not.  Either way the point is to not do any work after blocking in an actor .... actors should NEVER block.

velvia:
You are right, but that was a mistake I didn’t catch.   Let’s not repeat bad patterns.  Using Await() in an Actor is a complete anti pattern and should never be done.

velvia:
Actors are meant to be event driven and have extremely short message processing loops.  If they block, then no other message can be processed.  Work should be done on other threads.    I can’t seem to find a concise summary of it but it is written in many places...

noorul:
@velvia I tried not to use Await now. Can you take a look?

job-server/src/spark.jobserver/JobInfoActor.scala
```diff
@@ -33,38 +33,39 @@ class JobInfoActor(jobDao: JobDAO, contextSupervisor: ActorRef) extends Instrume
...
 
     case GetJobStatus(jobId) =>
-      val jobInfo = Await.result(jobDao.getJobInfo(jobId), 60 seconds)
-      val resp = if (jobInfo.nonEmpty) NoSuchJobId else jobInfo.get
-      sender ! resp
+      val originator = sender
+
+      jobDao.getJobInfo(jobId).collect {
+        case Some(jobInfo) => originator ! jobInfo
+        case None => originator ! NoSuchJobId
```
velvia: minor ... maybe line up the => ?

job-server/src/spark.jobserver/JobInfoActor.scala
```diff
@@ -33,38 +33,39 @@ class JobInfoActor(jobDao: JobDAO, contextSupervisor: ActorRef) extends Instrume
...
-            break
-          }
+      val originator = sender
 
-        // get the context from jobInfo
-        val context = jobInfo.get.contextName
+      jobDao.getJobInfo(jobId).collect {
+        case Some(jobInfo) =>
+          if (jobInfo.isRunning || jobInfo.isErroredOut)
+          {
```
velvia: (style) let's move the braces to be with the other lines, like

    if (jobInfo.isRunning || jobInfo.isErroredOut) {
      originator ! jobInfo
    } else {

job-server/src/spark.jobserver/JobInfoActor.scala
```diff
@@ -33,38 +33,39 @@ class JobInfoActor(jobDao: JobDAO, contextSupervisor: ActorRef) extends Instrume
...
+          }
+          else
+          {
+            // get the context from jobInfo
+            val context = jobInfo.contextName
 
-        val future = (contextSupervisor ? ContextSupervisor.GetResultActor(context)).mapTo[ActorRef]
-        val resultActor = Await.result(future, 3 seconds)
+            val future = (contextSupervisor ? ContextSupervisor.GetResultActor(context)).mapTo[ActorRef]
+            val resultActor = Await.result(future, 3 seconds)
```
velvia: We can get rid of this Await as well.

    for { resultActor <- (contextSupervisor ? ContextSupervisor.GetResultActor(context)).mapTo[ActorRef]
           result <- resultActor ? GetJobResult(jobId) } {

velvia:
@noorul thanks for adding tests and addressing.  Made some more minor comments, there is another Await we can collapse.   Maybe we can rebase these changes into fewer commits and we can be good to go?

noorul:
@velvia Done!

velvia:
Thanks!

-Evan

PR #525, opened 2016-06-26
yarn-cluster mode

buryat:


velvia:
@buryat I realized you probably included some content that you shouldn't have, but still if you got this to work, we'd be very interested in a proper PR.

PR #137, opened 2015-04-29
Fixes #136 - only use executor uri when provided

nickpoorman:
Fixes #136

Issue #136, opened 2015-04-28
spark.mesos.executor.home


nickpoorman:
Instead of specifying `spark.executor.uri` I've been providing `spark.mesos.executor.home` as a path where the spark executor is on the slave. See: [docs](https://spark.apache.org/docs/latest/running-on-mesos.html#configuration). --ie., the mesos slaves are launched as a docker container with our custom version of spark [included](https://registry.hub.docker.com/u/nickpoorman/docker-spark-postgresql-base/dockerfile/). 

It looks like `spark.executor.uri`  is being set to the empty string here: [server_start.sh#L69](https://github.com/spark-jobserver/spark-jobserver/blob/master/bin/server_start.sh#L69), even when I don't specify it. Then I end up with an error like this:

```
I0428 22:44:45.405141   165 logging.cpp:172] INFO level logging started!
I0428 22:44:45.406445   165 fetcher.cpp:214] Fetching URI ''
E0428 22:44:45.406471   165 fetcher.cpp:176] A relative path was passed for the resource but the slave's frameworks_home flag is not set; please either specify this slave configuration option or avoid using a relative path
Failed to fetch: 
Failed to synchronize with slave (it's probably exited)
```

The error:

```
$ curl d "input.string = a b c a b see" '10.0.0.3:8090/jobs?appName=test&classPath=spark.jobserver.WordCountExample&context=test-context&sync=true'
{
    "status": "ERROR",
    "result": {
        "message": "Job aborted due to stage failure: Task 0 in stage 0.0 failed 4 times, most recent failure: Lost task 0.3 in stage 0.0 (TID 22, 10.0.35.194): ExecutorLostFailure (executor 20150424-134647-1306460170-5050-1-S2 lost)\nDriver stacktrace:",
        "errorClass": "org.apache.spark.SparkException",
[...]

PR #528, opened 2016-06-27
Only set spark.executor.uri if env var is set

addisonj:
This fixes a regression that was fixed in #136 and #137, without this, it breaks if you aren't setting `SPARK_EXECUTOR_URI`


velvia:
@addisonj many thanks!

PR #529, opened 2016-06-28
Added escape to README

Vincibean:
In the README file, it was written that "...it would require serialising Stream[_] blob between processes. For now use Stream[_] job results..." yet, the underscore is a Markdown indicator, used to convey emphasis; the end result was that "] blob between processes. For now use Stream[" was written in italics, and the underscores were hidden. Fixed (escaped).


noorul:
Thank you @Vincibean

PR #530, opened 2016-06-28
Build Spark from source in Docker image

mariussoutier:
This allows to compile for Scala 2.11. To build for Scala 2.11, run `SCALA_VERSION="2.11.8 sbt docker`.

I've already built 0.6.2 images for both Scala versions using this build file:
https://hub.docker.com/r/mariussoutier/spark-jobserver/tags/

The 0.7.0 snapshot has an Akka issue inside Docker which I'm still investigating, but it doesn't seem related to building from source.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -108,6 +108,19 @@ object JobServerBuild extends Build {
     dockerfile in docker := {
       val artifact = (assemblyOutputPath in assembly in jobServerExtras).value
       val artifactTargetPath = s"/app/${artifact.name}"
+
+      val sparkBuild = s"spark-$sparkVersion"
+      val sparkBuildCmd = scalaBinaryVersion.value match {
+        case "2.10" =>
+          "./make-distribution.sh -Phadoop-2.4 -Phive"
+        case "2.11" =>
+          """
+            |./dev/change-scala-version.sh 2.11 && \
+            |./make-distribution.sh -Dscala-2.11 -Phadoop-2.4 -Phive
+          """.stripMargin.trim
+        case other => throw new RuntimeException(s"Scala version $other is not supported!")
+      }
+
       new sbtdocker.mutable.Dockerfile {
         from(s"java:$javaVersion")
         // Dockerfile best practices: https://docs.docker.com/articles/dockerfile_best-practices/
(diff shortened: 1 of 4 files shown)


noorul:
Can you squash them to single commit?

mariussoutier:
Done.

noorul:
1. I think it will be great to take hadoop version also as input.
2. I am not sure whether we should be compiling spark inside docker. Is there a benefit of doing this instead of using binary?

mariussoutier:
> This allows to compile for Scala 2.11.

hntd187:
This is probably not super far off from the docker file I made for this awhile back. https://github.com/hntd187/spark-jobserver-docker/blob/master/Dockerfile

project/Build.scala
```diff
@@ -123,21 +136,28 @@ object JobServerBuild extends Build {
         copy(artifact, artifactTargetPath)
         copy(baseDirectory(_ / "bin" / "server_start.sh").value, file("app/server_start.sh"))
         copy(baseDirectory(_ / "bin" / "server_stop.sh").value, file("app/server_stop.sh"))
+        copy(baseDirectory(_ / "bin" / "manager_start.sh").value, file("app/manager_start.sh"))
```
velvia: Thanks.  I somehow thought we had added this fix before, but must have somehow slipped.... :-p

velvia:
Looks fine to me.   If it is possible to download the binary with Scala 2.11 support also, that would be better, would significantly reduce Docker build times.

mariussoutier:
If that were possible I wouldn't bother to build from Source :)

project/Build.scala
mariussoutier: Actually in 0.6.2 it was present...

hntd187:
Also @velvia soon enough with spark 2.0 scala 2.11 will be the default

project/Build.scala
```diff
@@ -108,6 +108,19 @@ object JobServerBuild extends Build {
     dockerfile in docker := {
       val artifact = (assemblyOutputPath in assembly in jobServerExtras).value
       val artifactTargetPath = s"/app/${artifact.name}"
+
+      val sparkBuild = s"spark-$sparkVersion"
+      val sparkBuildCmd = scalaBinaryVersion.value match {
+        case "2.10" =>
+          "./make-distribution.sh -Phadoop-2.4 -Phive"
```
noorul: What if someone wants to compile it with hadoop 2.6?

velvia: Yes, this is a fair point.  We might as well make this Hadoop version configurable.

mariussoutier: The current docker image is also fixed to Hadoop 2.4 and I didn't want to change everything in one PR. I suggest trying this PR out and see if it fits the build process, then expand on it afterwards (like configuring Hadoop and Java, building SparkR, ...).

noorul:
I will add to notes once #517 is complete.

mariussoutier:
What needs to be done for your release process?

PR #531, opened 2016-06-28
POST /contexts/<contextName> takes typesafe config payload

casey-green:
#### Motivation
I have a lot of spark configs I want to set on my spark context. The only way to do this when creating a context is to supply the configs as URL parameters, but I get a `414: URI Too Long` because there are too many.

#### Solution
Add support for typesafe config in the payload of the `POST /contexts/<contextName>` endpoint. This is the same support for the `POST /jobs` endpoint - the user can supply:
```
spark.context-settings {
  # spark configs
  hadoop {
    # hadoop configs
  }
}

### DIFF
--- a/job-server/src/spark.jobserver/WebApi.scala
+++ b/job-server/src/spark.jobserver/WebApi.scala
@@ -7,7 +7,7 @@ import javax.net.ssl.SSLContext
 import akka.actor.{ActorRef, ActorSystem}
 import akka.pattern.ask
 import akka.util.Timeout
-import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigValueFactory, ConfigRenderOptions}
+import com.typesafe.config.{Config, ConfigException, ConfigFactory, ConfigRenderOptions, ConfigValueFactory}
 import ooyala.common.akka.web.JsonUtils.AnyJsonFormat
 import ooyala.common.akka.web.{CommonRoutes, WebService}
 import org.apache.shiro.SecurityUtils
@@ -227,6 +227,7 @@ class WebApi(system: ActorSystem,
    *    DELETE /data/<filename>       - deletes given file, no-op if file does not exist
    *    POST /data/<filename-prefix>  - upload a new data file, using the given prefix,
    *                                      a time stamp is appended to ensure uniqueness
+   *
    * @author TimMaltGermany
    */
   def dataRoutes: Route = pathPrefix("data") {
@@ -245,7 +246,6 @@ class WebApi(system: ActorSystem,
   def contextRoutes: Route = pathPrefix("contexts") {
     import ContextSupervisor._
 
-    import collection.JavaConverters._
     // user authentication
     authenticate(authenticator) { authInfo =>
       get { ctx =>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: job-server/src/spark.jobserver/WebApi.scala [hunk 1]
```diff
@@ -253,25 +254,35 @@ class WebApi(system: ActorSystem,
...
-                                                            authInfo, contextName)
+                // parse the config from the body, using url params as fallback values
+                val contextConfig = Try {
+                  ConfigFactory.parseString(configString).getConfig("spark.context-settings")
+                } getOrElse {
+                  ConfigFactory.empty
+                } withFallback {
+                  ConfigFactory.parseMap(params.asJava)
+                }
+                val (cName, config) = determineProxyUser(contextConfig.resolve(), authInfo, contextName)
```
velvia: I think it's a tiny bit better to have resolve() be called and assigned to contextConfig, instead of here, in case later someone tries to use it and forgets.

job-server/test/spark.jobserver/WebApiMainRoutesSpec.scala
```diff
@@ -352,6 +352,9 @@ class WebApiMainRoutesSpec extends WebApiSpec {
       Post("/contexts/meme?num-cpu-cores=3&coarse-mesos-mode=true") ~> sealRoute(routes) ~> check {
         status should be (OK)
       }
+      Post("/contexts/meme?num-cpu-cores=3", "spark.context-settings.test=1") ~> sealRoute(routes) ~> check {
```
velvia: Any way to check that test=1 got passed?

velvia:
@casey-green thanks so much, this is awesome and definitely needed.  Just some minor comments.

job-server/src/spark.jobserver/WebApi.scala
casey-green: > I think it's a tiny bit better to have resolve() be called and assigned to contextConfig, instead of here, in case later someone tries to use it and forgets.

Good call, fixed!

job-server/test/spark.jobserver/WebApiMainRoutesSpec.scala
casey-green: > Any way to check that test=1 got passed?

So, I hacked the checks into `DummyActor`, which seems... a little weird. Let me know if you're OK with this, or have another idea of how to do it. I'm not super familiar with akka, so I might be missing the easier way to do it.

noorul:
Great! My team members were asking why this is not yet supported in SJS

job-server/test/spark.jobserver/WebApiMainRoutesSpec.scala
casey-green: After understanding the flow a bit more, this is another approach to testing this: https://github.com/casey-green/spark-jobserver/compare/casey/context-spark-config-in-body...casey-green:casey/context-spark-config-in-body-alt-test-approach?expand=1

velvia:
👍

job-server/test/spark.jobserver/WebApiMainRoutesSpec.scala
velvia: Merged, thanks again!

PR #532, opened 2016-06-28
Make JMX port configurable

casey-green:
Also, log command output somewhere in case it fails.


bin/server_start.sh
```diff
@@ -48,7 +48,7 @@ cmd='$SPARK_HOME/bin/spark-submit --class $MAIN --driver-memory $JOBSERVER_MEMOR
   --driver-java-options "$GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES"
   $@ $appdir/spark-job-server.jar $conffile'
 if [ -z "$JOBSERVER_FG" ]; then
-  eval $cmd > /dev/null 2>&1 < /dev/null &
+  eval $cmd > $LOG_DIR/server_start.sh.log 2>&1 < $LOG_DIR/server_start.sh.log &
```
noorul: 1. I think you should < /dev/null as such. It has a purpose.
2. I think we can have server_start.log

casey-green: done

noorul:
Merged! Thank you @casey-green

Issue #507, opened 2016-06-17
Rest API delete file


big8extreme:
Hi all,
We send file to jobserver via the Rest API and when we need to delete them 

with `GET /data/` i have this
`
[
"/data/file/test-17T08_45_15.492Z.dat"
]
`
and then when i try to delete the file with 
`DELETE /data/"/data/file/test-17T08_45_15.492Z.dat"
or
DELETE /data/data/file/test-17T08_45_15.492Z.dat`
this dont work.

hntd187:
Because there is no ability to delete jars currently. You must delete them physically on disk. This is to prevent people from accidentially deleting things by mistake.

hntd187:
Sorry read this wrong initially, do the logs say anything?

noorul:
@big8extreme I think you should try "DELETE /data/file/test-17T08_45_15.492Z.dat"

noorul:
@big8extreme Did you try it?

big8extreme:
Yes it didn't work, i try everything

noorul:
Should work now on master. See the updated readme for example

PR #538, opened 2016-07-02
Fix for Issue #507: Fix DELETE /data/<filename> issue

noorul:
Currently data file delete api expects encoded URI. Enhance it to use
the path as such. This would help the user not to encode path.

Added examples

### DIFF
--- a/job-server/src/spark.jobserver/routes/DataRoutes.scala
+++ b/job-server/src/spark.jobserver/routes/DataRoutes.scala
@@ -40,7 +40,7 @@ trait DataRoutes extends HttpService {
       // DELETE /data/filename delete the given file
       delete {
         path(Segment) { filename =>
-          val future = dataManager ? DeleteData(URLDecoder.decode(filename, "UTF-8"))
+          val future = dataManager ? DeleteData(filename)
           respondWithMediaType(MediaTypes.`application/json`) { ctx =>
             future.map {
               case Deleted => ctx.complete(StatusCodes.OK)
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Also see [Chinese docs / 中文](doc/chinese/job-server.md).
   - [Contexts](#contexts)
   - [Jobs](#jobs)
   - [Data](#data)
+    - [Data API Example](#data-api-example)
   - [Context configuration](#context-configuration)
   - [Other configuration settings](#other-configuration-settings)
   - [Job Result Serialization](#job-result-serialization)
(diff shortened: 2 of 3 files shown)


velvia:
This should be reviewed by the guys at KNIME -- /cc @TimMaltGermany

TimMaltGermany:
@noorul What about file names that are not proper URIs? Such as:
curl -X POST http://localhost:8090/data//tmp/spark-jobserver/upload/test data file upload.txt
or our beloved German Umlauts:
curl -X POST http://localhost:8090/data//tmp/spark-jobserver/upload/täst-töff.txt

Do these still work? 
I think that when I originally implemented this I added the requirement that the names in the POST request are URI-encoded so that users can use any filename they like.

job-server/test/spark.jobserver/routes/DataRoutesSpec.scala
```diff
@@ -39,8 +39,7 @@ class DataRoutesSpec extends WebApiSpec {
     }
 
     it("DELETE - should be able to remove file from tmp dir") {
-      val encodedName = URLEncoder.encode("/tmp/fileToRemove", "UTF-8")
-      Delete("/data/" + encodedName) ~> sealRoute(routes) ~> check {
+      Delete("/data//tmp/fileToRemove") ~> sealRoute(routes) ~> check {
```
TimMaltGermany: I think you still need to encode the filename here. The handler in WebApiSpec must be changed instead.
Simplest solution would be this:
instead of 
`    case DataManagerActor.StoreData("/tmp/fileToRemove", _) => sender ! DataManagerActor.Stored("/tmp/fileToRemove-time-stamp")`
`      case DataManagerActor.StoreData("errorfileToRemove", _) => sender ! DataManagerActor.Error`

it should be 
`    case DataManagerActor.StoreData("errorfileToRemove", _) => sender ! DataManagerActor.Error`     
`     case DataManagerActor.StoreData(filename, _) => sender ! DataManagerActor.Stored(filename + "-time-stamp")`

and similiarly for the DELETE cases.

hntd187:
It also seems kind of not DRY to have the double slash but not much you could really do there other than encode it.

noorul:
@TimMaltGermany Incorporated review comments.

noorul:
@TimMaltGermany Can you take a look again?

noorul:
@hntd187 @velvia It looks like @TimMaltGermany  is busy. Can you guys review this one?

job-server/src/spark.jobserver/routes/DataRoutes.scala
```diff
@@ -40,7 +40,7 @@ trait DataRoutes extends HttpService {
       // DELETE /data/filename delete the given file
       delete {
         path(Segment) { filename =>
-          val future = dataManager ? DeleteData(URLDecoder.decode(filename, "UTF-8"))
+          val future = dataManager ? DeleteData(filename)
```
hntd187: Can we have the response return valid JSON instead of just the OK like #545

TimMaltGermany:
+1 Sorry, never saw your commit.

Issue #536, opened 2016-06-30
Mark killed jobs as failed instead of completed


leoromanovsky:
When using the API to kill a job, such as:

`curl -X DELETE http://url/jobs/5450e2ca-468b-486e-ae15-bcfd62345d85`

The job as marked as "completed", which makes monitoring and alerting more complicated, since the job has not completed successfully. In this case, shouldn't it be considered failed?

noorul:
I think "KILLED" should be more appropriate.

kishorez:
Just want to add on top of it. Not sure if it should be a different issue.

Env - Spark Job Server, Yarn, Spark 1.6

When we delete a job using curl we noticed the job entry being deleted in spark job server but we still see the respective job spark context running when we look in the resource manager. Unless we delete the context in the spark job server the job is not being completely deleted in the yarn.

noorul:
@leoromanovsky  Is this related to https://github.com/spark-jobserver/spark-jobserver/issues/222 ?

velvia:
@kishorez …. is this an ad hoc job or a long running context?  For long running contexts the behavior you describe is completely reasonable … you don’t want to delete the context if the job is deleted.

leoromanovsky:
I haven't personally had any problems with the job itself not terminating, this issue is mostly a discussion around the API behavior of when a job is forcefully terminated.

For further background, I'm developing a monitoring solution @strava to alert when jobs have not succeeded after a certain period of time. Jobs that are purposefully killed are currently recorded as having been `COMPLETED` by SJS which pushes back the time that an alert should have surfaced.

I'll get a PR together to address this.

kishorez:
@velvia Yes. you are right this is a streaming job (continuous and long running). can you share the expected behavior if a job(long running) is deleted by an rest http delete. I presumed when a job is deleted it should delete the respective job running in the cluster. (Ex: Yarn)

@noorul Yes you are right it looks the same.

Documenting this behavior will be helpful

noorul:
Closing

PR #541, opened 2016-07-11
Issue #536: Web API returns status "KILLED" for jobs cancelled by user.

leoromanovsky:
The goal of this PR is to fix: https://github.com/spark-jobserver/spark-jobserver/issues/536

`DELETE` requests from the web API will save an exception with the message `killed` in the job model. The API will return status `KILLED` for this class of exceptions.

The code to accomplish this is clumsy and works with the fact that the API attributes an exception in the Job model as an indication of failure. If this code, after I add tests, is acceptable to the project maintainers, I'd be ok with merging it.

However - I think a better long-term refactor is to add a column to the Job model with the status explicitly set, instead of assuming success or failure from other columns: https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server/src/spark.jobserver/WebApi.scala#L93

Either way, if this PR is accepted, I'll create a follow on one to update the UI with a new table section for `KILLED` jobs.

Thanks for your feedback.

### DIFF
--- a/job-server/src/spark.jobserver/CommonMessages.scala
+++ b/job-server/src/spark.jobserver/CommonMessages.scala
@@ -15,7 +15,7 @@ object CommonMessages {
   case class JobFinished(jobId: String, endTime: DateTime) extends StatusMessage
   case class JobValidationFailed(jobId: String, endTime: DateTime, err: Throwable) extends StatusMessage
   case class JobErroredOut(jobId: String, endTime: DateTime, err: Throwable) extends StatusMessage
-  case class JobKilled(jobId: String, endTime: DateTime) extends StatusMessage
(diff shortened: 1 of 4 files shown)


zeitos:
Hi @leoromanovsky thanks for taking the time to fix this!
   Would you mind adding a test for those new lines in WebApi.scala ? (killed and error)?

leoromanovsky:
@zeitos Ok will do

noorul:
@leoromanovsky We now have JobStatus for status values to avoid hardcoding. If you re-base against master you will get it. So you could use JobStatus.Killed instead of "KILLED"

velvia:
There is apparently a new PR to cover this, so closing

PR #544, opened 2016-07-11
Updated Spark version

hntd187:
Updated Spark version to 1.6.2.

Since this is a minor update, it should be backward compatible. Since we've been using the 1.6.x branch for sometime, this should just fix bugs in Spark.


velvia:
Restarted the build.    Do any docs or configs or the Docker stuff need to be updated?

hntd187:
The current docker version is based on that spark version. I went back and rechecked it in the code again to make sure it was the case. I think I might need to update a doc, lemme get on that.

zeitos:
@hntd187 no docker update needed? can I merge it?

hntd187:
I think this is fine now. There are no docker changes because that image is based on that dependency version so if you trace the usage you'll see if nothing is provided it will use that.

Issue #353, opened 2015-12-30
Proper reconnection / HA for forked JVM contexts


velvia:
Right now with the JVM-per-context PR merged, here is one problem that remains.  When the forked JVM contexts are not explicitly killed, and a user runs `server_stop.sh`, then only the HTTP WebApi process is killed, not the child JVMs.  This is a problem because when the process is re-launched, it is not able to re-connect with the forked JVMs.  While the master HTTP server is down, the forked manager JVM is not able to persist status changes, because the JobDAO is only run on the HTTP server.

While we could have the forked JVMs kill themselves when the master goes down, it seems much better to be able to re-start the HTTP component and re-connect to running contexts.   Thus the goals of this issue are:
1. If the HTTP process ("master") goes down, and comes back up, it is able to discover currently running forked JVM contexts and automatically add them to the list of current contexts.  This may mean that contexts that are auto-started do not need to be started up anymore.
2. If the forked JVM process goes down, or heartbeat dies, it is properly recognized by the HTTP master and marked as down
3. If the forked JVM process goes down, and comes back up, the HTTP master recognizes it
4. When the HTTP process ("master") is down, the forked JVM contexts should still be able to update status in the DB.
5. All of the above should be testable in the regular test suite, probably using sbt-multi-jvm and the Akka clustering testing plugin.

Some implications of the above:
- Forked JVM processes must have their own JobDAO.  I would in fact eliminate the JobDAOActor.  This satisfies the ability to update when the HTTP master is down.
- Discovery of forked JVMs implies that state about existing running contexts (as well as their port #'s or Akka connection info) must be persisted somewhere.
- Need to investigate if we need to connect with just one member of the remaining cluster or how many (or all of them).

velvia:
Reported by Nick Xie (thanks!):

I did a small set of testing.  Here are what I've found.

Methodology:

1. create two context, one is predefined in conf and the other one manually created.
2. put a bunch of jobs into non-stop loops in two script files.  The jobs are mostly spark mllib operations and reduce/group by on data from a couple of MB to 1GB.
3. run the two scripts continuously for about an hour.

Issues found:

1. after a while the predefined context died, unclear why.  The underlying yarn application seem to be gone.  But the job server is unaware and returning Akka timeout errors.   

2. While the other context is fine, the web UI show many of the jobs as STILL running when they completed long ago.  There seem to be some state update issues.

[...]

PR #527, opened 2016-06-27
Fix #526: Deprecate JobFileDAO

noorul:

PR #551, opened 2016-07-30
Partial fix for issue #353

noorul:
Forked JVM processes must have their own JobDAO

### DIFF
--- a/job-server/src/spark.jobserver/JobManager.scala
+++ b/job-server/src/spark.jobserver/JobManager.scala
@@ -44,12 +44,18 @@ object JobManager {
     } else {
       defaultConfig
     }
+
+    val system = makeSystem(config.resolve())
+    val clazz = Class.forName(config.getString("spark.jobserver.jobdao"))
+    val ctor = clazz.getDeclaredConstructor(Class.forName("com.typesafe.config.Config"))
+    val jobDAO = ctor.newInstance(config).asInstanceOf[JobDAO]
+    val daoActor = system.actorOf(Props(classOf[JobDAOActor], jobDAO), "dao-manager-jobmanager")
+
     logger.info("Starting JobManager named " + managerName + " with config {}",
       config.getConfig("spark").root.render())
     logger.info("..and context config:\n" + contextConfig.root.render)
 
-    val system = makeSystem(config.resolve())
-    val jobManager = system.actorOf(JobManagerActor.props(contextConfig), managerName)
+    val jobManager = system.actorOf(JobManagerActor.props(contextConfig, daoActor), managerName)
 
     //Join akka cluster
     logger.info("Joining cluster at address {}", clusterAddress)
--- a/job-server/src/spark.jobserver/JobServer.scala
+++ b/job-server/src/spark.jobserver/JobServer.scala
@@ -59,10 +59,11 @@ object JobServer {
       // Check if we are using correct DB backend when context-per-jvm is enabled.
       // JobFileDAO and H2 mem is not supported.
       if (contextPerJvm) {
-        if (clazz.getName != "spark.jobserver.io.JobSqlDAO") {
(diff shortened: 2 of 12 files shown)


noorul:
@hntd187 @velvia Can you take a look? Code coverage seems to falling short. I think that is fine in this case.

job-server/src/spark.jobserver/JobManager.scala
```diff
@@ -44,12 +44,18 @@ object JobManager {
     } else {
       defaultConfig
     }
+
+    val system = makeSystem(config.resolve())
+    val clazz = Class.forName(config.getString("spark.jobserver.jobdao"))
+    val ctor = clazz.getDeclaredConstructor(Class.forName("com.typesafe.config.Config"))
+    val jobDAO = ctor.newInstance(config).asInstanceOf[JobDAO]
```
velvia: If this code is repeated in the WebServer / base JVM, it should be moved to an Object.

velvia:
@noorul I think this is a good start, but before merging, I think we should

1) Add a check that when context-per-jvm is true, we cannot start the FileDAO or the embedded H2Dao
2) Switch the default from FileDAO to H2Dao

Also, personally I think that the JobDaoActor is not even needed once you have each JobManager have their own DAO instance.  This was the original design before context-per-jvm, and the only reason why everything was moved to an actor was that the JobDAO was present only in the server JVM.   This would simplify a LOT of code.

That said, one way we can make the embedded H2 DAO and FileDAO work even for context-per-jvm is to pass along the server JVM's copy of it.  That would of course continue the timeout issues we are seeing.

Good work.

job-server/src/spark.jobserver/JobManager.scala
noorul: The only change I am making here is to pass jobDAO. I am not sure what you mean by moved to an object.

noorul:
Switch the default from FileDAO to H2DAO : Do you think it is better to give a sample config for multi-jvm instead of making H2DAO as default in config? Because in #527 , you mentioned that we need to keep FileDAO in config as default.

noorul:
I think we can take up removing JobDAOActor as a separate activity.

velvia:
@noorul no I don't think I meant to keep FileDAO as a default.  I mean that people who are still using FileDAO should not suddenly have their job servers nonoperational after upgrading, but we should definitely change the default; otherwise how can we make progress?

noorul:
@velvia, Then I will make H2 as default as part of JobFileDAO deprecate PR

noorul:
@velvia Made H2 default in https://github.com/spark-jobserver/spark-jobserver/pull/527

noorul:
@velvia Can you take a look now?

job-server/src/spark.jobserver/JobServer.scala
```diff
@@ -55,12 +55,21 @@ object JobServer {
     val clazz = Class.forName(config.getString("spark.jobserver.jobdao"))
     val ctor = clazz.getDeclaredConstructor(Class.forName("com.typesafe.config.Config"))
     try {
+      val contextPerJvm = config.getBoolean("spark.jobserver.context-per-jvm")
+      // Check if we are using correct DB backend when context-per-jvm is enabled.
+      // JobFileDAO and H2 mem is not supported.
+      if (contextPerJvm) {
+        if (clazz.getName != "spark.jobserver.io.JobSqlDAO") {
```
velvia: I think instead of whitelisting JobSqlDao, we could blacklist the JobFileDao.   There is a PR for a Cassandra DAO, which would not be a SqlDao I think.

velvia:
Ok added a comment.

noorul:
Thank you @velvia . I incorporated review comment. Merging.

velvia:
Thanks @noorul

-Evan

PR #554, opened 2016-08-03
Remove hardcoded status values and use constants

noorul:


hntd187:
A style nitpick but a constant should be all capital letters. I doubt the world will end over this.

velvia:
👍

noorul:
@hntd187 But scala style guide says something different http://docs.scala-lang.org/style/naming-conventions.html

PR #561, opened 2016-08-08
New Java Api

hntd187:
Okay, so here is the first working version of the new Java API. Your first thought is probably wow, that's a lot of changes for any Api and you'd be correct in such an assumption. For awhile our project structure was not standard, so a lot of these changes are really just moving folders around. Here is the list of significant changes.

1. All projects have been moved into a /project/src/main/{scala,java} style and the package heirarchy has been cleaned up (spark.jobserver as one folder turns into /spark/jobserver) to make java work correctly because Java requires a specific folder structure for it's files.
2. The `ooyala.common.akka` package has been renamed to the `spark.jobserver.akka` package for consistency with the rest of the project.
3. There is simple documentation of the new Java Api in docs/java-api.md for the newly initiated.
4. Job Caching is now a configurable option in the application.conf. Users can put their options in spark.job-cache and these options are passed into the cache object, which is loaded at runtime. There is also now a generate cache trait for implementations of this. I originally had written a redis caching client as well, but there were a million serialization issues, so I abandoned it.
5. `JobContainer` has been paramaterized around the job that it will load and now all the job containers share a base trait.
6. `SparkContextFactory` now has a type for the job as well as the job container for the loadAndValidate method to be able to know what it's doing.
7. ContextFactories and Job imples have been created for all the alternative contexts in the extras project.
8. All the jobserver-tests have been written in equivalent Java. To the best of my knowledge they should function roughly the same.

I know there is a lot here, so it's probably to be caution and question everything is in order for this merge. Also, if you are wondering why a lot of the imports are changed, I ran an import optimize on the projects to remove unused imports in a lot of places. I also fixed a lot of small warnings and code style twitches along the way as well in my never ending quest to clean up tech debt.

Also @noorul I think I need to rename some things in the project for consistency, can you tell me what those are?


velvia:
Awesome!   Looking forward to it.  

-Evan

hntd187:
@velvia if you want I can give you a run down on how it's designed in more details if you want to have a call or a more detailed write up

velvia:
@hntd187 yeah a write up would be great.  Maybe write up since then other folks can have a look too.

@noorul and others I think since there are lots of changes including renames in here, it's better to merge the Python changes first... but I haven't finished looking through that change either  :-p

velvia:
@hntd187 sorry haven't had time to look into this.  A writeup would be great, or simply a pointer to the commits which isolate the actual API changes.   I might pull down the branch and have a look at the files, it's probably easier than looking through Github.

hntd187:
@velvia well I still have to merge from master, so it might not even be worth it right now because probably some of it will change a bit as I rectify all the merge conflicts.

hntd187:
@velvia @noorul Could I get some reviews on this so I can resolve the merge conflicts and possible changes at the same time?

velvia:
@hntd187 had a really quick look at README and a few of the Java classes.  Looks great to me.   I didn't see mention of NamedObjects - the assumption is the user accesses namedObjects through JobEnvironment, right?    I'm OK with merging followed by incremental changes.

noorul:
I have not got the time to review this. I trust Evan. Let us merge this and have incremental changes.

msohn:
there are still some packages left not following the standard Java package layout:
- ./job-server-extras/src/main/scala/spark.jobserver
- ./job-server-extras/src/test/scala/spark.jobserver
if you could also fix these (replacing spark.jobserver with spark/jobserver) this would help to enable using Eclipse for spark-jobserver development (see #619)

hntd187:
@noorul @velvia okay I spoke with Evan, I'm gonna close this request and put in 2 separate ones

1. for the folder format update
2. the Java API

I should have these shortly.

PR #571, opened 2016-08-18
Include Python exception stacktrace on failure

CalebFenton:
Debugging exceptions was very difficult without the exception stack trace. There was also a whitespace missing.


noorul:
Merged! Thank you @CalebFenton

PR #573, opened 2016-08-19
Add missing imports for Python contexts

CalebFenton:
This is a fix for https://github.com/spark-jobserver/spark-jobserver/issues/572.


velvia:
@CalebFenton many thanks, this looks great to me.

@mattinbits @noorul and others any thoughts?  Otherwise will merge it soon.

velvia:
@noorul @hntd187 what did we decide with this?   Add support or document and have user start the right context?  I think latter is best for consistency, but if folks really think merging is better for usability, I'm not against that either.

velvia:
@CalebFenton going to merge this, sorry this lapsed for so long.

PR #574, opened 2016-08-21
Issue #536: Web API returns status "KILLED" for jobs cancelled by user.

f1yegor:
Merged PR #541 because of inactivity.
Of course all honour to @leoromanovsky

### DIFF
--- a/job-server/src/spark.jobserver/JobManagerActor.scala
+++ b/job-server/src/spark.jobserver/JobManagerActor.scala
@@ -25,6 +25,8 @@ object JobManagerActor {
   case class StartJob(appName: String, classPath: String, config: Config,
                       subscribedEvents: Set[Class[_]])
   case class KillJob(jobId: String)
+  case class JobKilledException(jobId: String) extends Exception(s"Job $jobId killed")
+
   case object GetContextConfig
   case object SparkContextStatus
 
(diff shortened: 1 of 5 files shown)


job-server/src/spark.jobserver/JobManagerActor.scala
```diff
@@ -158,7 +158,7 @@ class JobManagerActor(contextConfig: Config) extends InstrumentedActor {
 
     case KillJob(jobId: String) => {
       jobContext.sparkContext.cancelJobGroup(jobId)
-      statusActor ! JobKilled(jobId, DateTime.now())
+      statusActor ! JobKilled(jobId, DateTime.now(), new Exception("killed"))
```
velvia: Why don't we use a specific SJS exception instead of generic one with a message?  Matching on types is much safer than matching on strings .... though I agree the probability of a user doing this is pretty low.

velvia:
@f1yegor many thanks.   Here is my suggestion.

- The JobKilled case class for the message doesn't need to be extended with an exception/throwable.  It is only created in one place, in JobManagerActor when the user wants a job killed, and since there is no real exception associated with a job, this seems like a waste to add another field.  So my recommendation is to remove that.

- In JobStatusActor, instead of copying the exception from JobKilled, which is not necessary, just create a unique exception, let's call it "JobKilledException", for example this can be

        case class JobKilledException(jobId: String) extends Exception(s"Job $jobId killed")

- WebApi matches on JobKilledException, which is much safer than matching on the exception message

What do you think?

f1yegor:
fixed

velvia:
@f1yegor this is almost what I meant.  There is no need for JobKilled message to have a throwable field.  I suggest this:
- remove throwable field from JobKilled
- Have JobStatusActor explicitly create the JobKilledException

Otherwise looks good to me....   others can merge

velvia:
@f1yegor many thanks, merging

PR #584, opened 2016-09-07
Convert SparkHiveJob to use new API

CalebFenton:
I wanted to use Hive jobs but it _looks_ like the only way to do that is by converting my jobs to the old, less type safe API. So I made this.

I would've converted SQL jobs also but I wasn't sure if there was a reason why SparkHiveJob used the older API.

### DIFF
--- a/job-server-extras/src/spark.jobserver/HiveTestJob.scala
+++ b/job-server-extras/src/spark.jobserver/HiveTestJob.scala
@@ -1,14 +1,16 @@
 package spark.jobserver
 
-import com.typesafe.config.{Config, ConfigFactory}
-import org.apache.spark._
+import com.typesafe.config.Config
+import org.apache.spark.sql.Row
 import org.apache.spark.sql.hive.HiveContext
+import org.scalactic._
+import spark.jobserver.api.{JobEnvironment, ValidationProblem}
 
 /**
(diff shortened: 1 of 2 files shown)


CalebFenton:
I'd also be really curious to know why there's a `spark.jobserver.SparkJobBase` and a `spark.jobserver.api.SparkJobBase` that are very similar.

velvia:
My only comments about this is whether folks are OK having the base HiveJob convert completely to the new API.   The `spark.jobserver.SparkJobBase` is the old API and is there so as not to make existing jobs not compilable.   I slightly favor having the updated API in a new namespace, but if everyone thinks it's OK to just switch then I'm OK too.

@hntd187 @noorul @zeitos what do you guys think?

hntd187:
The new Java API is based on the new API, I don't think there is any other compelling reason I can think of to keep it around.

velvia:
Ok.

We should convert the other jobs, streaming, sql, etc.too.

People with existing jobs will probably complain.

hntd187:
@velvia Well, 0.6.2 should still work for them, so if they want to go to 0.7 they should upgrade. At least we aren't just throwing it on them.

velvia:
Ok let's merge this, then we can merge @hntd187 's project changes and then we can update the other extras contexts.

PR #599, opened 2016-09-21
Use relative paths so it works if running in a context-path

sjoerdmulder:
We try to run the job-server behind reverse proxy with authentication as https://admin.example.org/spark-job-server/ but since the JSON is hardcoded to fetch root paths it will return 404. This change makes sure it requests the json relative to the root of the job-server directory


noorul:
will this break non-proxy setup?

sjoerdmulder:
No it should not. Since non-proxy setup runs f.e. on `http://192.168.1.2:8090/` it will still fetch `http://192.168.1.2:8090/jobs` as it would. Currently in the index.html the paths to CSS / JS are already relative, so those would then also be already broken ;)

PR #620, opened 2016-10-07
Ensures that the same version of scala-compiler is used as the versio…

aganim-chariot:
…n of scala-lang.  Without this, under scala 2.10, an old version of scala-compiler is brought in as a transitive dependency and not evicted


velvia:
@aganim-chariot thanks

aganim-chariot:
@velvia No problem.  Sorry it took me a couple of days before I followed up on it!

PR #624, opened 2016-10-12
Change cluster status of removed contexts to down (fixes #466).

### LINKED ISSUE #466: Timeout exception on create context when a delete context operation is executed
Steps to reproduce:

1. Create context <context-test-1>.
2. Delete context <context-test-1> and at the same time trigger another context <context-test-2> creation.  Check the logs and you can see that the context was created as expected, but the API call returns => "The server was not able to produce a timely response to your request."

NOTE: Spark-jobserver was deployed with spark.master = "yarn-client" and spark.jobserver.context-per-jvm = true
[timeoutException.txt](https://github.com/spark-jobserver/spark-jobserver/files/270606/timeoutException.txt)

derSascha:
Running jobserver with context-per-jvm results in remote jobmanager
actors, manually joining the jobserver cluster. After removing a
context, the actor stops and becomes unreachable within the cluster. We
have to manually remove the actor from the cluster (status down) before
the cluster accepts new clients again.

See http://doc.akka.io/docs/akka/2.3.15/scala/cluster-usage.html#Automatic_vs__Manual_Downing

### DIFF
--- a/job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/spark.jobserver/AkkaClusterSupervisorActor.scala
@@ -149,7 +149,9 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
     case StopContext(name) =>
       if (contexts contains name) {
         logger.info("Shutting down context {}", name)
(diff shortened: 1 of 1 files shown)


velvia:
@derSascha many thanks!

PR #627, opened 2016-10-13
Changed job-binary-paths to job-bin-paths in job-server/src/main/reso…

nikkiSproutling:
…urces/application.conf.


velvia:
@nikkiSproutling thanks!

PR #638, opened 2016-10-27
Previously, the OK responses returned a string equal to OK that did n…

roarking:
…ot comply to the application/json content-type. This was causing any client libraries to break. What I did was a simple (possible temporary) fix which modified all complete(StatusCodes.OK) in the WebApi class to return complete(StatusCodes.OK,Map(statusKey,resultKey). This way, the 200 responses will comply with the JSON content type.


velvia:
Many thanks…. will have a look later

velvia:
@roarking thanks, this LGTM.

Issue #644, opened 2016-11-03
akka-remote not excluded during assembly


jonbuffington:
Ran into a class verifier exception yesterday deploying SJS to a new cluster. Tracked it down to a bug/missing sbt feature in the exclude method <https://github.com/sbt/sbt/issues/1518>. `"com.typesafe.akka" %% "akka-cluster" % "2.3.15" exclude("com.typesafe.akka", "akka-remote")` does not exclude akka-remote leading to an old version of netty being assembled into the jar.

My crude work around was <https://github.com/fuseelements/spark-jobserver/commit/1ecee0a3f124f9d86eeb1e07ddc4904b9c34ac0d>. Any suggestions for creating a better commit and opening a pull request?

velvia:
:(  bummer that this hasn’t been fixed in SBT yet.   Yuck.

I can’t think of a much better way, but there is a way to reference the current Scala version, have you tried scalaVersion.value ?

Actually, instead of excluding akka-remote (which is in Spark itself too), what if we excluded Netty instead?

jonbuffington:
@velvia 

`> project akka-app`

Using `com.typesafe.akka" %% "akka-cluster" % "2.3.15" exclude("com.typesafe.akka", "akka-remote")`:
```
> dependencyTree
[warn] Credentials file /Users/jcb/.bintray/.credentials does not exist
[info] Updating {file:/Users/jcb/Projects/spark-jobserver/}akka-app...
[info] Resolving com.googlecode.concurrentlinkedhashmap#concurrentlinkedhashmap-[info] Resolving com.sun.jersey.jersey-test-framework#jersey-test-framework-griz[info] Resolving com.fasterxml.jackson.module#jackson-module-scala_2.10;2.4.4 ..[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] spark.jobserver:akka-app_2.10:0.6.2 [S]
[info]   +-com.typesafe.akka:akka-cluster_2.10:2.3.15 [S]
[info]   | +-com.typesafe.akka:akka-remote_2.10:2.3.15 [S]
[info]   |   +-com.google.protobuf:protobuf-java:2.5.0
[info]   |   +-com.typesafe.akka:akka-actor_2.10:2.3.15 [S]
[info]   |   | +-com.typesafe:config:1.2.1 (evicted by: 1.3.0)
[info]   |   | +-com.typesafe:config:1.3.0
[info]   |   | 
[...]

PR #647, opened 2016-11-05
fix: exclude org.jboss.netty classes

jonbuffington:
The akka-remote transitive dependency brings an older io.netty jar
into the assembly causing an occasional Java verifier exception when
starting SJS.

See <https://github.com/sbt/sbt/issues/1518> why the simple exclude of
akka-remote does not work.

Per discussion in https://github.com/spark-jobserver/spark-jobserver/issues/644

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -17,7 +17,7 @@ object Dependencies {
     // Akka is provided because Spark already includes it, and Spark's version is shaded so it's not safe
     // to use this one
     "com.typesafe.akka" %% "akka-slf4j" % akka % "provided",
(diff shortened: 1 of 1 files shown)


velvia:
@jonbuffington did you mean to close it?   It looks fine to me.

jonbuffington:
@velvia I found excluding io.netty does not pass tests. Looks like an implicit dependency on including older io.netty.

The alternative is <https://github.com/fuseelements/spark-jobserver/commit/ac4bb5a382e84c20ed523fa6811fb244b8792f58>. Excluding akka-remote does pass tests locally but not on Travis CI. I am trying to determine why.

PR #648, opened 2016-11-09
Send result before finish event in job manager actor

derSascha:
We poll the REST API for job status/result and sometimes the json contains status finished but no result. 
This happens if the job manager send a status update (finished) but did not yet finished the results transfer.

This PR fix the send order within the job manager actor and improves the web api error handling.

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -305,7 +305,6 @@ class JobManagerActor(contextConfig: Config, daoActor: ActorRef) extends Instrum
       }
     }(executionContext).andThen {
       case Success(result: Any) =>
-        statusActor ! JobFinished(jobId, DateTime.now())
         // TODO: If the result is Stream[_] and this is running with context-per-jvm=true configuration
         // serializing a Stream[_] blob across process boundaries is not desirable.
         // In that scenario an enhancement is required here to chunk stream results back.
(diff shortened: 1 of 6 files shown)


job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -497,6 +497,9 @@ class WebApi(system: ActorSystem,
                           resultToByteIterator(jobReport, s.toIterator))
                       case _ => ctx.complete(jobReport ++ resultToTable(result))
                     }
+                  case NoSuchJobId =>
+                    ctx.complete(500, errMap("Status, but no result with job ID "
```
zeitos: I think 500 wouldn't be the best error code here since there's no error.  In any case 503 which means a temporary error might be better. But I would rather reply a 200 with a text specifying that they should retry

derSascha:
There are two cases where the job status is finished but the result is empty:
1. job finished right now, result is in transfer (try again later)
2. job finished time ago or many concurrent jobs, result removed from result cache (result lost, restart job)

Sending the status after the result works well in my tests. I updated the PR and reverted the changes on the web API to handle the second case and added a block in the troubleshooting guide.

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
```diff
@@ -316,6 +315,7 @@ class JobManagerActor(contextConfig: Config, daoActor: ActorRef) extends Instrum
         // Either way an enhancement would be required here to make Stream[_] responses work
         // with context-per-jvm=true configuration
         resultActor ! JobResult(jobId, result)
+        statusActor ! JobFinished(jobId, DateTime.now())
```
velvia: I'm fine with this, but just to note that the JobResultActor should really be local to the driver JVM, which would be a better fix.

velvia:
Doesn't this fix some open issues?  Will wait a day for anybody else to comment.

velvia:
Merging.

marcoboi:
@velvia 
Am I wrong saying that the changes I see above do not provide a definite solution to the problem?
Is the issue still open?
Would it not help to simply sending the Result with an "ask" and send the JobFinished through the ask's Future?

PR #652, opened 2016-11-14
Add -p to mkdir to create full path

mcnels1:
Running ec2_deploy.sh fails with:

mkdir: /Users/mnelson/git/spark-jobserver/bin/../ec2Cluster/deploy.generic/root: No such file or directory

because mkdir is trying to create multiple directory levels without the -p option.


mcnels1:
There are other problems with ec2_deploy.sh.  I'm going to close this and submit an new request fixing those.

velvia:
Thanks Mark.

PR #653, opened 2016-11-14
Make ec2_deploy.sh work with current spark-ec2

mcnels1:
- Fix a broken mkdir by adding -p option
- Get spark-ec2 from its new location as a separate project
- Add additional configuration

### DIFF
--- a/bin/ec2_deploy.sh
+++ b/bin/ec2_deploy.sh
@@ -8,7 +8,7 @@ bin=`cd "$bin"; pwd`
 SPARK_DIR=ec2Cluster
 if [ ! -d "$bin"/../$SPARK_DIR ]; then
     mkdir "$bin"/../$SPARK_DIR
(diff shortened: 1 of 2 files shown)


mcnels1:
I'm not familiar with with your CI process, but I don't see how what I changed (ec2_deploy.sh and user-ec2-settings.sh.template) could have caused your tests to fail.

velvia:
@mcnels1 thanks, this looks fine to me.   I'll restart Travis.

mcnels1:
Any suggestions on what to do about the failing CI checks?

velvia:
Hmm, I would ignore it since your change clearly doesn’t change code.   Thought I restarted it, but could be wrong

mcnels1:
I have another pull request ready to submit that builds on this one to support deployment into an EC2 VPC.  I can submit it now or wait until after this one gets merged.  Preferences?

noorul:
@mcnels1 Thank you!

PR #657, opened 2016-11-17
Handle deployment into a VPC

mcnels1:
Add additional configuration and logic so that the cluster can be deployed into a VPC.  Adds a dependency on jq, but only if deploying to a VPC.  aws ec2 authorize-security-group-ingress identifies the security group by ID, not name, if the group is in a VPC.  I could not find any better way to look up a security group ID than to list them all with aws ec2 describe-security-groups and filter the resulting JSON with jq.


velvia:
@mcnels1 thanks, this looks good to me.

PR #661, opened 2016-11-22
Change to increase job config input size

rmettu-rms:
Passing in 155K job config input string resulted in Akka timeout response to the job.
In WebApi, when creating JobManagerActor it fails to serialize jobConfig object when greater
than 64K (not verified). Change is to serialize the config object as a string and recreate
it when processing the job.

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -3,9 +3,11 @@ package spark.jobserver
 import java.net.{URI, URL}
 import java.util.concurrent.Executors._
 import java.util.concurrent.atomic.AtomicInteger
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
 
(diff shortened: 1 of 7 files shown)


ram-copart:
@velvia just a friendly reminder to review this PR

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
```diff
@@ -41,6 +45,56 @@ object JobManagerActor {
   // Akka 2.2.x style actor props for actor creation
   def props(contextConfig: Config, daoActor: ActorRef): Props = Props(classOf[JobManagerActor],
     contextConfig, daoActor)
+
+  // Serializer StartJob message
+  class StartJobSerializer extends JSerializer {
```
velvia: I would move this class outside, perhaps to `spark.jobserver.util`

rmettu-rms: and add serialization test

rmettu-rms:
@velvia made suggested changes, please review

rmettu-rms:
@velvia friendly reminder to review the changes

PR #663, opened 2016-11-28
Spark driver cluster mode for mesos and upgrade to spark 2.0

LeightonWong:
Related to #608 , we have run this in our production environment for about two months. So We want to submit this new pr to the master branch to support driver in cluster mode because we need many drivers run in the same time, the client mode is not suitable for the limit of one machine.

Also we upgrade to spark 2.0.1 and test this for about one month , it works well, we did not change too much, only change the dependencies for spark 2.0, because `SparkSession` has good compatibility for the former `SparkContext`, `SparkContext` can be used as before, but if we want to use new feature of `SparkSession`, we have to change the context part to be new `SparkSession`, so this is only a temporary solution for spark 2.0.


noorul:
@LeightonWong We will not be able to merge spark 2.0 changes to master branch as of now

LeightonWong:
@noorul OK

velvia:
I would suggest this - break up the PR into two, the portion without the Spark 2.0 and the portion with it.  We can merge the portion without 2.0. 

-Evan

addisonj:
+1 to this change, I am running in mesos but its quite a bit hackier than this by copying jars back and forth out of S3.

This would make that way cleaner :)

LeightonWong:
@velvia @addisonj OK, I will try to make this into two pr, only commit 9aaa884 include spark 2.0 changes.

LeightonWong:
@velvia @addisonj @noorul , I have submit a new pr #681 without spark 2.0， and I will close this now.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -199,16 +199,16 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
...
-          return
-      }
+    val driverMode = Option(config.getString("spark.jobserver.driver-mode")).getOrElse("client")
+    val (workDir, contextContent) = generateContext(name, contextConfig, isAdHoc, contextActorName)
+    logger.info("Ready to create working directory {} for context {}", workDir: Any, name)
 
     //extract spark.proxy.user from contextConfig, if available and pass it to $managerStartCommand
-    var cmdString = s"$managerStartCommand $contextDir ${selfAddress.toString}"
+    var cmdString = if (driverMode == "mesos-cluster") {
+      s"$managerStartCommand $driverMode $workDir '$contextContent' ${selfAddress.toString}"
```
bsikander: @LeightonWong Could you please explain a little that why we need single quotes around $conextContent here but not in line 210 ?

LeightonWong: In mesos cluster mode, jobserver manager will start in remote mode with MesosClusterDispatcher differ from local mode, they are using different submit client, I found it will cause error when parsing the contextContent which include double quotes it self, so I just add the single quotes and it works, I don't know the exact reason, may be you have to read spark MesosClusterDispatcher/MesosRestServer/RestSubmitionClient code in spark ,it takes the responsibility to start the jobserver manager int cluster mode.

PR #667, opened 2016-12-01
Eliminate POST and DEL /job race conditions (for spark-2.0)

addisonj:
For both these routes, there was a race condition when you immediately
try and check the status of a job after starting or stopping it.

For example, if you POST'ed an async job and then immediately try and
do a GET on that jobId, the jobInfoActor may not have persisted the job
yet, so you would get a 404. Similiar could happen for stopping a job,
if you immediately tried to GET after cancelling, it may still show as
  RUNNING

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -159,7 +159,9 @@ class JobManagerActor(contextConfig: Config, daoActor: ActorRef) extends Instrum
 
     case KillJob(jobId: String) => {
       jobContext.sparkContext.cancelJobGroup(jobId)
-      statusActor ! JobKilled(jobId, DateTime.now())
+      val resp = JobKilled(jobId, DateTime.now())
+      statusActor ! resp
+      sender ! resp
     }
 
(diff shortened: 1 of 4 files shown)


addisonj:
This is failing due to an already broken spec in spark-2.0-preview, a fix for that is outstanding in #671, once that is merge, a retrigger here should fix the build!

velvia:
Ok try merging the latest spark-2.0-preview with your other fix

velvia:
@addisonj do you have time to merge with latest 2.0-preview?

addisonj:
okay, just rebased this on top of latest spark-2.0-preview, this looks to be a correct changeset, but not sure why tests aren't retriggering

velvia:
@addisonj I restarted the build.  For some reason it is trying to download Scala 2.10.

noorul:
@addisonj  Can you re-base this against latest?

addisonj:
This looks like it was merged already. https://github.com/spark-jobserver/spark-jobserver/commit/e83897252c317c546cf2f974c5d4581159a92a0f I think this was just getting it into the spark-2.0 preview branch? Either way, this is in master and it looks like spark-2.0-preview is now merged

PR #668, opened 2016-12-01
Eliminate POST and DEL /job race conditions

addisonj:
For both these routes, there was a race condition when you immediately
try and check the status of a job after starting or stopping it.

For example, if you POST'ed an async job and then immediately try and
do a GET on that jobId, the jobInfoActor may not have persisted the job
yet, so you would get a 404. Similiar could happen for stopping a job,
if you immediately tried to GET after cancelling, it may still show as
  RUNNING

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -159,7 +159,9 @@ class JobManagerActor(contextConfig: Config, daoActor: ActorRef) extends Instrum
 
     case KillJob(jobId: String) => {
       jobContext.sparkContext.cancelJobGroup(jobId)
-      statusActor ! JobKilled(jobId, DateTime.now())
+      val resp = JobKilled(jobId, DateTime.now())
+      statusActor ! resp
+      sender ! resp
     }
 
     case SparkContextStatus => {
--- a/job-server/src/main/scala/spark/jobserver/WebApi.scala
+++ b/job-server/src/main/scala/spark/jobserver/WebApi.scala
@@ -514,8 +514,12 @@ class WebApi(system: ActorSystem,
                 notFound(ctx, "No such job ID " + jobId)
               case JobInfo(_, contextName, _, classPath, _, None, _) =>
                 val jobManager = getJobManagerForContext(Some(contextName), config, classPath)
-                jobManager.get ! KillJob(jobId)
-                ctx.complete(Map(StatusKey -> JobStatus.Killed))
+                val future = jobManager.get ? KillJob(jobId)
+                future.map {
+                  case JobKilled(_, _) => ctx.complete(Map(StatusKey -> JobStatus.Killed))
+                }.recover {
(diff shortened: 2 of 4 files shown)


barrybecker4:
Both @kapil-malik and myself are still seeing the 404 errors occasionally after incorporating this patch.

hntd187:
@barrybecker4 can you open an issue and provide more details if any please?

barrybecker4:
See issue #682

barrybecker4:
If I add just a 10 millisecond delay before requesting the result after submitting a job request, then it seems that I will not get the 404 error (or at least much less often). It only happens when you make the request for the result immediately after submitting the job.

barrybecker4:
Another related problem I have observed using JobSqlDAO (with MySQL db) and not with JobFileDAO, is that sometimes a job with status "FINISHED" might briefly have a status of "RUNNING" again before returning to the "FINISHED" status. I know this because I have cucumber tests, that wait until the status is not "RUNNING" before getting the result. Then when I get the result, the status is sometimes "RUNNING" and not "FINISHED".

kapil-malik:
I've also observed related issue on JobSqlDAO + MySQL - 
Similar to the one mentioned above, job shows status as FINISHED too early, and there's no "result" field in the response. Interestingly, in my observations, it never reverts back to RUNNING, but keeps it as FINISHED, and when it's actually done, the "result" field is present and populated correctly.

I'd a polling mechanism to check on job status before returning the result to user (am using SJS with custom Zeppelin interpreter), with this behaviour, the protocol goes for a toss.

@barrybecker4 when you see FINISHED status for the first time, does it have result field too? / is the job actually finished?

barrybecker4:
Not sure, but here is another case I have observed that seems to indicate some sort of race condition.
Sometimes when I DELETE a job it will still have a status of RUNNING instead of KILLLED as it should. Most of the time when I get the status right after the cancel, the status will be KILLED, but occasionally it is RUNNING. The response from the DELETE request always returns {"status":"KILLED"}.

barrybecker4:
I would really like to create some additional tests cases that can prove that this is working (or not).
Unfortunately, I'm not sure how to create them or where to put them, and my knowledge of akka is limited.
Should they go in JobManagerSpec? Is seems that the tests mostly use InMemoryDAO. Does that mean these cases will be difficult to test?
The cases I would like to add are:
  - submit a job, then immediately verify that the status is RUNNING (and not 404 as we have seen sometimes)
  - submit a job, verify that it is running, then kill it and get the status immediately to verify that it is KILLED (and not RUNNING as we have seen sometimes).
  - submit a job, wait until it has a status which is FINISHED, then check the status several more times to verify that it stays FINISHED and does not switch back to running.
  
Can anyone give me an example of how I might try to add tests like these? Do any exist already?

velvia:
Hmm….  JobManagerSpec would be a good place to start.  However it really depends where the bug manifests. Perhaps the problem is not in JobManager but rather in the DAO in which case the StatusActor and DAO needs to be involved.

My gut instinct is that the SupervisorSpecs might be better.

PR #671, opened 2016-12-05
Fix NamedObjectsJobSepc under docker

addisonj:
Seems docker is just a bit slower, and this specific spec just needed a
longer timeout


addisonj:
Thanks!!

PR #672, opened 2016-12-05
Refactor contextSupervisors, fix atomic context

### LINKED ISSUE #349: Adding new context (ad-hoc or named) is not atomic
Right now, in `LocalContextSupervisorActor`, and in `AkkaClusterSupervisorActor` in the soon-to-be-merged `feature/context-per-jvm` branch, there are two ways to create a new SparkContext:  `AddContext`, which adds a named context and `StartAdHocContext`.  The problem is that the code in both of these checks to see if a context with the requested name is already running, or in the ad-hoc's case, it randomly generates a name and checks if it is already taken.  However, the hashmap `contexts` is not populated until the `JobManagerActor` successfully creates a new SparkContext, which is anywhere from 5-15 seconds at least.  In the intervening 5-15 seconds, someone could re-request a new context with the same name, and job server would stupidly try to create another one, since the original one was not populated in `contexts` hashmap.

This has not really been a problem because multiple contexts were not officially supported, though they kinda work with a Spark hack.

What is now needed is the following:
1) A way to track contexts that are being created via some state;
2) Do not create contexts if they are already being created, and return an appropriate error message
3) For Ad-Hoc contexts, generate a different random name if context with previously generated name is being started
4) If the contexts being started never respond, or other error situation, remove them from the "being created" list after some timeout so that creation could be re-tried.

This needs to be done for both ad-hoc and named contexts, and both Local and AkkaCluster supervisor actors.

addisonj:
This commit refactors the Local and AkkaCluster context supervisor actor
classes. It introduces a new abstract class, BaseSupervisorActor that
both LocalContextSupervisorActor and AkkaClusterSupervisorActor extend.
This new base class has some improved functionality:
- atomic context creation, you can't double create a context during it
  starting (fixes #349)
- contexts that are starting are listed in the contexts (we should
  probably give contexts a status)
- if you try and stop a context that is still starting, it will ensure it
  properly gets stopped
- if a context fails to start, it will get removed from list of starting
  contexts
- create more specific exceptions for when a context isn't found but is
  in creating mode, allows us to send a better API error indiciating the
  problem

The base class does have to take on some of the concerns of the
AkkaCluster for simplicity sake (just handling a few messages), but this
really DRYs up the two different classes

AkkaClusterSupervisorActor also had no unit tests, it now is fairly well tested save
for the `startContext` method which actually spawns the new process

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/672)

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -4,18 +4,15 @@ import java.io.IOException
 import java.nio.file.{Files, Paths}
 import java.nio.charset.Charset
 import java.util.concurrent.TimeUnit
+import scala.collection.mutable
+import scala.util.Try
 
 import akka.actor._
-import akka.cluster.Cluster
-import akka.cluster.ClusterEvent.{InitialStateAsEvents, MemberEvent, MemberUp}
-import akka.util.Timeout
-import com.typesafe.config.{Config, ConfigFactory, ConfigRenderOptions}
+import akka.cluster.{Cluster, Member}
+import akka.cluster.ClusterEvent.{InitialStateAsEvents, MemberEvent}
+import com.typesafe.config.{Config, ConfigRenderOptions}
 import spark.jobserver.util.SparkJobUtils
-import scala.collection.mutable
-import scala.util.{Failure, Success, Try}
 import scala.sys.process._
-
-import spark.jobserver.common.akka.InstrumentedActor
 
 /**
  * The AkkaClusterSupervisorActor launches Spark Contexts as external processes
@@ -35,13 +32,10 @@ import spark.jobserver.common.akka.InstrumentedActor
  *   }
  * }}}
  */
-class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
-  import ContextSupervisor._
+class AkkaClusterSupervisorActor(daoActor: ActorRef) extends BaseSupervisorActor {
   import scala.collection.JavaConverters._
   import scala.concurrent.duration._
 
-  val config = context.system.settings.config
-  val defaultContextConfig = config.getConfig("spark.context-settings")
   val contextInitTimeout = config.getDuration("spark.context-settings.context-init-timeout",
                                                 TimeUnit.SECONDS)
   val managerStartCommand = config.getString("deploy.manager-start-cmd")
@@ -51,17 +45,11 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
   //TODO: try to pass this state to the jobManager at start instead of having to track
   //extra state.  What happens if the WebApi process dies before the forked process
   //starts up?  Then it never gets initialized, and this state disappears.
-  private val contextInitInfos = mutable.HashMap.empty[String, (Boolean, ActorRef => Unit, Throwable => Unit)]
-
-  // actor name -> (JobManagerActor ref, ResultActor ref)
-  private val contexts = mutable.HashMap.empty[String, (ActorRef, ActorRef)]
+  protected val contextInitInfos =
+    mutable.HashMap.empty[String, (Boolean, ActorRef => Unit, Throwable => Unit)]
 
   private val cluster = Cluster(context.system)
   private val selfAddress = cluster.selfAddress
-
-  // This is for capturing results for ad-hoc jobs. Otherwise when ad-hoc job dies, resultActor also dies,
-  // and there is no way to retrieve results.
-  val globalResultActor = context.actorOf(Props[JobResultActor], "global-result-actor")
 
   logger.info("AkkaClusterSupervisor initialized on {}", selfAddress)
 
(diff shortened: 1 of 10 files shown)


job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -215,10 +113,12 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
     }
 
     val pb = Process(cmdString)
-    val pio = new ProcessIO(_ => (),
-                        stdout => scala.io.Source.fromInputStream(stdout)
-                          .getLines.foreach(println),
-                        stderr => scala.io.Source.fromInputStream(stderr).getLines().foreach(println))
+    val pio = new ProcessIO(
+      (_) => (),
```
zeitos: why are you refactoring these `x =>` to `(x) =>`   we don't have that convention . I would rather do that change in the whole project as a separate PR.

It makes code reviewing more complicated :(

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
```diff
@@ -0,0 +1,260 @@
...
+    case ListContexts =>
+      sender ! listContexts()
+
+    case AddContext(name, contextConfig) =>
+      val originator = sender()
+      val mergedConfig = contextConfig.withFallback(defaultContextConfig)
+      if (haveContext(name)) {
+        originator ! ContextAlreadyExists
+      } else {
+        registerAndStartContext(name, mergedConfig, false)(
```
zeitos: this is not atomic... an AddContext might be calling `registerAndStartContext` while another call might be executing `haveContext` and register again.

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
```diff
@@ -0,0 +1,260 @@
...
+
+    case GetContext(name) =>
+      if (haveContext(name)) {
+        val actorOpt = getContext(name)._2
+        if (actorOpt.isDefined) {
+          val future = (actorOpt.get ? SparkContextStatus)(getTimeout().seconds)
+          val originator = sender
+          future.collect {
+            case SparkContextAlive => originator ! getContext(name)
+            case SparkContextDead =>
```
zeitos: note that if someone calles getContext while it's initiating before the internal sparkcontext is created, it will try to stop it (it might even stop a recently created context).

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
addisonj: yeah... I debated that one. I looked around and saw examples of both in a few places, so wasn't sure it was super standard across the project, and moving the first anonymous function onto its own line, it seemed to read a lot better to me. I can keep the spacing and remove the `()` parens, but I do think long term parens around the single arg functions reads quite a bit better (I mostly hold that from writing a lot ES6 where that is becoming the norm), but its definitely not too important

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
addisonj: Good catch, will probably add some logic to the JobManagerActor to send back a new message for `SparkContextStarting` that we can explicitly handle here

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
velvia: I could be wrong but the standard I've seen in lots of Scala code is:

    (_.blah(...))   // Use parentheses for simple expressions without => or when using underscore

    { v => v.doSomething(...) }    // Braces for formal closures with =>

I'm not really feeling religious about it, maybe it's just what I'm used to....

addisonj: Yeah, I don't care too much either, like I said, mostly I hold that opinion from ES6 javascript where parens around single arg functions is becoming the standard. So if we want to do it one way, thats fine with me, but would be nice if scalastyle handled it :)

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
addisonj: lol yeah... been writing node.js for the last few months so my ability to reason about real threads has obviously diminished. I will move this up earlier in the call and wrap access to the underlying contexts object in synchronized

zeitos: You can also take a look at http://doc.akka.io/docs/akka/2.0/scala/transactors.html

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -229,49 +129,43 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
...
       failureFunc(processStart.failed.get)
     }
-
   }
 
-  private def createContextDir(name: String,
-                               contextConfig: Config,
-                               isAdHoc: Boolean,
-                               actorName: String): java.io.File = {
+  private def createContextDir(
```
velvia: Personally I prefer the original style, where the first param is on the same line as the function name.

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
```diff
@@ -0,0 +1,260 @@
...
+
+  val config = context.system.settings.config
+  val defaultContextConfig = config.getConfig("spark.context-settings")
+
+  import context.dispatcher   // to get ExecutionContext for futures
+  import akka.pattern.ask
+
+  // This is for capturing results for ad-hoc jobs. Otherwise when ad-hoc job dies, resultActor also dies,
+  // and there is no way to retrieve results.
+  val globalResultActor = context.actorOf(Props[JobResultActor], "global-result-actor")
```
velvia: We could probably get rid of the global result actor now that most people will fork off separate JVMs for each context.  Instead each context should have its own result actor.  However that can be left for a diff PR.

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
velvia: +1

job-server/src/main/scala/spark/jobserver/BaseSupervisorActor.scala
```diff
@@ -0,0 +1,260 @@
...
+  protected def addContext(name: String, context: SJSContext) = {
+    contexts(name) = context
+  }
+
+  protected def onMemberUp(member: Member): Unit
+  protected def onActorIdentity(memberActors: Any, actorRefOpt: Option[ActorRef]): Unit
+  protected def onStopContext(actor: ActorRef): Unit
+  protected def getTimeout(): Long
+
+  protected def startContext(
```
velvia: Again same here, prefer params starting on same line as function name, like

    protected def startContext(name: String,
                                                   actorName: String,  
     ...)

(sorry what I mean is name and actorName line up vertically)

velvia:
@addisonj I think @zeitos covered the review pretty well, this will be an awesome change and I like the general direction and cleanup, thanks!

noorul:
@addisonj Nice work!

velvia:
@addisonj let us know if you can address requested changes.  We are looking to release 0.7 soon and hope to get this one in.

noorul:
@addisonj  Can you help us get this merged quickly ?

zeitos:
hey guys... should we close this one since the code change a lot... I think its @addisonj 's call but I feel like it will need to be reworked.
Or maybe remove the atomic part and merge the rest of the features...

velvia:
Hmmm.  We still need the changes.  Lemme have a quick look and see.  

-Evan
"We can do no great things, just small things with great love." -mother Teresa

addisonj:
Sorry I haven't been able to look at this, been quite busy. I *might* get
some time this weekend but this has been pushed pretty low with some
deadlines.

noorul:
@addisonj Do you think you will get sometime?

zeitos:
@noorul I think we should close this and reopen if needed

velvia:
Everybody, @addisonj, thanks for the PR but I'm going to close it, this has been inactive for a very long time, the code base has changed significantly and there are other PRs concerning the supervision and context forking that I think should address this issue.

PR #681, opened 2016-12-14
Spark driver cluster mode for mesos

LeightonWong:
Related to #663 ，I break it into two pr. Without spark 2.0 upgrade, this pr only contain mesos cluster mode changes.
More detail usage, please refer to #608 .


job-server/src/main/scala/spark/jobserver/BinaryManager.scala
```diff
@@ -40,7 +40,7 @@ class BinaryManager(jobDao: ActorRef) extends InstrumentedActor {
   import scala.concurrent.duration._
   import akka.pattern.{ask, pipe}
   import context.dispatcher
-  implicit val daoAskTimeout = Timeout(3 seconds)
+  implicit val daoAskTimeout = Timeout(60 seconds)
```
hntd187: Why this change?

job-server/src/main/scala/spark/jobserver/JobCache.scala
```diff
@@ -28,7 +31,7 @@ class JobCacheImpl(maxEntries: Int,
 
   private val cache = new LRUCache[(String, DateTime, String, BinaryType), BinaryJobInfo](maxEntries)
   private val logger = LoggerFactory.getLogger(getClass)
-  implicit val daoAskTimeout: Timeout = Timeout(3 seconds)
+  implicit val daoAskTimeout: Timeout = Timeout(60 seconds)
```
hntd187: again

job-server/src/main/scala/spark/jobserver/BinaryManager.scala
LeightonWong: When we test mesos cluster mode, we often got Timeout error, because the jar we submit is a little big(about 30M) and our database is not fast enough,  and there are many synchronize call in spark-jobserver like `jobDao ? JobDAOActor.SaveBinary(appName, binaryType, uploadTime, binBytes)).` and `val resp = (jobDao ? JobDAOActor.GetApps(filterOpt)).mapTo[JobDAOActor.Apps]
      resp.map { msg => msg.apps } pipeTo requestor`, I simply enlarge the daoAskTimeout, but it is better to provide a config for this timeout.

job-server/src/main/scala/spark/jobserver/JobCache.scala
LeightonWong: As above, this `val jarBinaryReq = (dao ? JobDAOActor.GetBinaryContent(appName, BinaryType.Jar, uploadTime))
          .mapTo[JobDAOActor.BinaryContent]` in JobCache often cause timeout error, 3 seconds is not enough.

job-server/src/main/scala/spark/jobserver/BinaryManager.scala
noorul: I think we should make all timeouts configurable

noorul:
Some unit tests are missing that is why the score got decreased.

LeightonWong:
@noorul OK, I will try to add some unit tests to cover these diff.

velvia:
@LeightonWong this LGTM as well.

@noorul @zeitos @hntd187 are we OK with the drop in code coverage?  Let's try to get this in before publishing 0.7

noorul:
@LeightonWong Have you tested using dependent-jar-uris?

LeightonWong:
@noorul Sorry, I did not test dependent-jar-uris, In cluster mode, JobManager started at some remote node other than the job-server running machine, I think if dependent-jar-uri use remote protocal like http://xxx.jar it will work, but if using file://xxx.jar, we have to put all the dependent jars in this machine local file system and it might be work also.

In our use case, we put all spark job third party dependent jars such like `alluxio-1.4.0-spark-client-jar-with-dependencies.jar` in spark `libs` directory with spark 1.6.x version and `jars` directory with spark 2.x version, so we are not using job-server's `dependent-jar-uris` config.

velvia:
@LeightonWong sorry do you mind merging changes with master?  We can then submit this to master and release.   Thanks.

LeightonWong:
@velvia Ok, no problem, I will merge master and resolve conflicts right now.

velvia:
Thanks!

velvia:
@LeightonWong thanks.  Can't see diff for some reason but Travis passes, so will merge now.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -208,16 +208,16 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
...
     logger.info("Starting context with actor name {}", contextActorName)
 
-    val contextDir: java.io.File = try {
-        createContextDir(name, contextConfig, isAdHoc, contextActorName)
-      } catch {
-        case e: Exception =>
-          failureFunc(e)
-          return
-      }
+    val driverMode = Option(config.getString("spark.jobserver.driver-mode")).getOrElse("client")
```
noorul: @LeightonWong This line breaks the existing functionality on master. I think there should be a default value for this in application.conf or you should handle the case that there might not be one.

LeightonWong: Sorry, It's better to add a default value for driver-mode in application.conf, but will it break the functionality? If the config is not set, the default "client" mode will be used, and nothing change as before.

noorul: @LeightonWong I agree! Let us add a default in application.conf

PR #708, opened 2017-01-23
Cleanup LDAP authentication and add filter config.

derSascha:
This PR adds more filter configurations e.g. if you don't use the member attribute in your LDAP (see shiro template for details).


derSascha:
Flaky timeout test failes (JobWithNamedRddsSpec.scala:74), everything else works.

PR #710, opened 2017-01-23
Added process std error output to exception message exposed in python job API

windelinckx:


noorul:
Thank you @windelinckx

Issue #395, opened 2016-03-08
Make sure -X DELETE for contexts is synchronous


CBribiescas:
According to the documentation, sync option is only for jobs

> Asynchronous and synchronous job API. Synchronous API is great for low latency jobs!

It would be nice to have a sync option for context creation.  Quickly (or programmatically) running the following repeatedly results in out of sync issues.

``````
curl -d "" -X POST -H "application/x-www-form-urlencoded" "http://localhost:8090/contexts/context?num-cpu-cores=4&memory-per-node=512m&sync=true"
curl -d "" -X DELETE -H "application/x-www-form-urlencoded" "http://localhost:8090/contexts/context?sync=true"````
``````

**Comment by velvia** [comment]
Hi @CBribiescas 

Context creation is always synchronous -- when the route comes back, that means the context is created.  DELETE might not be synchronous though - which we can look into.

**Comment by CBribiescas** [comment]
Yes, I tested them separately and it does look like only the delete is not synchronous.  Create operated correctly.

**Comment by noorul** [comment]
@CBribiescas, @velvia  What exactly is the issue here. When I create and delete context, I don't see any timing issue.

**Comment by bsikander** [comment]
I am also facing this issue. Since, DELETE is async, when you try to create a new context right after deleting it, you get a 400 response back. The temporary solution which I found in the source code of jobserver is to do a Thread.sleep. See line [131 of this file](https://github.com/spark-jobserver/spark-jobserver/blob/0919588387d85460460401556614230148ceeae5/job-server/src/test/scala/spark/jobserver/LocalContextSupervisorSpec.scala).

**Comment by emlagowski** [comment]
I also have to kill context, create new one and then run a job.

My queries and responses:
```
DELETE ctxName -> OK
POST create ctxName -> 'context ctxName exists'
POST run job on ctxName -> 'context ctxName not found'
```

The possibility of synchronous DELETE could be useful.

doernemt:
@velvia issue can be closed, my PR to fix it was merged a month ago and is part of the 0.7 release. 
Thanks, Tim

velvia:
@doernemt thanks

PR #727, opened 2017-02-03
fix(webapi,akka-app): make sure -X DELETE for contexts is synchronous…

doernemt:
… #395

* DeleteContext operation is now synchronous, it doesn't return before
the corresponding
actor has been gracefully stopped. There is no need anymore to wait a
few 100 ms
before re-creating a context with the same name

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -8,16 +8,16 @@ import java.util.concurrent.TimeUnit
 import akka.actor._
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent.{InitialStateAsEvents, MemberEvent, MemberUp}
-import akka.pattern.gracefulStop
 import akka.util.Timeout
 import com.typesafe.config.{Config, ConfigFactory, ConfigRenderOptions}
-import spark.jobserver.common.akka.InstrumentedActor
 import spark.jobserver.util.SparkJobUtils
-
 import scala.collection.mutable
-import scala.concurrent.{Await, Future}
-import scala.sys.process._
 import scala.util.{Failure, Success, Try}
+import scala.sys.process._
+
+import spark.jobserver.common.akka.InstrumentedActor
+import scala.concurrent.Await
+import akka.pattern.gracefulStop
 
 /**
  * The AkkaClusterSupervisorActor launches Spark Contexts as external processes
@@ -39,7 +39,6 @@ import scala.util.{Failure, Success, Try}
  */
 class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
   import ContextSupervisor._
-
   import scala.collection.JavaConverters._
   import scala.concurrent.duration._
 
@@ -161,7 +160,7 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
           sender ! ContextStopped
         }
         catch {
-          case err :Exception => sender ! ContextStopError(err)
+          case err: Exception => sender ! ContextStopError(err)
         }
       } else {
         sender ! NoSuchContext
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -1,21 +1,23 @@
 package spark.jobserver
 
 import java.io.IOException
-import java.nio.file.{Files, Paths}
 import java.nio.charset.Charset
+import java.nio.file.{Files, Paths}
 import java.util.concurrent.TimeUnit
 
 import akka.actor._
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent.{InitialStateAsEvents, MemberEvent, MemberUp}
+import akka.pattern.gracefulStop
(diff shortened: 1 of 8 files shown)


job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -152,8 +155,14 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
...
         cluster.down(contextActorRef.path.address)
-        contextActorRef ! PoisonPill
-        sender ! ContextStopped
+        try {
+          val stoppedCtx = gracefulStop(contexts(name)._1, 2 seconds)
+          Await.result(stoppedCtx, 3 seconds)
+          sender ! ContextStopped
+        }
+        catch {
+          case err :Exception => sender ! ContextStopError(err)
```
hntd187: Colon next to variable not type plz

job-server/src/main/scala/spark/jobserver/LocalContextSupervisorActor.scala
```diff
@@ -144,8 +143,14 @@ class LocalContextSupervisorActor(dao: ActorRef) extends InstrumentedActor {
...
         logger.info("Shutting down context {}", name)
-        contexts(name)._1 ! PoisonPill
-        sender ! ContextStopped
+        try {
+          val stoppedCtx = gracefulStop(contexts(name)._1, 2 seconds)
+          Await.result(stoppedCtx, 3 seconds)
+          sender ! ContextStopped
+        }
+        catch {
+          case err :Exception => sender ! ContextStopError(err)
```
hntd187: Colon next to variable

doernemt:
@hntd187, not sure if I really understood your comment. Could you clarify your suggestion?

What I did is to change the (asynchronous) PoisonPill that lacks any feedback about the shutdown of the actor by gracefulShutdown, which -- to my understanding -- has the same semantics (process all messages in the queue and then shutdown), but provides feedback about the completion ( + timeout behavior).

The closest we get to a synchronous operation would be to call `context.stop(contexts(name)._1)`, but that would have a different semantic: it would not process all messages in the mailbox, but only the currently running one and then stop.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
doernemt: Thx

noorul:
No matter how many times I re-try travis job, one of them fails. Not sure why it is flaky.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -152,8 +155,14 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
         logger.info("Shutting down context {}", name)
         val contextActorRef = contexts(name)._1
         cluster.down(contextActorRef.path.address)
-        contextActorRef ! PoisonPill
-        sender ! ContextStopped
+        try {
+          val stoppedCtx = gracefulStop(contexts(name)._1, 2 seconds)
```
doernemt: Good point. Should I add a new config option `jobserver.context-deletion-timeout` ?

doernemt:
@noorul: Thanks for your feedback. I found out why travis fails sometimes: gracefulStop is a (at least) two-step procedure internally: (1) it stops the context and gives feedback about this, (2) it de-registers the name with the actor system. (2) is done asynchronously, so in very rare cases the call of gracefulStop already returns, but the name is then still taken (http://doc.akka.io/docs/akka/2.4/scala/actors.html). This causes the test to fail as it immediately re-lists the context after StopContext returns. 
In real-life applications, this should not be a problem, but needs to be fixed anyway.

I can reproduce this locally now -- every 20th test execution or so locally fails.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
noorul: I think so.

doernemt:
Oh, boy ;/ This time, Travis failed to download a Jar in one of the four environments. All others are green. Reopening to trigger rebuild

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
doernemt: Done with the latest commit. Parameters are:
spark.jobserver.yarn-context-deletion-timeout and spark.jobserver.context-deletion-timeout

doernemt:
@noorul, @hntd187: Any other comments regarding required changes? Tests are stable now, and timeout is configurable.

I hope to get this merged into 0.7, would allow my team to remove a workaround in our code. Others seem to have the same issue,  see #395.

Thanks, Tim

doernemt:
Retriggering Travis. 
Delete Job test randomly (?) failed. Not related at all to my changes

noorul:
@doernemt LGTM, will wait for @hntd187

doernemt:
Added another PR (747) that fixes the test issue with deleteBinary -- not related to this PR, but it should make the tests pass

doernemt:
Rebased on master

doernemt:
Re-Trigger Travis

noorul:
Thank you! Merged!

PR #731, opened 2017-02-08
Pin docker API version

noorul:


noorul:
There is pep8 issue now. Not sure what is going on. Merging this.

PR #733, opened 2017-02-08
Pass codecov.io env variables to docker

noorul:
Now the tests are running inside the docker. But codecov is unable to detect CI system. 
So pass codecov environment variables to docker run.


velvia:
Thanks @noorul

PR #751, opened 2017-02-27
Submitting job on windows cmd

mkbond777:
When submitting a job (as below) on windows command prompt use with " (double quotes) instead '(single quote) with curl command near localhost.

Example:
For:
curl -d "input.string = a b c a b see" 'localhost:8090/jobs?appName=test&classPath=spark.jobserver.WordCountExample'
Use:
curl -d "input.string = a b c a b see" "localhost:8090/jobs?appName=test&classPath=spark.jobserver.WordCountExample"

### DIFF
--- a/doc/troubleshooting.md
+++ b/doc/troubleshooting.md
@@ -125,3 +125,12 @@ Before typing `reStart` in sbt, type `project job-server-extras` and only then s
 ```scala
 ConfigFactory.parseReader(paramReader = new InputStreamReader(getClass().getResourceAsStream(s"/$myPassedConfigPath"))
 ```
+## Submitting job on windows cmd
+
+When submitting a job (as below) on windows command prompt use with " (double quotes) instead '(single quote) with curl command near localhost.
+
+Example:
+For:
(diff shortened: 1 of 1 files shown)


hntd187:
How is this different from doing it on linux?

mkbond777:
Actually, in the readme of job server, single quotes were used around localhost. So I thought of correcting it. As whenever you copy paste the commands it will throw the error like localhost not found.

I haven't checked in Linux, so I am not sure about it.

hntd187:
Why not correct it to be double quotes in the current docs instead of adding an entirely new section?

mkbond777:
Yes, we can correct it into the current docs as well.

noorul:
@mkbond777 I think @hntd187 meant that we don't need two examples under windows section. Only the correct example is fine.

noorul:
@mkbond777 Do you want to take a look at @hntd187  comment?

noorul:
Fixed via 7a2cf1b88931419ea114c5d710cfa13588f96fc0 and closing this.

Issue #739, opened 2017-02-13
No logs for contexts in forked JVMs


basvandenbrink:
When setting `context-per-jvm = true`, logging does not got to STD_OUT, even in the Docker image. Instead, logging should go to `/tmp/job-server/<context_name><random_id>`. However, there is only a `context.conf` file in that folder while running that job, no logging file. I am running version 0.6.2. in Docker.

noorul:
I would create a separate docker image to override default one and would have something like this 

    RUN ln -sf /proc/1/fd/1 /var/log/spark-jobserver/spark-jobserver.log

and would override log4j-server.properties with 

    # Rotating log file configuration for server deploys
    
    # Root logger option
    
    log4j.rootLogger=${LOG_LEVEL},LOGFILE
    
    log4j.logger.org.spark-project.jetty=WARN
    log4j.logger.org.spark-project.jetty.util.component.AbstractLifeCycle=ERROR
    log4j.logger.org.apache.spark.repl.SparkIMain$exprTyper=WARN
    log4j.logger.org.apache.spark.repl.SparkILoop$SparkILoopInterpreter=WARN
    log4j.logger.org.apache.parquet=ERROR
    log4j.logger.parquet=ERROR
    
    log4j.logger.org.apache.hadoop.hive.metastore.RetryingHMSHandler=FATAL
    log4j.logger.org.apache.hadoop.hive.ql.exec.FunctionRegistry=ERROR
    
    log4j.appender.LOGFILE=org.apache.log4j.FileAppender
    log4j.appender.LOGFILE.File=/var/log/spark-jobserver/spark-jobserver.log
    log4j.appender.LOGFILE.layout=org.apache.log4j.PatternLayout
    log4j.appender.LOGFILE.layout.ConversionPattern=%d{yy/MM/dd HH:mm:ss} %p %c{1}: [%X{akkaSource}] - %m%n

basvandenbrink:
[...]

PR #767, opened 2017-03-17
Add config option to allow the script starting forked JVM contexts to…

fmcgough:
… keep its process in the foreground

While running in Docker with `context-per-jvm = true`, I was experiencing an issue whereby the spark-submit process to launch the new contexts would exit immediately when its parent process, the `manager-start.sh` script, exited. I tried all the usual suspects (nohup, disown, screen etc), but nothing worked except keeping the process in the foreground. (Running Docker version 17.03.0-ce, on ubuntu 14.04).

This will also address #739 as it means all the stdout from the forked contexts will go to the parent process's stdout, which is typically what you want in a Docker container.

### DIFF
--- a/job-server/config/docker.conf
+++ b/job-server/config/docker.conf
@@ -59,6 +59,7 @@ spark {
 
 deploy {
   manager-start-cmd = "app/manager_start.sh"
+  wait-for-manager-start = false
 }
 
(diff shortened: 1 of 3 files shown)


velvia:
Thanks is job-server/config/docker.conf

fmcgough:
Yep, good point. Will do.

PR #774, opened 2017-03-23
Upgrade scalatest and scalactic.

aa8y:
Upgrade them to the latest versions and fix compilation errors.


aa8y:
I can't figure out how to get the python tests to pass locally to fix whatever the problem is. I tried installing `pyspark` from the Spark binary but there is no `setup.py` for 1.6.2.

noorul:
LGTM, thank you @aa8y

noorul:
@amarouni @velvia I missed to check the new process introduced. Sorry about that.

PR #820, opened 2017-05-15
fix(akka): mark actor as down on init errors

derSascha:
**Current behavior :**
Start job server with `context-per-jvm: true` and create new context with some invalid configuration (e.g. `memory-per-node: invalid-value`. Context creation fails with some error as expected. The akka cluster becomes now unstable and don't accept new actors => its impossible to create a new context.

**New behavior :**
Mark actor as down on initialization errors and the cluster becomes stable again.
Same issue fixed in #624 at context delete.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ dist: trusty
 env:
   global:
     - _JAVA_OPTIONS="-Xmx1500m -XX:MaxPermSize=512m -Dakka.test.timefactor=3"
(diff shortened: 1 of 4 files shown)


derSascha:
Spark 1.6.2 was removed from mirror used in ci: http://apache.mirror.anlx.net/spark/

noorul:
Merged, thank you @derSascha

PR #830, opened 2017-05-23
fix(akka): Stop job manager actor on context stop

derSascha:
**Current behavior:**
1. Create a new context in YARN mode
2. Kill YARN app or call context.stop
3. Context is now in stopped mode and not usable anymore. Job Server lists the context as active and does not allow to start a new context with same name.

**New behavior:**
Listen to context events and stop job manager actor if context get stopped. This ensures that context gets also removed in job server.

PR contains another `cluster.down` call in case that an actor gets terminated.

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -173,6 +173,7 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
       val name: String = actorRef.path.name
       logger.info("Actor terminated: {}", name)
       contexts.retain { case (name, (jobMgr, resActor)) => jobMgr != actorRef }
+      cluster.down(actorRef.path.address)
   }
 
   private def initContext(actorName: String, ref: ActorRef, timeoutSecs: Long = 1)
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -8,6 +8,8 @@ import akka.actor.{ActorRef, PoisonPill, Props}
 import com.typesafe.config.Config
 import org.apache.hadoop.conf.Configuration
 import org.apache.spark.{SparkConf, SparkEnv}
+import org.apache.spark.scheduler.SparkListener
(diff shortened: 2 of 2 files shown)


velvia:
I'm going to restart the failed Travis CI build.
@zeitos @noorul @hntd187 any thoughts?

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
```diff
@@ -121,6 +124,12 @@ class JobManagerActor(contextConfig: Config, daoActor: ActorRef) extends Instrum
     Option(jobContext).foreach(_.stop())
   }
 
+  // Handle external kill events (e.g. killed via YARN)
+  override def onApplicationEnd(event: SparkListenerApplicationEnd) {
+    logger.info("Got Spark Application end event, stopping job manger.")
```
noorul: This will get triggered for normal application end also.

noorul: I asked a question related to this in the mailing list but have not got any response.

http://apache-spark-user-list.1001560.n3.nabble.com/Application-kill-from-UI-do-not-propagate-exception-td28539.html

velvia:
I think that’s fine right?  The only thing is that on normal application end we want to make sure the proper status is forwarded back to the context supervisor actor / DAO first.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -173,6 +173,7 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
       val name: String = actorRef.path.name
       logger.info("Actor terminated: {}", name)
       contexts.retain { case (name, (jobMgr, resActor)) => jobMgr != actorRef }
+      cluster.down(actorRef.path.address)
```
zeitos: are we taking down the whole cluster if one context is stopped?

velvia:
@fotero cluster.down() just changes the status of a single cluster node to DOWN.   Doesn’t bring down the whole cluster.

zeitos:
@velvia then this looks good to me!

noorul:
@derSascha Do you think we should take similar approach here as in https://github.com/spark-jobserver/spark-jobserver/pull/836 ?

Issue #824, opened 2017-05-17
DELETE /binaries giving scala.MatchError (SJS version 0.7.0)


bsikander:
I am using SJS 0.7.0 with H2 database

If I upload a jar file using 
`curl --data-binary @WordCount-1.0-SNAPSHOT.jar localhost:8090/binaries/WordCount_test --header "Content-Type: application/java-archive"`

and then delete it using 

`DELETE: curl -X DELETE localhost:8090/binaries/WordCount_test`
I get 
```
{
  "status": "ERROR",
  "result": {
    "message": "Success(()) (of class scala.util.Success)",
    "errorClass": "scala.MatchError",
[...]

PR #834, opened 2017-05-25
Fix DELETE binaries endpoint and response

nguyenuy:
**Pull Request checklist**

**Current behavior :** 
Described here spark-jobserver/spark-jobserver#824

*Two things:*
1. The binaries DELETE endpoint returns a 500 error on successful deletion
2. The binaries DELETE endpoint does not initiate actual deletion of the cached binaries

**New behavior :**
*Two things:*
1. The binaries DELETE endpoint returns a 200 error on successful deletion
2. The binaries DELETE endpoint will now delete the cachedBinaries

**Other information**:

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/BinaryManager.scala
+++ b/job-server/src/main/scala/spark/jobserver/BinaryManager.scala
@@ -33,6 +33,7 @@ case object InvalidBinary
 case object BinaryStored
 case object BinaryDeleted
 case class BinaryStorageFailure(ex: Throwable)
+case class BinaryDeletionFailure(ex: Throwable)
 
 /**
  * An Actor that manages the jars stored by the job server.   It's important that threads do not try to
@@ -106,6 +107,9 @@ class BinaryManager(jobDao: ActorRef) extends InstrumentedActor {
 
     case DeleteBinary(appName) =>
       logger.info(s"Deleting binary $appName")
-      deleteBinary(appName).pipeTo(sender)
+      deleteBinary(appName).map{
+        case Success(_) => BinaryDeleted
+        case Failure(ex) => BinaryDeletionFailure(ex)
+      }.pipeTo(sender)
   }
 }
--- a/job-server/src/main/scala/spark/jobserver/BinaryManager.scala
+++ b/job-server/src/main/scala/spark/jobserver/BinaryManager.scala
@@ -107,7 +107,7 @@ class BinaryManager(jobDao: ActorRef) extends InstrumentedActor {
 
     case DeleteBinary(appName) =>
       logger.info(s"Deleting binary $appName")
-      deleteBinary(appName).map{
+      deleteBinary(appName).map {
         case Success(_) => BinaryDeleted
         case Failure(ex) => BinaryDeletionFailure(ex)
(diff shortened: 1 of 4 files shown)


velvia:
Thanks for following the template!

noorul:
@nguyenuy Can you please re-base and push?

nguyenuy:
@noorul Just re-based and pushed

job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
```diff
@@ -52,8 +52,10 @@ trait FileCacher {
         if (name.startsWith(prefix)) {
           val suffix = name.substring(prefix.length)
           (Pattern findFirstIn suffix).isDefined
```
noorul: Why do we need this line if we are going to return true always?

noorul:
@nguyenuy Also test is failing, see https://travis-ci.org/spark-jobserver/spark-jobserver/builds/237579990#L2387

job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
nguyenuy: Got rid of this in latest commit

nguyenuy:
@noorul addressed comments and tests are currently passing

job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
```diff
@@ -50,10 +50,10 @@ trait FileCacher {
       override def accept(dir: File, name: String): Boolean = {
         val prefix = appName + "-"
         if (name.startsWith(prefix)) {
-          val suffix = name.substring(prefix.length)
-          (Pattern findFirstIn suffix).isDefined
```
noorul: Why are you removing line 53 and 54?

nguyenuy: This is extra code that isn't needed. The focus here should be on binary names prefix matches. If a user submits a DELETE for binary "example", then I would assume all binaries in the user specified location will be deleted and here the prefix is determined as "example-"

noorul: I think it is ensuring that the filename has suffix as date. Why are we removing this logic?

job-server/src/main/scala/spark/jobserver/BinaryManager.scala
```diff
@@ -106,6 +107,9 @@ class BinaryManager(jobDao: ActorRef) extends InstrumentedActor {
 
     case DeleteBinary(appName) =>
       logger.info(s"Deleting binary $appName")
-      deleteBinary(appName).pipeTo(sender)
+      deleteBinary(appName).map{
```
velvia: space between map and {

job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
nguyenuy: Adding this logic back

noorul:
Since @velvia's comment is also addressed, I am merging this.

Thank you @nguyenuy

PR #841, opened 2017-05-31
misc: fix java 7 compatibility

derSascha:
**Current behavior :** 
Dependency versions are selected by local java compiler version rather than target java version.

**New behavior :**
* Use `JAVA_VERSION` environment variable and java compiler version to decide what dependency version are chosen (java 7 or 8 compatible). Can be used in build config via `export JAVA_VERSION=7-jdk` in `config/....sh`.
* Downgrade/change akka and postgres versions in java 7 builds.

**Other information**:
Latest CDH 5.10 sandbox has java 7 installed...

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Also see [Chinese docs / 中文](doc/chinese/job-server.md).
   - [WordCountExample walk-through](#wordcountexample-walk-through)
     - [Package Jar - Send to Server](#package-jar---send-to-server)
     - [Ad-hoc Mode - Single, Unrelated Jobs (Transient Context)](#ad-hoc-mode---single-unrelated-jobs-transient-context)
-    - [Persistent Context Mode - Faster & Required for Related Jobs](#persistent-context-mode---faster-&-required-for-related-jobs)
+    - [Persistent Context Mode - Faster & Required for Related Jobs](#persistent-context-mode---faster--required-for-related-jobs)
 - [Create a Job Server Project](#create-a-job-server-project)
   - [Creating a project from scratch using giter8 template](#creating-a-project-from-scratch-using-giter8-template)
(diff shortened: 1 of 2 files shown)


noorul:
I think there was a discussion that we will only be supporting java 8 from spark 2.0 support. @spark-jobserver/committers  ?

amarouni:
It's true that some Hadoop distribution are still using java7 in their sandboxes (Cloudera comes to mind), but they all support java8 and any serious production deployment will use java8. One exception is Mapr that still requires java7 but that's a different story.

I'm on favor of moving forward and requiring java8.

koettert:
Most of our customers are still running their Hadoop clusters with Java 7. Also Java 7 is still the default when setting up a new Cloudera or Hortonworks cluster even though they also support Java 8. That is why Sascha has initiated the pull request.

derSascha:
Effort to support java 7 at this point is relative low. Just downgrade the packages and everything works like expected. Why not discuss this if heavy changes are happen and supporting java 7 becomes difficult?

Another CDH fun fact:
After deploying CDH 5.10 via parcels, java 7 and 8 exists in `/usr/java`. Default java is now java 8, but `/opt/cloudera/parcels/CDH/lib/bigtop-utils/bigtop-detect-javahome` detects java 7...

hntd187:
I think this is fine for now, but I'd really like to stop doing this and just draw the line in the sand about java7 and scala 2.10

zeitos:
I think we should deprecate it but try to maintain compatibility until spark removes it, basically: if the version of spark we support supports that version of java we should *try* to support it.

*IF* spark 2.2 removes java7 support, then at that time we should do that too.

derSascha:
Feature and java 7 warning added to README and TOC fixed (someone forget to run doctoc?).

velvia:
@derSascha thanks for the doc fixes (including Doctoc - it's not automatic, maybe we should make this automated)

PR #869, opened 2017-06-16
feat(webapi): add route for getting context info

addisonj:
**Pull Request checklist**

Currently, you can't easily get the URL of the web API or other info about a context
**New behavior :**

This adds a route to retrieve information about a context, `GET /contexts/<name>` which returns some info. At the moment, this is primarily just the spark web UI url

**BREAKING CHANGES**

No breaking changes
**Other information**:

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -31,14 +31,19 @@ object JobManagerActor {
 
   case object GetContextConfig
   case object SparkContextStatus
+  case object GetSparkWebUIUrl
 
   // Results/Data
   case class ContextConfig(contextName: String, contextConfig: SparkConf, hadoopConfig: Configuration)
   case class Initialized(contextName: String, resultActor: ActorRef)
   case class InitError(t: Throwable)
   case class JobLoadingError(err: Throwable)
+  case class SparkWebUIUrl(url: String)
   case object SparkContextAlive
   case object SparkContextDead
+  case object NoSparkWebUI
+
+
 
   // Akka 2.2.x style actor props for actor creation
   def props(contextConfig: Config, daoActor: ActorRef): Props = Props(classOf[JobManagerActor],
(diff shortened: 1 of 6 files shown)


velvia:
@noorul and @hntd187 in particular, have a look?

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
```diff
@@ -209,6 +214,27 @@ class JobManagerActor(contextConfig: Config, daoActor: ActorRef)
         }
       }
     }
+    case GetSparkWebUIUrl => {
+      if (jobContext.sparkContext == null) {
```
hntd187: wouldn't you want `jobContext.sparkContext.isStopped` instead? I'd prefer anything over a null check

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -342,6 +343,22 @@ class WebApi(system: ActorSystem,
 
     // user authentication
     authenticate(authenticator) { authInfo =>
+      (get & path(Segment)) { (contextName) =>
+        respondWithMediaType(MediaTypes.`application/json`) { ctx =>
+          val future = supervisor ? GetSparkWebUI(contextName)
+          future.map {
+            case WebUIForContext(name, Some(url)) => ctx.complete(
+              200,
```
hntd187: why not just put the whole `ctx.complete()` on another line so it can all be together on the same line? Not a big deal

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -708,10 +727,15 @@ class WebApi(system: ActorSystem,
...
+        Left(GetContext(context.get))
       } else {
-        StartAdHocContext(classPath, contextConfig)
+        Right(StartAdHocContext(classPath, contextConfig))
       }
+    handleGetManager(msg)
+  }
+
+  private def handleGetManager(msgContainer: Either[GetContext, StartAdHocContext]): Option[ActorRef] = {
+    val msg = msgContainer.left.getOrElse(msgContainer.right.get)
```
hntd187: I'm not sure what an Either provides here, so this is an `Either[GetContext, StartAdHocContext]` ? Is this different than what you do above to switch the message based on a condition? you could just remove the `private def handleGetManager` entirely then.

Meehau:
@addisonj This is a great feature :) Maybe you could give me (@meehau) and another user (@zromi18) read+write access to your branch so that we could help you out with it?

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
addisonj: This keeps it consistent with the checks in the rest of the file. Also, I am fairly sure that due to the way the jobContext controls the lifetime of sparkContext, `isStopped` would result in a NPE. Might be a bigger refactor to make that doable

addisonj:
oookay, rebased and also somehow I totally borked this merge from our internal repo, so this actually has a chance of working now :)

This also addresses some of your changes @hntd187, except for the null check, which seems to be a more painful thing but I agree would be good to address (but in another commit, seems more of a refactor than I would want to do here)

noorul:
@addisonj Thank you! Merged!

Issue #786, opened 2017-04-05
Concurrent access to the jars when cashing on the file system resulting in ClassNotFoundError


zromi18:
When the SparkJob is run for the first time in two or more SparkContext at the same time in context-per-jvm mode, all the processes/jvms try to cache the same jar - literally the same file in the file system. The result of such a scenario is ClassNotFoundError, event thouth the original jar is there and in a good shape. I suggest either synchronization when cashing jars on the file system or using process/jvm specific files in the file system for jars, so there is no collision between processes/jvms.

amarouni:
@zromi18 I had a similar issue so this might help, are you using the H2 file database ? If so I'd suggest using the H2 server database when working with the context-per-jvm mode.

zromi18:
@amarouni Thank you for the advice, but that is not the case. I am using MySQL database and the nature of the problem is as I have described in the first comment.

Meehau:
The fix for this issue was successfully merged ( https://github.com/spark-jobserver/spark-jobserver/pull/872 ). I think we can close this issue. @zromi18 - please close

zromi18:
@Meehau great job. Thank you.

PR #872, opened 2017-06-21
fix(job-server): Fix for issue #786 - Concurrent access to the jars w…

Meehau:
…hen cashing on the file system resulting in ClassNotFoundError

* Modify/extend the FileCacher.cacheBinary(...) method to store the JAR/binary data into a temporary file first, and only after this completes, try to rename it to the expected full original name (with a new timestamp and so on). If the rename succeeds - everything is fine. If the rename fails (ex. there is already a file with the same target full filename), then it means that the other process was faster than us, and we're OK with this (but we have to delete our temporary file, its not needed anymore).

**Pull Request checklist**

**Current behavior :**
Main issue: #786 

**New behavior :**
As described in the commit message (see above)

**BREAKING CHANGES**
None

**Other information**:
None

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
+++ b/job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
@@ -27,24 +27,15 @@ trait FileCacher {
     appName + "-" + uploadTime.toString("yyyyMMdd_hhmmss_SSS") + s".${binaryType.extension}"
   }
 
-  // Random temporary filename suffix possible chars
-  val tempFileNamePossibleChars: Seq[Char] = ('a' to 'z') ++ ('0' to '9')
-  // Generate random temporary filename suffix
-  def randomTempFileNameSuffix(length: Int = 8): String = {
-    import scala.util.Random
-    (1 to length).map(_ => tempFileNamePossibleChars(Random.nextInt(tempFileNamePossibleChars.size)))
-      .mkString("_tmp-", "", "")
-  }
-
   // Cache the jar file into local file system.
   protected def cacheBinary(appName: String,
                             binaryType: BinaryType,
                             uploadTime: DateTime,
                             binBytes: Array[Byte]) {
     val targetFullBinaryName = createBinaryName(appName, binaryType, uploadTime)
-    val tempSuffix = randomTempFileNameSuffix()
-    val tempBinaryName = targetFullBinaryName + tempSuffix
-    val tempOutFile = new File(rootDir, tempBinaryName)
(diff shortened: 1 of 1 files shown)


job-server/src/main/scala/spark/jobserver/io/FileCacher.scala
```diff
@@ -27,21 +27,45 @@ trait FileCacher {
     appName + "-" + uploadTime.toString("yyyyMMdd_hhmmss_SSS") + s".${binaryType.extension}"
   }
 
+  // Random temporary filename suffix possible chars
```
hntd187: http://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-

Why make your own?

Meehau: Hi, thanks for the review! Here is my reasoning:

- When using the Java implementation, I cannot control the exact pattern of the file (the "random" part). And I wanted it to have exact, explicitly adjustable length and the character set to choose from.

- When using the Java implementation, I cannot control the exact folder in which the temporary file is stored. Its system-specific. Here we want to control the file location, because of three reasons:

  1. The rename operation is faster and has much less I/O overhead than moving/copying the file itself (which we would have to do, since this would be a different folder than our SJS binaries one)
  2. If the automatic deletion fails, we want to be able find the file easily than to search through the various system tmp directories
  3. If the automatic deletion fails, the FileCacher.cleanCacheBinaries(...) method also will be able to delete the temporary JARs during its normal execution

hntd187: I don't see your reasoning for wanting to control the random part, it's a temp file, it really shouldn't matter, but you can add whatever you want to the suffix and prefix. Also the second point, you can control the file location, it's there in the documentation.

Meehau: I wanted to have more control over the file created (also having future development in mind), but as long as I can control the target directory ( https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-java.io.File- ) I can and will change it to the Java implementation.

hntd187: I think less code to maintain is good for future development :-)

velvia:
I would agree with @htnd187, but overall thanks for fixing this!

Meehau:
@hntd187 I've just pushed my changes, could you review it? Thanks :)

Issue #736, opened 2017-02-12
Who is saying `Guava Issue #1635`?


yiu31802:
Hello,

I am still new to Spark and this Server.

I am trying to setup a spark-job-server which accesses to Cassandra DB to get dataframe back. I've been struggling for several days already, but I am not really sure about where this `Guava` issue is actually raised. I am already aware that this comes from Guava's versions conflict of Spark's and Cassandra's. So far, none of Internet investigation helped me to resolve this issue.

- Am I doing wrong with `server_start.sh --jars spark-cassandra-connector-2.0.0-M2-s_2.11.jar`? But I made sure this `connector` jar has `guava-18.0` according to its `pom.xml`.
- Am I doing wrong with my job? But it also has `16.0.1`. (My job is pasted below)
- my Spark version is `2.0.2`
- spark-job-server is `spark-2.0-preview`.
- Cassandra version is bit old 2.2.8, but I believe this is not an issue.
- I have ever successfully used `spark-cassandra-connector` in the same setup but not via spark-job-server.

```
[2017-02-12 21:09:20,815] ERROR .jobserver.JobManagerActor [] [akka://JobServer/user/context-supervisor/mycontext] - Got Throwable
java.lang.ExceptionInInitializerError
        at com.datastax.spark.connector.cql.DefaultConnectionFactory$.clusterBuilder(CassandraConnectionFactory.scala:35)
        at com.datastax.spark.connector.cql.DefaultConnectionFactory$.createCluster(CassandraConnectionFactory.scala:92)
        at com.datastax.spark.connector.cql.CassandraConnector$.com$datastax$spark$connector$cql$CassandraConnector$$createSession(CassandraConnector.scala:154)
        at com.datastax.spark.connector.cql.CassandraConnector$$anonfun$3.apply(CassandraConnector.scala:149)
        at com.datastax.spark.connector.cql.CassandraConnector$$anonfun$3.apply(CassandraConnector.scala:149)
        at com.datastax.spark.connector.cql.RefCountedCache.createNewValueAndKeys(RefCountedCache.scala:31)
        at com.datastax.spark.connector.cql.RefCountedCache.acquire(RefCountedCache.scala:56)
        at com.datastax.spark.connector.cql.CassandraConnector.openSession(CassandraConnector.scala:82)
        at com.datastax.spark.connector.cql.CassandraConnector.withSessionDo(CassandraConnector.scala:110)
[...]

PR #903, opened 2017-08-05
Fixing Guava 1635 issue with spark cassandra connector + other minor …

sreeraaman:
This PR fixes 4 issues as noted below:

1. Currently, the spark jobserver does not work nicely with the spark cassandra connector because of the Guava issue #736 because of the Cassandra driver leaking Guava 16+ whereas spark uses Guava 14+.

The resolution was to remove cassandra-drive-core and cassandra-driver-mapping and instead use spark-cassandra-connector as the dependency.

2. While building a docker image for spark 2.0+ from source tree, there seems to be an issue with the make_distribution.sh that got moved into a separate folder /dev under the source tree. However, for spark 1.6+, the make_distribution.sh seems to be at the root of the source tree. 

This has been fixed with a if else construct by checking the Versions.spark value.

3. Even after adding the spark-cassandra-connector as a dependency, it was not getting pulled because of the exclusion list in Assembly which had spark as a filter.

This was fixed by removing the spark filter from the filter list.

4. Finally while building a docker image, an error occurs after tagging where it detects the docker version and with my version of docker (Version 17.06.0-ce-mac19 (18663)), the sbt docker plugins fails with an error unable to parse version.

Looks like this was fixed in version 1.4.1.

5. Finally, a minor change was to bump the spark version from the default 2.1.0 to 2.1.1

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/903)


job-server/config/docker.conf
```diff
@@ -33,13 +33,14 @@ spark {
...
 
   # predefined Spark contexts
-  # contexts {
-  #   my-low-latency-context {
-  #     num-cpu-cores = 1           # Number of cores to allocate.  Required.
-  #     memory-per-node = 512m         # Executor memory per node, -Xmx style eg 512m, 1G, etc.
-  #   }
-  #   # define additional contexts here
-  # }
+   contexts {
```
noorul: Why are we setting the default to 4G?

job-server/config/local.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
noorul: I think you should change the template file.

project/Assembly.scala
```diff
@@ -5,12 +5,16 @@ import sbtassembly.AssemblyPlugin.autoImport._
 object Assembly {
   lazy val settings = Seq(
     assemblyJarName in assembly := "spark-job-server.jar",
-    // uncomment below to exclude tests
+/*
+    assemblyShadeRules in assembly := Seq(ShadeRule.rename("com.google.**" -> "sjs.com.google.@1").inAll),
```
noorul: Why are you adding commented code?

project/Assembly.scala
```diff
@@ -5,12 +5,16 @@ import sbtassembly.AssemblyPlugin.autoImport._
...
+*/
+
+      // uncomment below to exclude tests
     // test in assembly := {},
     assemblyExcludedJars in assembly <<= (fullClasspath in assembly) map { _ filter { cp =>
       List("servlet-api", "guice-all", "junit", "uuid",
         "jetty", "jsp-api-2.0", "antlr", "avro", "slf4j-log4j", "log4j-1.2",
-        "scala-actors", "spark", "commons-cli", "stax-api", "mockito").exists(cp.data.getName.startsWith(_))
```
noorul: I am not sure this is required. Can we use master=local without this?

project/Dependencies.scala
```diff
@@ -55,9 +55,12 @@ object Dependencies {
     "org.flywaydb" % "flyway-core" % flyway
   )
 
+
+
   lazy val cassandraDeps = Seq(
-    "com.datastax.cassandra" % "cassandra-driver-core" % cassandra,
-    "com.datastax.cassandra" % "cassandra-driver-mapping" % cassandra
+    /*"com.datastax.cassandra" % "cassandra-driver-core" % cassandra,
+    "com.datastax.cassandra" % "cassandra-driver-mapping" % cassandra*/
```
noorul: Why do we need commented code?

noorul:
Can you remove .DS_Store file?

project/Versions.scala
```diff
@@ -1,7 +1,7 @@
 import scala.util.Properties.isJavaAtLeast
 
 object Versions {
-  lazy val spark = sys.env.getOrElse("SPARK_VERSION", "2.1.0")
+  lazy val spark = sys.env.getOrElse("SPARK_VERSION", "2.1.1")
```
noorul: Latest is 2.2.0, can we go to that?

build.sbt
```diff
@@ -137,7 +137,13 @@ lazy val dockerSettings = Seq(
     val sparkBuild = s"spark-${Versions.spark}"
     val sparkBuildCmd = scalaBinaryVersion.value match {
       case "2.11" =>
-        "./make-distribution.sh -Dscala-2.11 -Phadoop-2.7 -Phive"
+        if (Versions.spark.startsWith("1")){
+          "./make-distribution.sh -Dscala-2.11 -Phadoop-2.7 -Phive"
+        }
+        else {
```
zeitos: we usually do
```
} else {
```

also: consider doing 
```
Versions.spark match {
   case s if s.startsWith("1") => {}
   case _ => {}
}
```

sreeraaman:
@noorul. I have applied the changes as requested.

noorul:
@sreeraaman You need to re-base as C* connector version that we use has changed in master.

velvia:
This looks fine to me.   As long as @noorul is happy  :)

velvia:
@sreeraaman are you still working on this PR?

.gitignore
```diff
@@ -9,6 +9,7 @@ test-reports/
 .idea*
 .vagrant
 *.pyc
+.DS_Store
```
noorul: This should be in your global gitignore.

project/Assembly.scala
```diff
@@ -5,12 +5,12 @@ import sbtassembly.AssemblyPlugin.autoImport._
 object Assembly {
   lazy val settings = Seq(
     assemblyJarName in assembly := "spark-job-server.jar",
-    // uncomment below to exclude tests
+      // uncomment below to exclude tests
```
noorul: This change is not required.

zeitos:
LGTM with @noorul 's comments

PR #918, opened 2017-09-04
fix(job-server): fix for 916,917: logging for manager_start.sh

marcoboi:
* Add logging when starting context through manager_start.sh
* manager_start.sh process is now fully asynchronous
* spark-submit logs routed to main application logs

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/918)

### DIFF
--- a/bin/manager_start.sh
+++ b/bin/manager_start.sh
@@ -49,3 +49,4 @@ else
 fi
 
 eval $cmd
+
--- a/bin/manager_start.sh
+++ b/bin/manager_start.sh
@@ -48,5 +48,4 @@ else
   $appdir/spark-job-server.jar $2 $3 $4 $conffile'
 fi
 
-eval $cmd > /dev/null 2>&1 &
-# exec java -cp $CLASSPATH $GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES $MAIN $1 $2 $conffile 2>&1 &
+eval $cmd
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -20,6 +20,7 @@ import spark.jobserver.common.akka.InstrumentedActor
 import scala.concurrent.Await
 import akka.pattern.gracefulStop
 import org.joda.time.DateTime
+import org.slf4j.LoggerFactory
 import spark.jobserver.io.JobDAOActor.CleanContextJobInfos
 
 /**
@@ -236,28 +237,11 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
       cmdString = cmdString + s" ${contextConfig.getString(SparkJobUtils.SPARK_PROXY_USER_PARAM)}"
     }
 
-    val pb = Process(cmdString)
-    val pio = new ProcessIO(_ => (),
-                        stdout => scala.io.Source.fromInputStream(stdout)
-                          .getLines.foreach(println),
-                        stderr => scala.io.Source.fromInputStream(stderr).getLines().foreach(println))
-    logger.info("Starting to execute sub process {}", pb)
-    val processStart = Try {
-      val process = pb.run(pio)
-      if (waitForManagerStart) {
-        val exitVal = process.exitValue()
-        if (exitVal != 0) {
-          throw new IOException("Failed to launch context process, got exit code " + exitVal)
-        }
-      }
-    }
+    val contextLogger = LoggerFactory.getLogger("manager_start")
+    val process = Process(cmdString.split(" "))
+    process.run(ProcessLogger(out => contextLogger.info(out), err => contextLogger.warn(err)))
 
-    if (processStart.isSuccess) {
-      contextInitInfos(contextActorName) = (isAdHoc, successFunc, failureFunc)
-    } else {
-      failureFunc(processStart.failed.get)
-    }
-
+    contextInitInfos(contextActorName) = (isAdHoc, successFunc, failureFunc)
   }
 
   private def createContextDir(name: String,
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -1,6 +1,5 @@
 package spark.jobserver
 
-import java.io.IOException
 import java.nio.file.{Files, Paths}
 import java.nio.charset.Charset
 import java.util.concurrent.TimeUnit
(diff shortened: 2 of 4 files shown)


marcoboi:
This commit addresses the issues I pointed out in 916 and 917.

Background:
In the previous setup, the Process in charge of executing the manager_start script would run synchronously. Yet, as the script executed the spark-submit command as a forked process, synchronous execution could not be guaranteed for that crucial part. Hence:
* errors in spark-submit would not make the script fail
* sterr and stdout from spark-submit could be logged (also due to some improper logging).
* spark-submit, the most important bit of the script, was anyway being executed asynchronously (forked off) from a sync process. This was also deceiving for someone who would read the code.
Since starting a new context is an asynchronous operation by its nature, the spark-submit command should be directly executed by an asynchronous  Process.

What was done:
* changed the manager_start process to run asynchronously
* created a custom logger for that Process 
* removed the forking when the spark-submit command was executed in order to route the stderr and stdout of that spark-submit command to the main logs

Remarks:
Since the process is executed asynchronously, there is now no way to tell if the process failed or succeeded. Hence, the failureFunction and successFunction (see line 218) do not make sense anymore. Do those function accomplish any other tasks of can they be removed?

thoHeinze:
@marcoboi I personally totally get your point that you want to see the logs of the spark-submit. But I would prefer if you not append this logs directly to the spark-jobserver.log (where they will be very hard to find), but rather write it into a separate file for each context, similar to the logs the context writes. This could be accomplished by some small changes into the manager_start.sh script.

marcoboi:
@thoHeinze, I think those logs are best suited being part of the main application's logs. Here're the reasons in support of my approach:
* there are no context's logs until the spark-submit has succeeded. 
* If the spark-command fails, there is no context being created and no context's logs to append that information to
* the logs of the spark-submit could be appended to a separate file in the context's folder, but that would create one extra file to check if things go wrong. Also, that file would only contain the logs from the manager_start, which seems a bit of an overkill.
* the AkkaClusterSupervisor is the entity that issues the spark-submit, so it should be the one receiving the logs

thoHeinze:
@marcoboi I just have the feeling that it will be hard to search errors if one specific context fails and you have many contexts running in parallel. Especially, your current log will only contain something like no "NoClassDefException", but no context id.

velvia:
@marcoboi so just to understand, with this change, after manager_start.sh succeeds and the forked process is up and running, where will the logs go?
The reason why the logs were forked to begin with is to aid in debugging, this was something I found helpful personally and this sentiment seems to be echoed by users.

marcoboi:
Once the JobManager is up and running, it will log into its own folder. Nothing changed there.
What this commit adds is logging for the process that executes the script.
Let me know if anything is unclear there, I'll try to clarify further.

marcoboi:
@thoHeinze, that is an error that would typically happen if your JobManager cannot find the class of the job you want to execute. My PR does not change that. 
Just to reiterate, the logging of the JobManager will stay as it is, i.e., a separate file. Once the JobManager is up, it will have its own logging.
The problem I addresses here was that the JobManager is started by executing a shell script in a Java process. That shell script did not produce any logging. Now it does, and that logging is redirected to the main application.
Hope this helps.

thoHeinze:
@marcoboi You don't get my point. The error above you won't currently find in any log, because it will only be shown in the error output of the spark submit. Your PR will make this problem finally visible, but imho at the wrong place.

marcoboi:
@thoHeinze, leaving aside for a minute which logs we're writing to, let me point out that the changes I suggested will output the stderr and stdout of the execution of _whole_ of the manager_start.sh to the main logs.
That includes the spark-submit command, but also _all the logic_ that is currently executed in the manager_start.
The manager_start script executes a long list of checks from _spark-jobserver/bin/setenv.sh_. 
All those checks will silently fail if not exposed through the logs.

velvia:
Hey guys, I’m going to settle the debate a bit and say that overall this is still a great step forward, regardless of the exact logging output of manager_start.   Sorry that I’ve been too busy to review the code but will try to do so soon.

If the main log is not desired, perhaps an option can be added to output it somewhere else, but IMHO this can be done later.

thoHeinze:
@velvia I totally agree to your statement.

marcoboi:
@velvia, @thoHeinze, cool, thanks.

bin/manager_start.sh
```diff
@@ -48,5 +48,4 @@ else
   $appdir/spark-job-server.jar $2 $3 $4 $conffile'
 fi
 
-eval $cmd > /dev/null 2>&1 &
-# exec java -cp $CLASSPATH $GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES $MAIN $1 $2 $conffile 2>&1 &
+eval $cmd
```
zeitos: can you add a blank line at the end of the file?

marcoboi: @zeitos Nicely spotted. Just did that.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -236,28 +237,11 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef) extends InstrumentedActor {
...
 
-    val pb = Process(cmdString)
-    val pio = new ProcessIO(_ => (),
-                        stdout => scala.io.Source.fromInputStream(stdout)
-                          .getLines.foreach(println),
-                        stderr => scala.io.Source.fromInputStream(stderr).getLines().foreach(println))
-    logger.info("Starting to execute sub process {}", pb)
-    val processStart = Try {
-      val process = pb.run(pio)
-      if (waitForManagerStart) {
```
derSascha: @marcoboi does the `wait-for-manager-start` config has any effect anymore? if not, please remove them everywhere in this class and in the default `application.conf`.

derSascha:
Logging the output of `manager_start.sh` into the global log should be ok. Only on heavy errors (wrong classpath etc.) the start fails and logs some output.

> Hence, the failureFunction and successFunction (see line 218) do not make sense anymore.

https://github.com/spark-jobserver/spark-jobserver/blob/b88c21fa1264504fc6b98b4ca392ac6d59896597/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala#L196-L217

They make sense, e.g. they handle context creation timeouts and successful context initialization.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
marcoboi: ok

marcoboi: removed

PR #922, opened 2017-09-13
Add YARN cluster mode and simplify cluster deployment

### LINKED ISSUE #915: Context's Json should not be passed as Java argument when starting Context
A Json representing the context's settings is passed as argument to the {{spark-submit}} command that starts the JobManager.
Json encoding does not comply with Java parameter encoding requirements. Therefore, the spark-submit command (which ultimately relies on the java command) will not be able to parse the Json string correctly.
For instance, Json strings can contain white spaces, which are interpreted as parameter separators by Java.
As there is no way to guarantee the Json will have the form of a valid Java parameter, passing Json strings as Java parameters should be avoided altogether.

See also: https://stackoverflow.com/questions/23958338/how-to-stop-java-from-separating-json-string-passed-as-a-command-line-argument-i

I temporarily fixed the issue by reverting to the old system where the context string was dumped to a file and read in by the JobManager.

https://github.com/spark-jobserver/spark-jobserver/blob/eaebce3f1d32dbdbe61ee5b7b9bb2121bde85f7e/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala#L232

derSascha:
**New behavior :**

*Deploy mode and spark master in job server config*
- `spark.master` can be `local[...]`, `yarn` or `mesos://...`
- `spark.submit.deployMode` can be `client` or `cluster`

Both options are forwarded trough `manager_start.sh` to `spark-submit`. No need to edit shell variables in `settings.sh` to provide Mesos master.

*Embedded H2 server in job server config*

To launch a H2 database available to remote job manager instances in a cluster:
```
spark {
  jobserver {
    # start a H2 DB server, reachable in your cluster
    sqldao {
      jdbc {
        url = "jdbc:h2:tcp://CLUSTER-IP:9092/h2-db;AUTO_RECONNECT=TRUE"
      }
    }
    startH2Server = true
  }
}
```

*Akka interface*
To run in akka in a cluster (to interact with remote job manager) add the following config:
```
akka {
  remote.netty.tcp {
    hostname = "FRONT-END-IP"
  }
}
```
This config ensures that akka binds to right interface. The hostname gets removed in Job manager configs.

*Remote files and context config*
Some files are now transferred via `spark-submit`:
- job-server-config: default config, no need to copy this by hand to all hosts
- context.conf: each context has its own config file (this reverts #681). providing the config as JSON argument has many limitations (maximal arguments length, special charactes / white spaces and security risk). Arguments in YARN with e.g. `{{VARNAME}}` are replaced with variables at runtime and `}}` makes trouble... This might also fix #915
- log4j-cluster.properties: same config as the server options but with log output to stderr instead of a logfile. works well with YARN cluster containers.

*Remote data files and jars*
PR #924 contains a solution to access files uploaded via the data API (`/data`). Jars and other binaries are already transferred via SQL backend.

**Breaking Changes**
`spark.submit.deployMode` has to be set to `cluster` in job server config. Mesos master variable can be removed from `settings.sh`.

**Documentation**
This PR rewrites the cluster documentations in yarn/mesos docs. Im not sure if the configuration described in [Mesos](doc/mesos.md) with file based H2 database really works?!?!

**Other information**
The changes by this PR has been tested with YARN client and cluster mode. Can someone test this on a Mesos cluster?

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/922)

### DIFF
--- a/bin/manager_start.sh
+++ b/bin/manager_start.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Script to start the job manager
-# args: <work dir for context> <cluster address> [proxy_user]
+# args: <master> <deployMode> <workDir> <akkaAdress> [<proxyUser>]
 set -e
 
 get_abs_script_path() {
@@ -13,12 +13,8 @@ get_abs_script_path
 
 . $appdir/setenv.sh
 
-# Override logging options to provide per-context logging
-LOGGING_OPTS="-Dlog4j.configuration=file:$appdir/log4j-server.properties
-              -DLOG_DIR=$2"
-
 GC_OPTS="-XX:+UseConcMarkSweepGC
-         -verbose:gc -XX:+PrintGCTimeStamps -Xloggc:$appdir/gc.out
+         -verbose:gc -XX:+PrintGCTimeStamps
          -XX:MaxPermSize=512m
          -XX:+CMSClassUnloadingEnabled "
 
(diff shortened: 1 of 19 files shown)


bin/manager_start.sh
```diff
@@ -27,26 +23,34 @@ JAVA_OPTS="-XX:MaxDirectMemorySize=$MAX_DIRECT_MEMORY
...
+fi
+
+cmd='$SPARK_HOME/bin/spark-submit --class $MAIN --driver-memory $JOBSERVER_MEMORY
+      --conf "spark.executor.extraJavaOptions=$LOGGING_OPTS"
+      $SPARK_SUBMIT_OPTIONS
+      --driver-java-options "$GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES $SPARK_SUBMIT_JAVA_OPTIONS"
+      $appdir/spark-job-server.jar $2 $3 $4 $conffile'
+
+mkdir -p "$3"
+eval $cmd >"$3/spark-job-server.out" 2>&1 </dev/null &
```
derSascha: @marcoboi this might also fix #918

thoHeinze:
@derSascha I'm sorry I can't understand how you can revert #681 and cluster mode still works: Based on my understanding of your code you create a context.conf, which is created on the Jobserver VM and will not be available on the machine where the driver/ JobManager will run. But also in cluster mode you simply open a file in the current dir for reading. Can you please explain ?

thoHeinze:
Imho this change is quite big, I personally would prefer if you split this into two changes: one for the YARN cluster mode and one for the remote data file cache. This would simplify the review a lot.

derSascha:
@thoHeinze god point. the huge commit is now divided into three smaller ones.

PR #681 changed the context config handling. Before the PR,  config gets written into a `context.conf`. Due to the fact that the driver process might stay on another host in cluster deploy mode, PR #681 changed this, converts the config into JSON and adds it as argument. This has many downsides. This PR reverts this, generates the `context.conf` as bevor and use `spark-submit --files...` to copy the `context.conf` into the driver container.

bin/manager_start.sh
marcoboi: Hey @derSascha, line 56 will write down the output of the spark-submit command, but will still leave the rest of the manager_start execution muted.
What I tried to point out in #918, was the need to make the stderr and stdout of the _whole_ manager_start process visible.

bsikander:
I tried this PR with spark-standalone cluster mode.
I am getting the following error when I try to post a context.
```
Could not find system configuration file settings.conf
```

I checked, the correct settings.conf is provided to --files parameter of spark-submit. Here is the path where the JobManager is looking for the settings.conf file but the file is not there.
```
<my_dir>/spark/worker_dir/driver-2017000000000-0000/settings.conf
```

According to [this](https://stackoverflow.com/questions/34900023/read-files-sent-with-spark-submit-by-the-driver), this file will be available as soon as the Context is initialized. At the position from where I get this error, context is not initialized.

Is it due to different behavior or Yarn and stand-alone ?

bin/manager_start.sh
derSascha: @marcoboi a ok. I removed the lines from this PR, keeping the discussion about in #918.

marcoboi: thanks @derSascha, nice of you.

derSascha:
@bsikander this sounds bad. I tried this only on YARN. YARN transfers all files and then starts the JVM. Did you get the standalone cluster mode working without this PR?

bsikander:
@derSascha Yes, I got it working without this PR. My changes were based on mesos-cluster change by @LeightonWong .

derSascha:
@bsikander after playing around with sparks standalone cluster mode: how did you managed the `settings.conf` in your setup, did you copy the file on each host in your cluster?

bsikander:
@derSascha yes.

derSascha:
@bsikander do you need to copy the jobserver jar on each node in the cluster?

bsikander:
@derSascha No, sjs jar is automatically copied on slaves in Spark's worker_dir/driver-* and worker_dir/app-* directories.

derSascha:
@bsikander i did not get my test cluster running... i tried the following:
`/tmp/spark-2.1.0-bin-hadoop2.7/bin/spark-submit --class "SimpleApp" --master spark://192.168.130.2:6066 --deploy-mode cluster --jars target/scala-2.11/simple-project_2.11-1.0.jar simple-project_2.11-1.0.jar`

But then i got: `java.nio.file.NoSuchFileException: /home/.../simple-project_2.11-1.0.jar`

After adding the jar to each host in the cluster this works:
`/tmp/spark-2.1.0-bin-hadoop2.7/bin/spark-submit --class "SimpleApp" --master spark://192.168.130.2:6066 --deploy-mode cluster file:///tmp/simple-project_2.11-1.0.jar`
Spark now place a copy of `/tmp/simple-project_2.11-1.0.jar` in worker_dir/driver-* and worker_dir/app-* directories.

Are you sure that you don't have a copy of jobserver jar on each host?

bsikander:
@derSascha sorry for this confusion. I just checked, I do have a copy on the slaves.

derSascha:
Closing this until #915 has been fixed and than a solution to support standalone cluster mode might be possible.

bsikander:
@derSascha so --files parameter behaves differently for YARN and standalone cluster mode ? For standalone mode the files will only be available after the context is initialized ?

derSascha:
@bsikander yes, the `--files` parameter behaves differently.

If you run `spark-submit` with files and in YARN cluster mode, the files are transferred by YARN. The application master / driver process container is allocated, the files are placed in the container dir and the process starts. You don't need to copy any files in your cluster and YARN does all the magic.

If you run `spark-submit` with files in standalone cluster mode, you have to transfer the files manual (copy them on each host or into HDFS) and then add the files path (e.g. hdfs://...) as `--files` parameter. Spark than allocates the driver container, downloads the files into the worker dir and starts it.

After the driver process and the context is launched, the driver provides the files in the cluster to the executors (see `spark.broadcast.factory`).

May be you can try PR #927.

bsikander:
@derSascha great. 
+1 for the PR. I will give it a try.

zenglian:
@derSascha could you please take a look of issue #1010?

PR #924, opened 2017-09-14
feat(jobenv): Remote data file cache

derSascha:
**Current behavior :**
Files uploaded via Data API (`/data`) are stored on the host running the REST frontend and can't simple accessed by driver instances on other hosts in the cluster.

**New behavior :**
Files can now accessed via `JobEnvironment` and are transferred via akka:
```scala
  def runJob(sc: SparkContext, runtime: JobEnvironment, input: JobData): JobOutput = {
    runtime match {
      case remote: DataFileCache => remote.getDataFile("...")
      case local => Files.read.. // job does not support remote files
    }
```
Local copies are removed on jvm/context shutdown and on file delete against the REST API.

**BREAKING CHANGES**
No public interfaces changed. Jobs compiled against older job server version should work without any changes.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/924)

### DIFF
--- a/job-server-api/src/main/scala/spark/jobserver/api/SparkJobBase.scala
+++ b/job-server-api/src/main/scala/spark/jobserver/api/SparkJobBase.scala
@@ -1,5 +1,7 @@
 package spark.jobserver.api
 
+import java.io.{File, IOException}
+
 import com.typesafe.config.{Config, ConfigFactory}
 import org.apache.spark.SparkContext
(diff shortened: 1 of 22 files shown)


thoHeinze:
@derSascha Thanks a lot for splitting your PR! And very nice feature.

noorul:
Wow cool!

PR #927, opened 2017-09-18
Add cluster support and transfer context conf via akka

### LINKED ISSUE #915: Context's Json should not be passed as Java argument when starting Context

derSascha:
**Current behavior :** 
- Bevor #681: context config was written into a local `context.conf` file. this does not work with a job manager running on another host in the cluster (aka cluster mode)
- After #681: context config was converted into JSON and then transferred as argument (this results in problems with YARN (parsing `}}`) and has other downsides (see #915))
- Setting the spark master requires changes in job server config and `manager_start.sh` or `settings.sh`
- Cluster setup requires an external database accessible within the cluster.

**New behavior :**
*context config*
The job manager gets now initialized with cluster address and actor name as argument. After joining the cluster, the context config gets transferred via akka and the context initialized. This fixes #915.

*spark master*
The following config options can be set and will be used in `spark-submit` too:
- `spark.master` can be `local[...]`, `yarn`, `spark://...` or `mesos://...`
- `spark.submit.deployMode` can be `client` or `cluster`

*spark standalone cluster mode*
Spark does not transfer your job server config, jar and log4j config. You need to copy them on each host or adding them in e.g. HDFS and then set the following variable in your `settings.sh`: `REMOTE_JOBSERVER_DIR`
If your don't set the remote dir, files are transferred via `spark-submit --files...` (this works fine with e.g. YARN).

*embedded H2 database*
Job server can now host the database instance for you:
- set `spark.jobserver.startH2Server` to `true`
- set `spark.jobserver.sqldao.jdbc.url` to `"jdbc:h2:tcp://CLUSTER-IP:9092/h2-db;AUTO_RECONNECT=TRUE"`

*Akka interface*
To run akka in a cluster (to interact with remote job manager) add the following config:
```
akka {
  remote.netty.tcp {
    hostname = "FRONT-END-IP"
  }
}
```
This config ensures that akka binds to the right interface. The hostname option will be removed in the job manager config and the job then detects the right interface.

*Remote data files and jars*
PR #924 contains a solution to access files uploaded via the data API (`/data`). Jars and other binaries are already transferred via SQL backend.

*logging*
Same behavior as before in client mode. Cluster mode use `log4j-cluster.properties` (same config as the server options but with log output to stderr instead of a logfile). This works well with YARN and spark standalone containers.

**Other information**:
Tested on virtual cluster with YARN and spark standalone. Might be work with Mesos too.

This PR changes many files (constructor changes...). Code, test and doc changes are in its own commits to reduce the review complexity. Comments are welcome.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/927)

### DIFF
--- a/bin/manager_start.sh
+++ b/bin/manager_start.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Script to start the job manager
-# args: <work dir for context> <cluster address> [proxy_user]
+# args: <master> <deployMode> <akkaAdress> <actorName> <workDir> [<proxyUser>]
 set -e
 
 get_abs_script_path() {
@@ -13,12 +13,8 @@ get_abs_script_path
 
 . $appdir/setenv.sh
 
-# Override logging options to provide per-context logging
-LOGGING_OPTS="-Dlog4j.configuration=file:$appdir/log4j-server.properties
-              -DLOG_DIR=$2"
-
 GC_OPTS="-XX:+UseConcMarkSweepGC
-         -verbose:gc -XX:+PrintGCTimeStamps -Xloggc:$appdir/gc.out
+         -verbose:gc -XX:+PrintGCTimeStamps
          -XX:MaxPermSize=512m
          -XX:+CMSClassUnloadingEnabled "
 
@@ -27,26 +23,46 @@ JAVA_OPTS="-XX:MaxDirectMemorySize=$MAX_DIRECT_MEMORY
 
 MAIN="spark.jobserver.JobManager"
 
-MESOS_OPTS=""
-if [ $1 == "mesos-cluster" ]; then
-    MESOS_OPTS="--master $MESOS_SPARK_DISPATCHER --deploy-mode cluster"
-    appdir=$REMOTE_JOBSERVER_DIR
-fi
+# copy files via spark-submit and read them from current (container) dir
+if [ $2 = "cluster" -a -z "$REMOTE_JOBSERVER_DIR" ]; then
+  SPARK_SUBMIT_OPTIONS="$SPARK_SUBMIT_OPTIONS
+    --master $1 --deploy-mode cluster
+    --conf spark.yarn.submit.waitAppCompletion=false
+    --files $appdir/log4j-cluster.properties,$conffile"
+  JAR_FILE="$appdir/spark-job-server.jar"
+  CONF_FILE=$(basename $conffile)
+  LOGGING_OPTS="-Dlog4j.configuration=log4j-cluster.properties"
 
-if [ ! -z $5 ]; then
-  cmd='$SPARK_HOME/bin/spark-submit --class $MAIN --driver-memory $JOBSERVER_MEMORY
-  --conf "spark.executor.extraJavaOptions=$LOGGING_OPTS"
-  --proxy-user $5
-  $MESOS_OPTS
-  --driver-java-options "$GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES"
-  $appdir/spark-job-server.jar $2 $3 $4 $conffile'
+# use files in REMOTE_JOBSERVER_DIR
+elif [ $2 == "cluster" ]; then
+  SPARK_SUBMIT_OPTIONS="$SPARK_SUBMIT_OPTIONS
+    --master $1 --deploy-mode cluster
+    --conf spark.yarn.submit.waitAppCompletion=false"
+  JAR_FILE="$REMOTE_JOBSERVER_DIR/spark-job-server.jar"
+  CONF_FILE="$REMOTE_JOBSERVER_DIR/$(basename $conffile)"
+  LOGGING_OPTS="-Dlog4j.configuration=$REMOTE_JOBSERVER_DIR/log4j-cluster.properties"
+
+# client mode, use files from app dir
 else
-  cmd='$SPARK_HOME/bin/spark-submit --class $MAIN --driver-memory $JOBSERVER_MEMORY
-  --conf "spark.executor.extraJavaOptions=$LOGGING_OPTS"
-  --driver-java-options "$GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES"
-  $MESOS_OPTS
-  $appdir/spark-job-server.jar $2 $3 $4 $conffile'
+  JAR_FILE="$appdir/spark-job-server.jar"
+  CONF_FILE="$conffile"
+  LOGGING_OPTS="-Dlog4j.configuration=file:$appdir/log4j-server.properties -DLOG_DIR=$5"
+  GC_OPTS="$GC_OPTS -Xloggc:$5/gc.out"
 fi
 
-eval $cmd
+if [ -n "$6" ]; then
+  SPARK_SUBMIT_OPTIONS="$SPARK_SUBMIT_OPTIONS --proxy-user $6"
+fi
+
+if [ -n "$JOBSERVER_KEYTAB" ]; then
+  SPARK_SUBMIT_OPTIONS="$SPARK_SUBMIT_OPTIONS --keytab $JOBSERVER_KEYTAB"
+fi
+
+cmd='$SPARK_HOME/bin/spark-submit --class $MAIN --driver-memory $JOBSERVER_MEMORY
+      --conf "spark.executor.extraJavaOptions=$LOGGING_OPTS"
+      $SPARK_SUBMIT_OPTIONS
+      --driver-java-options "$GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES $SPARK_SUBMIT_JAVA_OPTIONS"
+      $JAR_FILE $3 $4 $CONF_FILE'
+
+eval $cmd 2>&1 > $5/spark-job-server.out
 
--- a/job-server-api/src/main/scala/spark/jobserver/util/SparkJobUtils.scala
+++ b/job-server-api/src/main/scala/spark/jobserver/util/SparkJobUtils.scala
@@ -130,7 +130,7 @@ object SparkJobUtils {
 
   private def getContextTimeout(config: Config, yarn : String, standalone : String): Int = {
     config.getString("spark.master") match {
-      case "yarn-client" =>
+      case "yarn" =>
         Try(config.getDuration(yarn,
               TimeUnit.MILLISECONDS).toInt / 1000).getOrElse(40)
       case _ =>
--- a/job-server-api/src/main/scala/spark/jobserver/util/SparkJobUtils.scala
+++ b/job-server-api/src/main/scala/spark/jobserver/util/SparkJobUtils.scala
@@ -82,11 +82,6 @@ object SparkJobUtils {
     // Set the Jetty port to 0 to find a random port
     conf.set("spark.ui.port", "0")
 
-    // Set spark broadcast factory in yarn-client mode
-    if (sparkMaster == "yarn-client") {
-      conf.set("spark.broadcast.factory", config.getString("spark.jobserver.yarn-broadcast-factory"))
-    }
-
     // Set number of akka threads
     // TODO: need to figure out how many extra threads spark needs, besides the job threads
     conf.set("spark.akka.threads", (getMaxRunningJobs(config) + 4).toString)
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -80,6 +80,9 @@ spark {
       }
     }
 
+    # Start embeded H2 Server (usefull in cluster deployment)
+    startH2Server = false
+
     # The ask pattern timeout for Api
     short-timeout = 3 s
 
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -1,6 +1,11 @@
 # Settings for safe local mode development
 spark {
+  # local[...], yarn or mesos://...
   master = "local[4]"
+
+  # client or cluster deployment
+  submit.deployMode = "client"
+
   # spark web UI port
   webUrlPort = 8080
 
@@ -10,12 +15,6 @@ spark {
 
     # If true, a separate JVM is forked for each Spark context
     context-per-jvm = false
-
-    # Default client mode will start up a new JobManager int local machine
-    # You can use mesos-cluster mode with REMOTE_JOBSERVER_DIR and MESOS_SPARK_DISPATCHER
-    # environment value set in xxxx.sh file to launch JobManager in remote node
-    # Mesos will take responsibility to offer resource to the JobManager process
-    driver-mode = client
 
     # Number of job results to keep per JobResultActor/context
     job-result-cache-size = 5000
@@ -80,7 +79,7 @@ spark {
       }
     }
 
-    # Start embeded H2 Server (usefull in cluster deployment)
+    # Start embedded H2 Server (usefull in cluster deployment)
     startH2Server = false
 
     # The ask pattern timeout for Api
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -97,11 +97,6 @@ spark {
 
     # in yarn deployment, time out for job server to wait while creating contexts
     yarn-context-creation-timeout = 40 s
-
-    # spark broadcst factory in yarn deployment
-    # Versions prior to 1.1.0, spark default broadcast factory is org.apache.spark.broadcast.HttpBroadcastFactory.
-    # Can't start multiple sparkContexts in the same JVM with HttpBroadcastFactory.
-    yarn-broadcast-factory = org.apache.spark.broadcast.TorrentBroadcastFactory
   }
 
   # predefined Spark contexts
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -59,7 +59,8 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
   //TODO: try to pass this state to the jobManager at start instead of having to track
   //extra state.  What happens if the WebApi process dies before the forked process
   //starts up?  Then it never gets initialized, and this state disappears.
-  private val contextInitInfos = mutable.HashMap.empty[String, (Boolean, ActorRef => Unit, Throwable => Unit)]
+  private val contextInitInfos = mutable.HashMap.empty[String,
+                                                      (Config, Boolean, ActorRef => Unit, Throwable => Unit)]
 
   // actor name -> (JobManagerActor ref, ResultActor ref)
   private val contexts = mutable.HashMap.empty[String, (ActorRef, ActorRef)]
@@ -95,9 +96,10 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
         val actorName = actorRef.path.name
         if (actorName.startsWith("jobManager")) {
           logger.info("Received identify response, attempting to initialize context at {}", memberActors)
-          (for { (isAdHoc, successFunc, failureFunc) <- contextInitInfos.remove(actorName) }
+          (for { (contextConfig, isAdHoc, successFunc, failureFunc) <- contextInitInfos.remove(actorName) }
            yield {
-             initContext(actorName, actorRef, contextInitTimeout)(isAdHoc, successFunc, failureFunc)
+             initContext(contextConfig, actorName,
+                         actorRef, contextInitTimeout)(isAdHoc, successFunc, failureFunc)
            }).getOrElse({
             logger.warn("No initialization or callback found for jobManager actor {}", actorRef.path)
             actorRef ! PoisonPill
@@ -183,7 +185,10 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
       cluster.down(actorRef.path.address)
   }
 
-  private def initContext(actorName: String, ref: ActorRef, timeoutSecs: Long = 1)
+  private def initContext(contextConfig: Config,
+                          actorName: String,
+                          ref: ActorRef,
+                          timeoutSecs: Long = 1)
                          (isAdHoc: Boolean,
                           successFunc: ActorRef => Unit,
                           failureFunc: Throwable => Unit): Unit = {
@@ -191,7 +196,7 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
 
     val resultActor = if (isAdHoc) globalResultActor else context.actorOf(Props(classOf[JobResultActor]))
     (ref ? JobManagerActor.Initialize(
-      Some(resultActor), dataManagerActor))(Timeout(timeoutSecs.second)).onComplete {
+      contextConfig, Some(resultActor), dataManagerActor))(Timeout(timeoutSecs.second)).onComplete {
       case Failure(e: Exception) =>
         logger.info("Failed to send initialize message to context " + ref, e)
         cluster.down(ref.path.address)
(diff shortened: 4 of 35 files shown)


bin/manager_start.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Script to start the job manager
-# args: <work dir for context> <cluster address> [proxy_user]
+# args: <master> <deployMode> <akkaAdress> <actorName> <workDir> [<proxyUser>]
```
velvia: Do you think the actor path would be safer and more future-proof than just the actor name?

bin/manager_start.sh
```diff
@@ -27,26 +23,44 @@ JAVA_OPTS="-XX:MaxDirectMemorySize=$MAX_DIRECT_MEMORY
...
-MESOS_OPTS=""
-if [ $1 == "mesos-cluster" ]; then
-    MESOS_OPTS="--master $MESOS_SPARK_DISPATCHER --deploy-mode cluster"
-    appdir=$REMOTE_JOBSERVER_DIR
-fi
+# copy files via spark-submit and read them from current (container) dir
+if [ $2 = "cluster" -a -z "$REMOTE_JOBSERVER_DIR" ]; then
+  SPARK_SUBMIT_OPTIONS="$SPARK_SUBMIT_OPTIONS
+    --master $1 --deploy-mode cluster
+    --files $appdir/log4j-cluster.properties,$conffile"
```
velvia: Does the --files option handle a directory?  If so we can package all the files inside a subdir of $appdir.  Maybe not worth the change given there are only two files.

velvia:
@derSascha thanks so much for the PR.   It'll take a while to go through all the files, I just started.  

On more of an organizational note, would like to get you and some of the other current PR authors on Gitter so we can chat about how to merge all these PRs in.   Would that work?  We need to figure out some logistics, and folks would like a release as well.

bin/manager_start.sh
derSascha: What path did you mean / suggest?

I did not change anything on this:
https://github.com/spark-jobserver/spark-jobserver/blob/306335e76da63190d2b0b2b26f790e7675097bd6/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala#L222
https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server/src/main/scala/spark/jobserver/JobManager.scala#L75

bin/manager_start.sh
derSascha: Not sure how directories are handled. What additional files do we need? The job-server.jar + spark jar and all other one in classpath are already uploaded in a YARN setup (no need to add them again). The spark standanlone mode behaves different and did not upload any files in cluster mode...

derSascha:
@velvia thanks for starting the review. is there a special channel on gitter or do you mean the main channel https://gitter.im/spark-jobserver/spark-jobserver ?

bsikander:
@derSascha 
Sorry for a delayed response.

I tried cluster mode. Here is what I found

In cluster mode, I started a context, started a job in a context and then deleted the context. Then I checked the job through curl and it was killed but when i went to Spark master UI, the driver was still running. Which is strange.
screen shot 2017-09-26 at 6 27 39 pm


Here is the log
```
2017-09-26 16:55:29,856 ERROR pool-4-thread-1 spark.jobserver.JobManagerActor []: Got Throwable
java.lang.IllegalStateException: Cannot call methods on a stopped SparkContext.
This stopped SparkContext was created at:

org.apache.spark.SparkContext.<init>(SparkContext.scala:76)
org.apache.spark.streaming.StreamingContext$.createNewSparkContext(StreamingContext.scala:836)
org.apache.spark.streaming.StreamingContext.<init>(StreamingContext.scala:84)
spark.jobserver.context.StreamingContextFactory$$anon$1.<init>(StreamingContextFactory.scala:18)
spark.jobserver.context.StreamingContextFactory.makeContext(StreamingContextFactory.scala:18)
spark.jobserver.context.StreamingContextFactory.makeContext(StreamingContextFactory.scala:8)
spark.jobserver.context.SparkContextFactory$class.makeContext(SparkContextFactory.scala:64)
spark.jobserver.context.StreamingContextFactory.makeContext(StreamingContextFactory.scala:8)
spark.jobserver.JobManagerActor$$anonfun$wrappedReceive$1.applyOrElse(JobManagerActor.scala:169)
scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
spark.jobserver.common.akka.ActorStack$$anonfun$receive$1.applyOrElse(ActorStack.scala:33)
scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
spark.jobserver.common.akka.Slf4jLogging$$anonfun$receive$1$$anonfun$applyOrElse$1.apply$mcV$sp(Slf4jLogging.scala:25)
spark.jobserver.common.akka.Slf4jLogging$class.spark$jobserver$common$akka$Slf4jLogging$$withAkkaSourceLogging(Slf4jLogging.scala:34)
spark.jobserver.common.akka.Slf4jLogging$$anonfun$receive$1.applyOrElse(Slf4jLogging.scala:24)
scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
spark.jobserver.common.akka.ActorMetrics$$anonfun$receive$1.applyOrElse(ActorMetrics.scala:23)
akka.actor.Actor$class.aroundReceive(Actor.scala:484)
spark.jobserver.common.akka.InstrumentedActor.aroundReceive(InstrumentedActor.scala:8)
akka.actor.ActorCell.receiveMessage(ActorCell.scala:526)

The currently active SparkContext was created at:

(No active SparkContext.)

	at org.apache.spark.SparkContext.assertNotStopped(SparkContext.scala:100)
	at org.apache.spark.SparkContext$$anonfun$parallelize$1.apply(SparkContext.scala:716)
	at org.apache.spark.SparkContext$$anonfun$parallelize$1.apply(SparkContext.scala:715)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:112)
	at org.apache.spark.SparkContext.withScope(SparkContext.scala:701)
	at org.apache.spark.SparkContext.parallelize(SparkContext.scala:715)
	at org.apache.spark.SparkContext$$anonfun$makeRDD$1.apply(SparkContext.scala:799)
	at org.apache.spark.SparkContext$$anonfun$makeRDD$1.apply(SparkContext.scala:799)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:151)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:112)
	at org.apache.spark.SparkContext.withScope(SparkContext.scala:701)
	at org.apache.spark.SparkContext.makeRDD(SparkContext.scala:798)
	at com.sap.hcpbd.Streaming.runJob(Streaming.scala:36)
	at com.sap.hcpbd.Streaming.runJob(Streaming.scala:12)
	at spark.jobserver.SparkJobBase$class.runJob(SparkJob.scala:31)
	at com.sap.hcpbd.Streaming.runJob(Streaming.scala:12)
	at com.sap.hcpbd.Streaming.runJob(Streaming.scala:12)
	at spark.jobserver.JobManagerActor$$anonfun$getJobFuture$4.apply(JobManagerActor.scala:353)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:805)
2017-09-26 16:55:29,859 ERROR pool-4-thread-2 spark.jobserver.JobManagerActor []: Exception from job 48ee0925-063f-465e-965d-d2c961a61de3:
```

derSascha:
@bsikander thanks for the review. i tested the standalone mode only by creating a context. does your streaming job work with spark standalone in client mode?

bsikander:
@derSascha I tried your change with client mode, It worked fine. No errors.

bsikander:
@derSascha did you get a chance to have a look at this?

job-server/src/main/scala/spark/jobserver/JobManager.scala
```diff
@@ -70,28 +58,37 @@ object JobManager {
...
     def makeManagerSystem(name: String)(config: Config): ActorSystem = {
       val configWithRole = config.withValue("akka.cluster.roles",
         ConfigValueFactory.fromIterable(List("manager").asJava))
       ActorSystem(name, configWithRole)
     }
-    start(args, makeManagerSystem("JobServer")(_))
+
+    def waitForTermination(system: ActorSystem, deployMode: String) {
+      if (deployMode == "cluster") {
+        system.awaitTermination // Wait for actor system shutdown
```
bsikander: This line is causing the problem.
If I replace this line with `system.registerOnTermination(System.exit(0))` then the problem is resolved for cluster mode.

It seems that the same problem should also exist for YARN mode. Could you check ? My steps
- Start a streaming context
- Start a job
- Delete the context
- Check master UI and the driver process is still there

Now, why `system.awaitTermination` is not working? I am not fully sure about the lifecycle after cluster.down() is called from AkkaClusterSupervisorActor. It seems that some operations are still being done using SparkContext but since normally we execute `System.exit(0)`, we don't see the error.

job-server/src/main/scala/spark/jobserver/JobManager.scala
```diff
@@ -15,51 +15,39 @@ import scala.collection.JavaConverters._
...
-        sys.exit(1)
-      }
-      ConfigFactory.parseFile(configFile).withFallback(defaultConfig)
+    val systemConfig = ConfigFactory.parseFile(systemConfigFile).withFallback(defaultConfig)
+    val config = if (deployMode == "cluster") {
+      logger.info("Cluster mode: Removing akka.remote.netty.tcp.hostname from config!")
+      logger.info("Cluster mode: Replacing spark.jobserver.sqldao.rootdir with container tmp dir.")
+      val sqlDaoDir = Files.createTempDirectory("sqldao")
+      val sqlDaoDirConfig = ConfigValueFactory.fromAnyRef(sqlDaoDir.toAbsolutePath.toString)
+      systemConfig.withoutPath("akka.remote.netty.tcp.hostname")
```
bsikander: Even though context creatio and job starting works but shouldn't we replace the hostname with the IP of current VM? In the logs of FrontEnd VM, we see `akka.tcp://JobServer@127.0.0.1:45012` which is misleading because instead of localhost it should be the IP of one of the slaves. Currently, we store the reference of slave in contextInitInfos hashmap in AkkaClusterSupervisorActor and using this reference we can communicate with the JobManagerActor but this hashmap is kindof temporary and will probably go away as we starting moving towards HA. @velvia @derSascha

job-server/src/main/scala/spark/jobserver/JobManager.scala
bsikander: @velvia Could you think of a reason why using `system.awaitTermination` leads to exception posted in my previous comment ?

job-server/src/main/scala/spark/jobserver/JobManager.scala
derSascha: @bsikander You have to set `akka.remote.netty.tcp.hostname` in your config to the cluster side IP of your host running the web frontend. The code, mention above, runs inside the job manager (application master / context java vm). This process runs on another host than the web frontend in cluster setup and that's why hostname settings gets removed here.  Did you set `akka.remote.netty.tcp.hostname` in your config?

job-server/src/main/scala/spark/jobserver/JobManager.scala
derSascha: @bsikander Thanks for pointing this out. I tested the YARN cluster mode with a batch job and the old behavior (`system.registerOnTermination(System.exit(0))`) results in a immediately finished context. It looks like that the finished main method also finished the YARN job. The `system.awaitTermination` avoids this and keeps the method running until the context has finished. Sound like we have to change this to something like `deployMode == "cluster" && master == "yarn"` right?

derSascha:
@bsikander small update: the `system.awaitTermination` now runs only in YARN cluster mode

job-server/src/main/scala/spark/jobserver/JobManager.scala
bsikander: Yes, I did set `akka.remote.netty.tcp.hostname` and I also understand that on the JobMangerActor side we remove this setting.

Here is the flow:
1- User sends POST /context to create a context
2- manager_start.sh is used to spawn a JVM which runs JobManager.
3- In cluster mode, JobManager loads the config file already placed in the Slave VM.
4- This code removes the `akka.remote.netty.hcp.hostname`.
5- Now, when JobManagerActor communicates back with web frontend the Akka address which web front receives is of the format `akka:tcp://JobServer@127.0.0.1:<port>`. This part is a little odd. Since this is a remote actor the address should be something like `akka:tcp://JobServer@10.250.x.x:<port>`. That is why I was asking that is this intentional and is the behavior correct?

Also, is is possible that instead of removing the hostname we replace the hostname with the current IP?

job-server/src/main/scala/spark/jobserver/JobManager.scala
bsikander: I do not have much experience in YARN so cannot say but if this check fixes the problem then we should do it.

+1 for the update.

bsikander:
+1. I was able to test cluster mode and now it behaves as expected.

@derSascha Avoid force pushing :P I had a hard time figuring out that you removed a parameter from spark-submit in manager_start.sh :P

@velvia @noorul Could we go ahead and merge this change?

msohn:
+1, please merge this change

job-server/src/main/scala/spark/jobserver/JobManager.scala
bsikander: @derSascha any opinion?

bsikander:
@derSascha Could you resolve the conflict?

bin/manager_start.sh
velvia: I mean the entire ActorPath that has something like /user/$contextActorName  ....  this matters if we were to put another management actor layer in front of our actors, then the path to look them up might need to be fixed.  If you pass the entire actorPath then it could be used to look up actors remotely even if the hiearchy changes

job-server-api/src/main/scala/spark/jobserver/util/SparkJobUtils.scala
```diff
@@ -82,11 +82,6 @@ object SparkJobUtils {
     // Set the Jetty port to 0 to find a random port
     conf.set("spark.ui.port", "0")
 
-    // Set spark broadcast factory in yarn-client mode
-    if (sparkMaster == "yarn-client") {
```
velvia: Just curious... is this not needed anymore?

job-server/src/main/resources/application.conf
```diff
@@ -1,6 +1,10 @@
 # Settings for safe local mode development
 spark {
+  # local[...], yarn or mesos://...
   master = "local[4]"
+  # client or cluster deployment
```
velvia: (nit) please add a blank line to separate configs and comments

job-server/src/main/resources/application.conf
```diff
@@ -80,6 +78,9 @@ spark {
       }
     }
 
+    # Start embeded H2 Server (usefull in cluster deployment)
```
velvia: (nit) embedded (spelling)

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -183,15 +185,15 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
       cluster.down(actorRef.path.address)
   }
 
-  private def initContext(actorName: String, ref: ActorRef, timeoutSecs: Long = 1)
+  private def initContext(contextConfig: Config, actorName: String, ref: ActorRef, timeoutSecs: Long = 1)
```
velvia: Maybe we should have one param per line for better readability

job-server/src/main/scala/spark/jobserver/JobManager.scala
velvia: Hmm.  My opinion is that 127.0.0.1 would not work in a real cluster, but what is everyone's testing results?  OTOH if you leave out the hostname Akka tries to get the hostname, so the question is what would u put there instead of what Akka guesses?

job-server/src/main/scala/spark/jobserver/JobManager.scala
```diff
@@ -70,28 +62,37 @@ object JobManager {
...
+
     def makeManagerSystem(name: String)(config: Config): ActorSystem = {
       val configWithRole = config.withValue("akka.cluster.roles",
         ConfigValueFactory.fromIterable(List("manager").asJava))
       ActorSystem(name, configWithRole)
     }
-    start(args, makeManagerSystem("JobServer")(_))
+
+    def waitForTermination(system: ActorSystem, master: String, deployMode: String) {
+      if (master == "yarn" && deployMode == "cluster") {
```
velvia: Nice

job-server/src/main/scala/spark/jobserver/JobServer.scala
```diff
@@ -48,49 +51,75 @@ object JobServer {
...
-        sys.exit(1)
     }
+
+    // cluster mode requires network base H2 server
+    if (driverMode == "cluster" && jobDaoClass.getName == "spark.jobserver.io.JobSqlDAO") {
+      val jdbcUrl = config.getString("spark.jobserver.sqldao.jdbc.url")
+        if (jdbcUrl.startsWith("jdbc:h2") && !jdbcUrl.startsWith("jdbc:h2:tcp")
+            && !jdbcUrl.startsWith("jdbc:h2:ssl")) {
+          throw new InvalidConfiguration(
+            """H2 backend and cluster mode is not support with file or in-memory storage,
```
velvia: (minor nit) not supported

job-server/src/main/scala/spark/jobserver/JobServer.scala
```diff
@@ -48,49 +51,75 @@ object JobServer {
...
+               use tcp or ssl server.""")
+        }
+    }
+
+    // start embedded H2 server
+    if (config.getBoolean("spark.jobserver.startH2Server")) {
```
velvia: Dumb question.  Doesn't the startH2Server setting conflict potentially with the sqlDAO setting?  IE it seems for yarn or cluster mode I need to point to a server H2, and it is checked above, but then here we can start an embedded one, so if I use the embedded option what do I set the sqldao setting to and jdbc path?

velvia:
@derSascha of course go ahead and resolve the conflict, maybe address the small nits and misspellings, then I'm happy to merge

bin/manager_start.sh
derSascha: I don't know how to start the actor in the `JobManger` with a given path... Due to the fact that this has nothing to do with cluster mode and providing only the actor name is the same behavior like before, let us address this in another issue/pull request.

job-server-api/src/main/scala/spark/jobserver/util/SparkJobUtils.scala
derSascha: The torrent based broadcast factory seems to be the default since spark 1.1.0 and the option was entirely removed in spark 2.0.0.

job-server/src/main/scala/spark/jobserver/JobServer.scala
derSascha: See the cluster docs added. You have to set the jdbc setting to something like `jdbc:h2:tcp://CLUSTER-IP:9092/h2-db;AUTO_RECONNECT=TRUE` to connect to the embedded H2 server.

job-server/src/main/scala/spark/jobserver/JobManager.scala
derSascha: @bsikander i can't reproduce this. Are you sure that you are not looking in some older logs? All my logs have something like `akka.tcp://JobServer@10.250.x.x`.

The basic idea is: Remove the hostname, set the cluster to join to `10.250.x.x` and now let akka choose the right network interface to connect to `10.250.x.x`.

bin/manager_start.sh
```diff
@@ -27,26 +23,46 @@ JAVA_OPTS="-XX:MaxDirectMemorySize=$MAX_DIRECT_MEMORY
...
-MESOS_OPTS=""
-if [ $1 == "mesos-cluster" ]; then
-    MESOS_OPTS="--master $MESOS_SPARK_DISPATCHER --deploy-mode cluster"
-    appdir=$REMOTE_JOBSERVER_DIR
-fi
+# copy files via spark-submit and read them from current (container) dir
+if [ $2 = "cluster" -a -z "$REMOTE_JOBSERVER_DIR" ]; then
+  SPARK_SUBMIT_OPTIONS="$SPARK_SUBMIT_OPTIONS
+    --master $1 --deploy-mode cluster
+    --conf spark.yarn.submit.waitAppCompletion=false
```
derSascha: The default is `waitAppCompletion=true` and means that the client write the current application status (running) every second into the log. Setting this to `false` mean, create the app and disconnect after the request has been accepted.

bin/manager_start.sh
```diff
@@ -27,26 +23,46 @@ JAVA_OPTS="-XX:MaxDirectMemorySize=$MAX_DIRECT_MEMORY
...
 fi
 
-eval $cmd
+cmd='$SPARK_HOME/bin/spark-submit --class $MAIN --driver-memory $JOBSERVER_MEMORY
+      --conf "spark.executor.extraJavaOptions=$LOGGING_OPTS"
+      $SPARK_SUBMIT_OPTIONS
+      --driver-java-options "$GC_OPTS $JAVA_OPTS $LOGGING_OPTS $CONFIG_OVERRIDES $SPARK_SUBMIT_JAVA_OPTIONS"
+      $JAR_FILE $3 $4 $CONF_FILE'
+
+eval $cmd 2>&1 > $5/spark-job-server.out
```
derSascha: The yarn cluster mode client does not read the log4j properties and logs to stdout... All errors before this line are going into the global `spark-job-server.log` and all output of spark-submit  into its context specific log.

derSascha:
Small update:
1. All changes suggested by @velvia implemented
2. `spark.yarn.submit.waitAppCompletion=false`added (don't log every second that the app is running)
3. redirect spark-submit ouput into its context specific log
4. some more notes about the `waitForTermination` stuff added:
YARN cluster mode: We need `system.awaitTermination` but should never call `System.exit(0)`.
Standalone cluster mode: We need to call `System.exit(0)` to kill the driver on context stop.

@bsikander the `manger_start.sh` has been changed again

velvia:
@derSascha so I can't tell from the last commits where the fixes are, they look like the same commits as before.  However they were just really minor comments, so I will go ahead and merge it.

derSascha:
@velvia the fixes are merged into the `feat(deploy): context via akka and cluster mode` commit. Thanks for merging this

job-server/src/main/scala/spark/jobserver/JobManager.scala
bsikander: Yea, I see 127.0.0.1. Have you modified your /etc/hosts by any chance ?

I will try your new change today and will let you know.

job-server/src/main/scala/spark/jobserver/JobManager.scala
```diff
@@ -12,54 +12,46 @@ import org.slf4j.LoggerFactory
...
 import scala.collection.JavaConverters._
+import scala.util.Try
 
 /**
  * The JobManager is the main entry point for the forked JVM process running an individual
- * SparkContext.  It is passed $workDir $clusterAddr $configFile
- *
- * Each forked process has a working directory with log files for that context only, plus
- * a file "context.conf" which contains context-specific settings.
+ * SparkContext.  It is passed $clusterAddr $actorName $systemConfigFile
```
LeightonWong: Thanks for your great work. I am not sure if this $systemConfigFile is still necessary now after transfer configuration using akka message. Because In mesos cluster mode, the $CONF_FILE in manager_start.sh is set to be a remote address, but In JobManager we still read it like a normal local file using `val systemConfigFile = new File(args(2))`

PR #886, opened 2017-07-11
fix(job-server): Fix/partial fix for issue #876 - Performance (of) SqlDAO in SJS 0.7

Meehau:
* Separate the BINARIES table into BINARIES and BINARIES_CONTENTS, so that only the BINARIES_CONTENTS table contains the big "BINARY" (byte array) column. Both tables can be joined on the same BIN_ID (and sometimes are).
* Optimize the JobDAO.getLastUploadTimeAndType(...) method - made it abstract and implemented an optimized version in the JobFileDAO and JobSqlDAO.

**Pull Request checklist**
- [x] The commit(s) message(s) follows the contribution [guidelines](doc/contribution-guidelines.md#commit-message-format) ?
- [x] Tests for the changes have been added (for bug fixes / features) ?
- [x] Docs have been added / updated (for bug fixes / features) ?

**Current behavior :**
SqlDAO could use some performance improvements. See issue #876 

**New behavior :**
See commit comments above.

**BREAKING CHANGES**
None.

**Other information**:
None.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/886)

Issue #876, opened 2017-06-26
Performance SqlDAO in SJS 0.7


thoHeinze:
Hi all, 

with the recent release 0.7 we were forced to switch from the FileDAO to SQLDAO. Since than we see major performance problems. We investigated already the root cause of these problems and are willing to contribute patches, but we like first to understand the current architecture a bit better. 

Our setup:
* we use Spark Jobserver with a H2 database running in server mode on the same VM and use context-per-jvm = true
* our Jobserver constantly runs 8 to 10 Spark jobs concurrently, which are updated frequently
* our users push jars larger than >30 MB and given that they don't always delete their jars afterwards we have several douzen up to hundreds of JARs existing on the jobserver; the H2 database will grow to several gigabyte of memory
* a simple "Select count(*) from BINARIES" will take 10 to 15 seconds for this setup
* most calls from the jobserver itself to the database fail due to the hard-coded timeouts inside the current code

Questions:
**BINARY vs. BLOB**: Currently you use BINARY as column type for storing the JAR inside H2. All BINARY data will be stored in Memory and is not thought for storing files. Why are you not using BLOB as recommended by H2 for storing files ? 
**Separate Metadata**: The BINARIES table is not only used for accessing the jars itself, but also for getting information about the jobs. Why don't you seperate the raw file content into a separate table and just have a hash inside the BINARIES table ?
**Streaming vs. loading data into ByteArray**: Is there any reason you always need to load the whole file into a ByteArray before transfering it? You could just use a streaming API for it.
**FileCacher**: What is the purpose of the FileCacher? In our setup with Jobserver and DB on the same host, it only results in a redundant storage of all jars. Imho this should be an optional feature, which is turned of by default (it only make sense if the DB runs on a separate host).

msohn:
The H2 documentation says the following about the BINARY type [1]:

"Represents a byte array. For very long arrays, use BLOB. The maximum size is 2 GB, but **the whole object is kept in memory when using this data type**. The precision is a size constraint; only the actual data is persisted. For large text data BLOB or CLOB should be used."

I did some experiments and found that as soon as the total size of all binaries stored in the BINARIES table grows towards max heap size of the H2 process even simple SQL queries not explicitly selecting the BINARY column (e.g. "select BIN_ID from BINARIES") is slowing down heavily and finally crashes H2 probably due to an OOM.

[...]

PR #932, opened 2017-09-25
fix(job-server): JobSQLDAO: Move blobs to a separate table (#876)

msohn:
* Separate the BINARIES table into BINARIES and BINARIES_CONTENTS, so
that only the BINARIES_CONTENTS table contains the "BINARY" (Blob)
column. Both tables can be joined on the same BIN_ID.

* Optimize the JobDAO.getLastUploadTimeAndType(...) method - make it
abstract and implement an optimized version in the JobFileDAO and
JobSqlDAO.

This is the result of rebasing pull request #886 onto current master and
reimplementing the migration in JDBC migration since blob columns cannot
be migrated using a SQL script.

**Pull Request checklist**

- [ ] Docs have been added / updated (for bug fixes / features) ? not needed

**Current behavior :** 
Uploaded jars are stored in blob column "BINARY" in table "BINARIES".

**New behavior :**
Uploaded jars are stored in blob column "BINARY" in separate table "BINARIES_CONTENTS" which only contains BIN_ID (primary key) and the blob column.

**BREAKING CHANGES**
No breaking changes

**Other information**:
Existing blobs are migrated from BINARIES to BINARIES_CONTENTS

### DIFF
--- a/job-server/src/main/scala/db/h2/migration/V0_7_4/V0_7_4__Migrate_Blobs.scala
+++ b/job-server/src/main/scala/db/h2/migration/V0_7_4/V0_7_4__Migrate_Blobs.scala
@@ -1,0 +1,63 @@
+package db.h2.migration.V0_7_4
+
+import java.sql.Connection
+
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.TimeoutException
+import scala.concurrent.duration.DurationInt
+
+import org.flywaydb.core.api.migration.jdbc.JdbcMigration
+import org.slf4j.LoggerFactory
+
+import javax.sql.rowset.serial.SerialBlob
+import slick.dbio.DBIO
+import slick.driver.H2Driver.api.actionBasedSQLInterpolation
+import slick.jdbc.GetResult
+import slick.jdbc.PositionedParameters
+import slick.jdbc.SetParameter
+import spark.jobserver.slick.unmanaged.UnmanagedDatabase
+import java.sql.Blob
+
+class V0_7_4__Migrate_Blobs extends JdbcMigration {
+  private val Timeout = 10 minutes
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  private case class BinaryContent(id: Int, binary: Blob)
+
+  private def logErrors = PartialFunction[Throwable, Unit] {
+    case e: Throwable => logger.error(e.getMessage, e)
+  }
+
+  private def insertBlob(db: UnmanagedDatabase, b: BinaryContent): Unit = {
+    implicit object SetSerialBlob extends SetParameter[SerialBlob] {
+      def apply(v: SerialBlob, pp: PositionedParameters) {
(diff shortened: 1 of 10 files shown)


msohn:
I added empty line at the end of the files and added the missing try catch rollback in case something goes wrong while running migration

job-server/src/main/scala/db/h2/migration/V0_7_4/V0_7_4__Migrate_Blobs.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import javax.sql.rowset.serial.SerialBlob
+import slick.dbio.DBIO
+import slick.driver.H2Driver.api.actionBasedSQLInterpolation
+import slick.jdbc.GetResult
+import slick.jdbc.PositionedParameters
+import slick.jdbc.SetParameter
+import spark.jobserver.slick.unmanaged.UnmanagedDatabase
+import java.sql.Blob
+import scala.util.control.NonFatal
```
velvia: minor nit: please group the scala imports together, in the second section (lines 5-8)

job-server/src/main/scala/db/postgresql/migration/V0_7_4/V0_7_4__Migrate_Blobs.scala
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  private def insertBlob(db: UnmanagedDatabase, b: BinaryContent): Unit = {
+    implicit object SetSerialBlob extends SetParameter[SerialBlob] {
+      def apply(v: SerialBlob, pp: PositionedParameters) {
+        pp.setBlob(v)
+      }
+    }
+    val blob = new SerialBlob(b.binary.getBytes(1, b.binary.length().toInt))
+    val insertBlob = sqlu"""INSERT INTO "BINARIES_CONTENTS" ("BIN_ID", "BINARY") VALUES (${b.id}, ${blob})"""
```
velvia: I wish there was a way to unify these different migrations for different DBs together.  It seems most of the code is really identical, other than the SQL statements are a tiny bit different.

msohn:
we also need to handle deletion of BINARIES_CONTENTS when the corresponding BINARIES record is deleted

msohn:
* rebased onto origin/master
* fixed missing quoting in v0_7_4 migration SQL statements which crashed this migration on postgres
* reordered imports
* factored out database independent migration common code for migration v0_7_5 into trait "Migration"
* fixed deletion of jars which needs to take into account that content is now stored in table "BINARIES_CONTENTS"

msohn:
please consider to merge this PR soon, I'd like to avoid a third rebase onto yet another db migration

noorul:
@msohn Merged! Thank you!

msohn:
Katja will soon push another PR to avoid storage of duplicate identical blobs in the BINARIES_CONTENTS table which currently can be created if the same jar is uploaded multiple times

velvia:
Sounds great, thanks

PR #950, opened 2017-10-24
feat(jobserver): Enable/disable jar caching on upload

ecandreev:
Currently as soon as the jar is upload using /binaries
endpoint, it is cached on filesystem. On a high usage
SJS, disk is important and a trend has been seen where
users upload jars but don't start jobs. This change
gives the option to admin to cache jars on job start
only.

The property "spark.jobserver.cache-on-upload" is
configurable and can have the value true/false.

**Pull Request checklist**

**Current behavior :** As soon as the jar is upload using /binaries
endpoint, it is cached on filesystem.

**New behavior :** If the property "spark.jobserver.cache-on-upload" is set to false, jar will be only cached on job start.

**BREAKING CHANGES**

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/950)

### DIFF
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -31,7 +31,7 @@ spark {
 
     # Cache binary on upload.
     # If set to false, the binary will be cached on job start only.
-    cache-on-upload = true;
+    cache-on-upload = true
 
     filedao {
       rootdir = /tmp/spark-jobserver/filedao/data
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -29,6 +29,10 @@ spark {
     #   test = /path/to/my/test.jar
     # }
 
+    # Cache binary on upload.
+    # If set to false, the binary will be cached on job start only.
+    cache-on-upload = true;
+
     filedao {
       rootdir = /tmp/spark-jobserver/filedao/data
     }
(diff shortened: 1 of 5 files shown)


bsikander:
+1. Looks good. Can somebody please merge this?

job-server/src/main/scala/spark/jobserver/io/JobSqlDAO.scala
```diff
@@ -137,8 +138,11 @@ class JobSqlDAO(config: Config) extends JobDAO with FileCacher {
                           binaryType: BinaryType,
                           uploadTime: DateTime,
                           binBytes: Array[Byte]) {
-    // The order is important. Save the jar file first and then log it into database.
-    cacheBinary(appName, binaryType, uploadTime, binBytes)
+    val cacheOnUploadEnabled = Try(config.getBoolean("spark.jobserver.cache-on-upload")).getOrElse(true)
```
noorul: If you have default value in application.conf then you don't need to use Try

ecandreev: Done
Created default value in application.conf and removed Try from JobSqlDAO

job-server/src/main/resources/application.conf
```diff
@@ -29,6 +29,10 @@ spark {
     #   test = /path/to/my/test.jar
     # }
 
+    # Cache binary on upload.
+    # If set to false, the binary will be cached on job start only.
+    cache-on-upload = true
```
zeitos: I think this should be set to false for backwards compatibility

zeitos: my mistake missread

zeitos:
Hi @ecandreev great idea! Can you add a test case for this?

noorul:
@ecandreev Thank you! Merged!

PR #960, opened 2017-11-06
fix(job-server): JobSQLDAO: Prevent saving the same binary multiple times

ecandreev:
Save hash of the binary in extra column in BINARIES and use it as a primary key in BINARIES_CONTENTS table instead of binary id.
For hash calculation SHA-256 is used.

**Pull Request checklist**

- [* ] The commit(s) message(s) follows the contribution [guidelines](doc/contribution-guidelines.md#commit-message-format) ?
- [* ] Tests for the changes have been added (for bug fixes / features) ?

**Current behavior :** 
Every time by loading a binary its content is also saved in db, even if binary with the same content was already saved before.

**New behavior :** The content of a binary is only saved in db if binary content with the same hash was not there before. Otherwise BIN_HASH in BINARIES table will point to existing BINARIES_CONTENTS entry.

**BREAKING CHANGES**

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/960)


noorul:
LGTM

Issue #1040, opened 2018-03-05
Restart of Jobserver


bsikander:
Currently, if we restart the jobserver, it loses all the connections to the context JVMs and there is no way to connect to them. We believe that with the following changes in jobserver, we can achieve basic restart of Jobserver and jobs.
 
1- Persistence of Context in DB
2- Removal of in-memory data structures from master
3- Reconnection to context JVMs if Jobserver restarts (https://github.com/spark-jobserver/spark-jobserver/issues/353)
4- Restart of context JVMs through supervise mode
 
Since these changes will be done in multiple steps, this issue is an initial announcement of upcoming changes.

**Used Spark version**
2.2.1

zenglian:
Pls provide a rest API for restarting a job by the way.

thoHeinze:
@zenglian  What is your scenario? We will enable the Spark supervise mode, there will be no need  to manually restart the job.

thoHeinze:
@bsikander can you please update this issue? Most of the work is done imho.

bsikander:
@thoHeinze yes, all the tasks are done.

PR #1066, opened 2018-05-07
feat(jobserver): Introduce supervise mode

bsikander:
Note: This change is a combination of 6 related commits. Each commit has its own explanation.
This change was only tested with spark-standalone.

This change allows contexts and jobs within these contexts
to get restarted if they get stopped unexpectedly. This
is achieved using the supervise mode provided by Spark.

This change introduces a global property
spark.driver.supervise. The value can be set to true or false.
True = Enable supervise mode by default for all the new contexts.
False = Disable supervise mode for all contexts but user can
decide if the context needs to be supervised or not.

For each context, a property launcher.spark.driver.supervise is
provided, user can set individually for each context, if it
should use this mode or not.

Here are all the possible cases:

Default mode  |  context mode  |  Action
true             true             enabled
true             false            disabled
false            true             enabled
false            false            disabled
true             Not defined      enabled
false            Not defined      disabled

Curently, this mode is tested with streaming/batch jobs.
A few things to note
* The criteria for a restart:
    * Job is either in RUNNING state, or
    * Job is in ERROR state with Context Terminated
        exception as error message.
    * The job is async
* The jobs for restart are selected based on unique context Id. So, we won't get jobs from any other context.
* After the restart, we get the contextId from the managerName provided by slave launch command.
* If a batch context has multiple jobs, then all the jobs will be restarted
* A context without any successfully restarted job will be deleted automatically. If there is at least one job, which doesn't fail, the context won't be deleted.
* If a context fails during the restart, this job will be marked as failed in the db.
* Since currently we have 1 master, to avoid overloading the master with all the restart scenario, the job restart logic is implemented on slaves. Only the context initialization logic is on master
* On master we detect a supervise scenario based on contextInitInfos hashmap. If the hashmap has an entry, it is a normal initialize of context but if it doesn't then it is a restart scenario. This apporach is safe because contextInitInfos's key is the slaves job manager actor name, this name is a combination of jobManager-<uuid>. If the context was initialized properly then this entry will be removed for sure. If the initialization failed, then contextInitInfos might have the entry but it is useless because context was never initialized and restart is not possible.
* If an existing job is restarted successfully, we change the error status back to running and also purge the endtime.

Note: If supervise mode is active, JVMs will only be restarted if the
JVM was killed with a non-zero exit code.

Some behaviors of cluster + supervise mode:
1- If we kill the application from MasterUI -> kills the app + driver and no restart
2- If we kill the driver from Master UI -> kills app + driver and no restart
3- If we kill the context from SJS -> Kills the driver + app and no restart
4- If we kill the job from SJS -> Keep the driver + app running and no restart of the job
5- If we kill the Driver Wrapper JVM using kill -9 -> Driver comes up with different PID but same driver ID, application ID changes, restart happened.
6- If we kill the Executor JVM using kill -9 -> Drivers stays, app also stays, the executor JVM is relauched with different pid, not related to Restart, normal behavior of Spark
7 - If we have a job which fails on the Executor and the executor restarts -> We can delete the job through Master UI, killing it through SJS makes it a zombie.
8- If we have a job which fails on the Driver -> Driver restarts (only if supervise mode is enabled), we can stop it through SJS, we can always stop it from UI.

**Pull Request checklist**

Supervise mode is not supported

**New behavior :**
Supervise mode is now supported

**BREAKING CHANGES**

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1066)

### DIFF
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -5,6 +5,8 @@ spark {
 
   # client or cluster deployment
   submit.deployMode = "client"
+
+  # For more details see https://github.com/spark-jobserver/spark-jobserver/tree/master/doc/supervise-mode.md
   driver.supervise = false
 
   # spark web UI port
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -5,6 +5,7 @@ spark {
 
   # client or cluster deployment
   submit.deployMode = "client"
+  driver.supervise = false
 
   # spark web UI port
   webUrlPort = 8080
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -23,6 +23,8 @@ spark {
     kill-context-on-supervisor-down = false
     manager-initialization-timeout = 40s
 
+    ignore-akka-hostname = true
+
     # Note: JobFileDAO is deprecated from v0.7.0 because of issues in
     # production and will be removed in future, now defaults to H2 file.
     jobdao = spark.jobserver.io.JobSqlDAO
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -21,7 +21,7 @@ import scala.concurrent.Await
 import akka.pattern.gracefulStop
 import org.joda.time.DateTime
 import org.slf4j.LoggerFactory
-import spark.jobserver.JobManagerActor.{GetContexData, ContexData, SparkContextDead}
+import spark.jobserver.JobManagerActor.{GetContexData, ContexData, SparkContextDead, RestartExistingJobs}
 import spark.jobserver.io.{JobDAOActor, ContextInfo, ContextStatus}
 import spark.jobserver.util.{InternalServerErrorException, NoCallbackFoundException}
 
@@ -124,20 +124,31 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
           (contextFromDAO, callbacks) match {
             case (Some(JobDAOActor.ContextResponse(Some(contextInfo))),
                 Some((successCallback, failureCallback))) =>
-              val contextConfig = ConfigFactory.parseString(contextInfo.config)
-              val isAdhoc = contextConfig.getBoolean("is-adhoc")
-              initContext(contextConfig, actorName,
-                     actorRef, contextInitTimeout)(isAdhoc, successCallback, failureCallback)
+              initContext(contextInfo, actorName, actorRef, false)(successCallback, failureCallback)
             case (Some(JobDAOActor.ContextResponse(None)), _) =>
               logger.error(
                  s"No such contextId ${contextId} was found in DB. Cannot initialize actor ${actorName}")
               actorRef ! PoisonPill
             case (Some(JobDAOActor.ContextResponse(Some(contextInfo))), None) =>
-              val exception = NoCallbackFoundException(contextInfo.id, actorRef.path.toSerializationFormat)
-              logger.error(exception.getMessage, exception)
-              actorRef ! PoisonPill // Since no watch was added, Terminated will not be fired
-              daoActor ! JobDAOActor.SaveContextInfo(contextInfo.copy(
-                  state = ContextStatus.Error, endTime = Some(DateTime.now()), error = Some(exception)))
+              isSuperviseModeEnabled(config, ConfigFactory.parseString(contextInfo.config)) match {
+                case isRestartScenario @ true =>
+                  logger.info(
+                     s"Restart request for context (${contextId}) received, probably due to supervise mode")
+                  initContext(contextInfo, actorName, actorRef, isRestartScenario)(
+                     { ref =>
+                       logger.info(s"Successfully reinitialized context (${contextId}) under supervise mode")
+                     },
+                     { ref =>
+                         logger.error(s"Failed to reinitialize context (${contextId}) under supervise mode")
+                     })
+                case false =>
+                  val exception =
+                    NoCallbackFoundException(contextInfo.id, actorRef.path.toSerializationFormat)
+                  logger.error(exception.getMessage, exception)
+                  actorRef ! PoisonPill // Since no watch was added, Terminated will not be fired
+                  daoActor ! JobDAOActor.SaveContextInfo(contextInfo.copy(
+                      state = ContextStatus.Error, endTime = Some(DateTime.now()), error = Some(exception)))
+              }
             case (None, Some((_, failureCallback))) =>
               val exception = InternalServerErrorException(contextId)
               logger.error(exception.getMessage, exception)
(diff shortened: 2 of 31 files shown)


bsikander:
This is the final piece of issue -> https://github.com/spark-jobserver/spark-jobserver/issues/1040

bsikander:
2-3 testcases are failing for supervise mode, I will have a look into them. On my setup, they run though.

job-server/src/main/resources/application.conf
```diff
@@ -5,6 +5,7 @@ spark {
 
   # client or cluster deployment
   submit.deployMode = "client"
+  driver.supervise = false
```
velvia: How about add comment so others can understand what it is... can refer to README for more details

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -124,20 +124,31 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
...
                  s"No such contextId ${contextId} was found in DB. Cannot initialize actor ${actorName}")
               actorRef ! PoisonPill
             case (Some(JobDAOActor.ContextResponse(Some(contextInfo))), None) =>
-              val exception = NoCallbackFoundException(contextInfo.id, actorRef.path.toSerializationFormat)
-              logger.error(exception.getMessage, exception)
-              actorRef ! PoisonPill // Since no watch was added, Terminated will not be fired
-              daoActor ! JobDAOActor.SaveContextInfo(contextInfo.copy(
-                  state = ContextStatus.Error, endTime = Some(DateTime.now()), error = Some(exception)))
+              isSuperviseModeEnabled(config, ConfigFactory.parseString(contextInfo.config)) match {
+                case isRestartScenario @ true =>
```
velvia: Wait, wouldn't `isRestartScenario` always be true here?

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -124,20 +124,31 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
...
+                  logger.info(
```
velvia: Is it that restart request is received right at this moment (which makes it seem like the context is about to be restarted), or do you mean that supervise mode is true, therefore restarts are enabled (which is what I read out of the code here)

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
```diff
@@ -115,10 +124,15 @@ class JobManagerActor(daoActor: ActorRef, supervisorActorAddress: String, contex
   private var contextConfig: Config = _
   private var contextName: String = _
   private var isAdHoc: Boolean = _
-  private var statusActor: ActorRef = _
+  @VisibleForTesting
```
velvia: What is this VisibleForTesting?

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
```diff
@@ -496,4 +570,89 @@ class JobManagerActor(daoActor: ActorRef, supervisorActorAddress: String, contex
...
+   daoActor ! JobDAOActor.SaveJobInfo(updatedJobInfo)
+  }
+
+  private def restartJob(existingJobInfo: JobInfo, existingJobConfig: Config) {
+    // Add response to master for testing
+    var respMsg = s"Restarting the last job (JobId=${existingJobInfo.jobId} & "
+    respMsg = respMsg + s"contextName=${existingJobInfo.contextName})"
+    logger.info(respMsg)
+    val fullJobConfig = existingJobConfig.withFallback(config).resolve()
+    val contextConfig = Try(existingJobConfig.getConfig("spark.context-settings"))
```
velvia: It seems to me a good idea to have code/methods to get context settings out.  More typesafe than having strings everywhere.

job-server/src/main/resources/application.conf
bsikander: Makes sense. I will add a Readme file with details regarding this mode.

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
bsikander: Yes, it will always be true. I just added these conditions to make the restart scenario more tight (just in case, some unknown scenario happen).

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
bsikander: Your previous comment is correct.

Just to add here, If context is started with supervise mode then on unexpectedly terminated, it will be terminated and will be restarted by Spark. As the Context JVM starts up, it will connect with the SJS HTTP JVM and will land in ActorIdentity. Here we just check that if this context is eligible for restart or not. If yes, we let it thorugh (which is almost always) and stop it other wise.

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
bsikander: Visibility of some of the variables were changed from private to protected so that we can have access to them in testcases. Normally, it is not too obvious. `@VisibleForTesting` makes it obvious that we changed the visibility just for this purpose. This is just for documentation purpose.

job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
bsikander: Good point. I will try to adapt this change.

velvia:
@bsikander I was referring to stuck jobs, like say you have a job which does not terminate when you call runJob(). I suppose spark streaming jobs might fall into this category too. So is there any way to cancel or kill such jobs?

bsikander:
Sorry for a delayed response.
I have added documentation for supervise mode according to comments by @velvia

bsikander:
@velvia 
_>> I was referring to stuck jobs, like say you have a job which does not terminate when you call runJob(). I suppose spark streaming jobs might fall into this category too. So is there any way to cancel or kill such jobs?_

Currently, SJS has no way. From Spark Master UI, we can kill.

I have seen that if the job is doing some I/O and we try to stop it, it becomes a zombie and SJS loses 
Or if a job is reading something from Kafka and I kill the Kafka or topic, the job gets stuck.
Or if alot of jobs are pending (generated by jobGenerator of Spark) and we try to stop it. It gets stuck.

Btw what is your experience with stuck jobs?

### For cluster mode:
I think we should introduce a force kill option in SJS which uses Sparks REST API (hidden though)

```
curl -X POST http://spark-cluster-ip:6066/v1/submissions/kill/<driver-id>
```

### For client mode:
I think `SparkLauncher` gives us` SparkAppHandle` which has a `kill()` method. I have not tried but it might work.

What do you think?

thoHeinze:
+1 from me. @velvia can we maybe merge this change ?

velvia:
@thoHeinze @bsikander yeah let's go ahead.

noorul:
@velvia @bsikander @thoHeinze Merged!

bsikander:
Thanks :)

PR #1077, opened 2018-06-25
feat(jobserver): Add restarting state and improve reconnect logic

pk-work:
**Pull Request checklist**

- [X] Docs have been added / updated (for bug fixes / features) ?

**Current behavior :** 
#1040
Currently, if a job is restarting and the user submits
/jobs, figuring out the restarting scenario is not too
obvious by just looking at the output.

User needs to know that if state is ERROR, error msg is
Context Terminated exception and supervise mode is
enabled for that context, only then it will be restarting.

The /contexts endpoint only shows contexts in state "RUNNING". 

If the reconnect failed, the state of the context is set to error, but the state of jobs which were running
in this context isn't touched.

**New behavior :**

This PR adds an explicit state ("RESTARTING") which is also shown in /jobs.
The /contexts endpoint will also show contexts in state "RESTARTING".
If the reconnect failed, the state of jobs which are running is set to ERROR.

**Other information**:

Known issues which will be fixed in upcoming changes:
* In Restarting state if POST /context is submitted with
the same name, then SJS will allow the creation of context
and restart of context will fail. This should be avoided.
* In Restarting state if DELETE /context is submitted,
then SJS will respond with NoSuchContext because the
Akka address in DB is not valid. It should be handled
more gracefully.
* There is a scenario in which the new restarting feature
can produce zombies:
  1. Driver gets killed and try to restart
  2. SJS Master writes state "Restarting" into database for
this Driver
  3. SJS Master crashes before Driver restarted and SJS Master
wrote state "Running" in the database for the Driver
  4. Driver comes up and becomes a zombie
If this happens, clean up on SJS Master start will not happen,
because Driver is in state "Restarting" not "Running".
Also the Driver cannot connect back to the SJS Master,
because it was restarted in the meantime.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1077)

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -478,10 +478,12 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
 
   private def getActiveContextByName(name: String): (Boolean, Option[ContextInfo]) = {
     val resp = getContextByName(name)
+    val statesInWhichContextCanBeActive =
+      List(ContextStatus.Started, ContextStatus.Running, ContextStatus.Restarting)
     resp match {
       case None => (false, None)
       case Some(JobDAOActor.ContextResponse(Some(c)))
-        if c.state == ContextStatus.Running || c.state == ContextStatus.Started => (true, Some(c))
+        if statesInWhichContextCanBeActive.contains(c.state) => (true, Some(c))
       case Some(_) => (true, None)
     }
   }
--- a/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
@@ -22,7 +22,8 @@ import akka.pattern.gracefulStop
 import org.joda.time.DateTime
 import org.slf4j.LoggerFactory
 import spark.jobserver.JobManagerActor.{GetContexData, ContexData, SparkContextDead, RestartExistingJobs}
-import spark.jobserver.io.{JobDAOActor, ContextInfo, ContextStatus}
+import spark.jobserver.JobManagerActor.ContextTerminatedException
+import spark.jobserver.io.{JobDAOActor, ContextInfo, ContextStatus, JobStatus, ErrorData}
 import spark.jobserver.util.{InternalServerErrorException, NoCallbackFoundException}
 
 object AkkaClusterSupervisorActor {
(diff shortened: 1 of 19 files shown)


pk-work:
Close-Reopened PR to rerun tests, because tests failed with:

should survive from one job to another one *** FAILED *** (4 seconds, 333 milliseconds)
[info]   java.lang.AssertionError: assertion failed: expected class spark.jobserver.CommonMessages$JobResult, found class spark.jobserver.CommonMessages$NoJobSlotsAvailable (NoJobSlotsAvailable(2))

pk-work:
Uploaded new commits to fix a type in commit: feat(jobserver): Add restarting state

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -271,25 +273,60 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
...
+        (isSuperviseModeEnabled(config, ConfigFactory.parseString(c.config)), state) match {
+            case (true, ContextStatus.Killed) =>
+              setRestartingStateForContextAndJobs(c)
+            case _ =>
+              val contextInfo = c.copy(endTime = Option(DateTime.now()), state = state)
+              daoActor ! JobDAOActor.SaveContextInfo(contextInfo)
+              daoActor ! JobDAOActor.CleanContextJobInfos(c.id, DateTime.now())
+         }
+      case Some(JobDAOActor.ContextResponse(None)) =>
+        logger.error("No context for deletion is found in the DB")
```
zeitos: nit: Do you think that if this error ever happen the reader will need more information? Maybe `contextId` or `name` ?

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -271,25 +273,60 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
...
+      case None =>
+        logger.error("Error occurred after Terminated message was recieved")
```
zeitos: and the same here as in line 297

zeitos:
Hey @pk-work thanks for the PR! I won't approve or reject the PR since I haven't been involved with SJS lately but added some comments

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
```diff
@@ -271,25 +273,60 @@ class AkkaClusterSupervisorActor(daoActor: ActorRef, dataManagerActor: ActorRef)
...
+
+  protected def handleTerminatedEvent(actorRef: ActorRef) {
+    val name: String = actorRef.path.name
+    logger.info("Actor terminated: {}", name)
+    val contextId = name.split(AkkaClusterSupervisorActor.MANAGER_ACTOR_PREFIX).apply(1)
+    val resp = getDataFromDAO[JobDAOActor.ContextResponse](JobDAOActor.GetContextInfo(contextId))
+    resp match {
+      case Some(JobDAOActor.ContextResponse(Some(c))) =>
+        val state =
+                if (c.state == ContextStatus.Stopping) ContextStatus.Finished else ContextStatus.Killed
```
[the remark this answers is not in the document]
pk-work: Hey zeitos,
I agree, importing ContextStatus._ would make the line shorter and probably increase readability (but only if you know that this is a ContextStatus).

But in this change we also import Jobstatus, which has the same states... so I guess your proposal wouldn't work, wouldn't it?

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
pk-work: Done

job-server/src/main/scala/spark/jobserver/AkkaClusterSupervisorActor.scala
pk-work: Done

bsikander:
Looks good.

PR #1134, opened 2018-11-23
feat(jobserver): Add strategies for IP resolution for drivers

bsikander:
Akka framework by default picks up the IP specified against
the hostname. It can be externally accessible IP or loopback.
In client mode, loopback works fine but in cluster mode,
drivers won't be able to communicate with Master.

For some users, configuring /etc/hosts file correctly is not
an option and current solution only provides 2 options
- Let Akka decide
- User needs to manually provide the right IP for drivers

This change refines the old solution and adds another option
to use SPAKR_LOCAL_IP env variable by Spark or automatically
pick the first available IPv4.

* Akka - This strategy leaves the resolution of IP to Akka.
  Akka will use InetAddress.getLocalHost.getHostAddress, which
  picks up the IP against your hostname as defined in /etc/hosts
  file. Generally, your hostname should point to the externally
  accessible IP. This is the best approach.
* Manual - This strategy leaves everything to user. User can put
  the correct hostname in the configuration before driver loads
  it. Jobserver will respect it. Forexample, at startup of VM,
  user can put a config in a pre-known path with right
  IP and drivers will pick it up and will work.
* Auto - This strategy is for users, which don't have much
  control on how /etc/hosts file is setup. Jobserver will look
  for SPARK_LOCAL_IP env variable or all the network interfaces
  and pick the first externally accessible IP.

Note:
* This only for cluster mode
* The default value in application.conf is set to "Akka"
* Tested only for Spark Standalone on Unix/Linux

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1134)


bsikander:
All tests passed. Code coverage is a bit low for NetworkAddressStrategy class.

bsikander:
@velvia @noorul could you please review this change?

noorul:
LGTM

Issue #1161, opened 2019-02-25
The job not finished with an error for `java.lang.OutOfMemoryError: Java heap space`


fengxucn:
**Used Spark version**
2.2.0

**Used Spark Job Server version**
(Released version, git branch or docker image version)
0.8.0

**Deployed mode**
(client/cluster on Spark Standalone/YARN/Mesos/EMR or default)
YARN Client

**Actual (wrong) behavior**
The job should finished with an error

**What I did in my code**
`sparkSession.sparkContext.broadcast(castarData.map(row => row.getAs[String]("id")).distinct().collect.toSet)`

**Logs**

```
2019-02-25 20:44:36 ERROR JobManagerActor:606 - Got Throwable
java.lang.OutOfMemoryError: Java heap space
        at scala.collection.mutable.ResizableArray$class.ensureSize(ResizableArray.scala:103)
        at scala.collection.mutable.ArrayBuffer.ensureSize(ArrayBuffer.scala:48)
        at scala.collection.mutable.ArrayBuffer.$plus$eq(ArrayBuffer.scala:84)
        at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeCollect$1$$anonfun$apply$2.apply(SparkPlan.scala:279)
        at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeCollect$1$$anonfun$apply$2.apply(SparkPlan.scala:279)
        at scala.collection.Iterator$class.foreach(Iterator.scala:893)
        at org.apache.spark.sql.execution.SparkPlan$$anon$1.foreach(SparkPlan.scala:257)
        at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeCollect$1.apply(SparkPlan.scala:279)
        at org.apache.spark.sql.execution.SparkPlan$$anonfun$executeCollect$1.apply(SparkPlan.scala:278)
        at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
        at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:186)
        at org.apache.spark.sql.execution.SparkPlan.executeCollect(SparkPlan.scala:278)
        at org.apache.spark.sql.Dataset.org$apache$spark$sql$Dataset$$collectFromPlan(Dataset.scala:2854)
        at org.apache.spark.sql.Dataset$$anonfun$collect$1.apply(Dataset.scala:2391)
[...]

PR #1190, opened 2019-04-23
fix(jobserver): ensure job cleanup after fatal error - ref #1161

sgautrin:
`Future` only completes on success or when encountering a `NonFatal`
error; when encountering a fatal error, the future does not complete at
all and the error is passed to the execution context.

This change wraps all `Throwable` so the `Future` can complete and the
`Failure` handlers can cleanup the job and properly mark it as failed.

**Pull Request checklist**

#1161 

**New behavior :**

When a fatal error occurs in a job, the `JobManagerActor` is now able to properly mark it as failed and cleanup.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1190)

### DIFF
--- a/job-server-tests/src/main/scala/spark/jobserver/SparkTestJobs.scala
+++ b/job-server-tests/src/main/scala/spark/jobserver/SparkTestJobs.scala
@@ -15,6 +15,13 @@ class MyErrorJob extends SparkTestJob {
   }
 }
 
+/** @see [[scala.util.control.NonFatal]] */
+class MyFatalErrorJob extends SparkTestJob {
+  def runJob(sc: SparkContext, config: Config): Any = {
+    throw new OutOfMemoryError("this is a fatal error")
+  }
+}
+
 class ConfigCheckerJob extends SparkTestJob {
   import scala.collection.JavaConverters._
 
--- a/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobManagerActor.scala
@@ -600,11 +600,13 @@ class JobManagerActor(daoActor: ActorRef, supervisorActorAddress: String, contex
         case e: java.lang.AbstractMethodError => {
           logger.error("Oops, there's an AbstractMethodError... maybe you compiled " +
             "your code with an older version of SJS? here's the exception:", e)
-          throw e
+          // wrap so it can complete as Failure even if not a scala.util.control.NonFatal
+          throw new RuntimeException(e)
         }
         case e: Throwable => {
           logger.error("Got Throwable", e)
-          throw e
+          // wrap so it can complete as Failure even if not a scala.util.control.NonFatal
+          throw new RuntimeException(e)
         };
       }
     }(executionContext).andThen {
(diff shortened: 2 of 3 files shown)


bsikander:
Overall the code looks good. But I have some questions

Scala considers the following types of Errors are Fatal:
VirtualMachineError, ThreadDeath, InterruptedException, LinkageError, ControlThrowable

OOM is a subclass of `VirtualMachineError`. In case of this, will the JVM be even in a position to process the code further (in sjs `case Failure(wrapped: Throwable) =>`) ?

I doubt that JVM will be able to allocate memory for `val error = wrapped.getCause` in case of OOM.
What is your opinion on this?

I can imagine this PR to be helpful for `StackOverflowError` which is also a subclass of `VirtualMachineError` because recovery is possible.

Maybe, to be correct, instead of throwing OOM inside the sparkjob, use `StackOverflowError` to be correct.

sgautrin:
> Scala considers the following types of Errors are Fatal:
> VirtualMachineError, ThreadDeath, InterruptedException, LinkageError, ControlThrowable
> 
> OOM is a subclass of `VirtualMachineError`. In case of this, will the JVM be even in a position to process the code further (in sjs `case Failure(wrapped: Throwable) =>`) ?
> 
> I doubt that JVM will be able to allocate memory for `val error = wrapped.getCause` in case of OOM.
> What is your opinion on this?

It is certainly possible that the JVM could in some cases not be able to get enough memory even for that, but then it would be completely crashed. However, OOM can get thrown when there is actually plenty of memory left, as such an exception is thrown when the required memory allocation is not possible; few cases where an OOM would happen when there is still enough memory to allocate an Exception object:

* request for 1MB memory when there's only 500k available on the heap
* most memory allocations request (e.g. byte buffer) contiguous space, so if you'd request a 256MB byte buffer, you could get an OOM while still having say 2G free heap space when the heap is fragmented enough that there is no 256M contiguous free space available.

> I can imagine this PR to be helpful for `StackOverflowError` which is also a subclass of `VirtualMachineError` because recovery is possible.
> 
> Maybe, to be correct, instead of throwing OOM inside the sparkjob, use `StackOverflowError` to be correct.

The JVM running the job is the one throwing the OOM, not the job itself. Obviously though the job itself can catch the OOM and wrap it itself, but then that's exactly the same as doing it in the job server, except every job has to implement that.

Wrapping OOM: heap space errors do work though (at the job level but it should be no different there), at least for the situations encountered in the original issue #1161 and by myself). It is indeed possible that the error occurs because the heap is actually completely so full that the error cannot be wrapped, but then I don't know what would happen before or after the fix: in both cases I guess there would be random OOM elsewhere in SJS, and with the fix I guess at worse either there would be an OOM thrown trying to wrap the OOM leading us back to square one (OOM would be handled by the execution context and the job not cleaned up), or there would be an OOM thrown while handling the cleanup, in which case the OOM would this time affect the JobActor, with a chance at recovery by akka.

bsikander:
>> However, OOM can get thrown when there is actually plenty of memory left, as such an exception is thrown when the required memory allocation is not possible;

Valid point. I agree.

>> Maybe, to be correct, instead of throwing OOM inside the sparkjob

Here i meant the spark job inside your test :P . But leave it since your first point it valid.

bsikander:
Thank you. Merging.

Issue #1149, opened 2019-01-24
Identify the state of Spark-JobServer actors


SrivigneshM:
**Spark: 2.2.4**

**SJS: v0.6.2**

**Deployment: Cluster**

**Issue:**
Actors like JobInfoActor, JobManagerActor etc., gets shut down sometimes with below exception. Auto-restart of the actors does not happen in this case and it becomes dead.

Is it possible to know if all the actors are alive through an API or through some other means? If so, it might help in identifying the non-alive actors and bring back the system.

**Logs:**

```
18/08/22 08:52:00 ERROR OneForOneStrategy: [akka://JobServer/user/job-info] - Futures timed out after [3 seconds]
java.util.concurrent.TimeoutException: Futures timed out after [3 seconds]
at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:223)
at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:190)
at akka.dispatch.MonitorableThreadFactory$AkkaForkJoinWorkerThread$$anon$3.block(ThreadPoolBuilder.scala:167)
at scala.concurrent.forkjoin.ForkJoinPool.managedBlock(ForkJoinPool.java:3640)
at akka.dispatch.MonitorableThreadFactory$AkkaForkJoinWorkerThread.blockOn(ThreadPoolBuilder.scala:165)
at scala.concurrent.Await$.result(package.scala:190)
at spark.jobserver.JobInfoActor$$anonfun$wrappedReceive$1$$anonfun$applyOrElse$1.apply$mcV$sp(JobInfoActor.scala:64)
at scala.util.control.Breaks.breakable(Breaks.scala:38)
at spark.jobserver.JobInfoActor$$anonfun$wrappedReceive$1.applyOrElse(JobInfoActor.scala:46)
at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
at ooyala.common.akka.ActorStack$$anonfun$receive$1.applyOrElse(ActorStack.scala:33)
at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
at ooyala.common.akka.Slf4jLogging$$anonfun$receive$1$$anonfun$applyOrElse$1.apply$mcV$sp(Slf4jLogging.scala:26)
[...]

Issue #323, opened 2015-11-25
Enhancement: health check


ChenYanting:
Hi 

I have a suggestion for SJS health check. Since SJS has to cooperate with Spark master, we can view them as a working unit. As a result, I think the health check of SJS could include not only SJS itself but also spark masters.

The result of health check has to tell admins if both of SJS and Spark masters are on or any of both is down.

velvia:
Not a bad idea.   Or the health check can return a JSON blob which contains the health of SJS + Master.

My only concern is that in theory, each job can have a separate master.   Then what do you check?

ChenYanting:
In SJS configuration, we can set which masters SJS are going to connect. My idea is that the health check can confirm if the masters in master string are all on.

SeaAndHillMe:
this future is very goog

PR #1196, opened 2019-05-03
feat(webapi): check sjs actors liveness

SrivigneshM:
* implement healthz webapi
    * send messages to akka actors
    * check if the actors reply with expected response
    * add tests to healthz webapi

**Pull Request checklist**

- [Y] The commit(s) message(s) follows the contribution [guidelines](doc/contribution-guidelines.md#commit-message-format) ?
- [Y] Tests for the changes have been added (for bug fixes / features) ?
- [Y] Docs have been added / updated (for bug fixes / features) ?

**New behavior :**

**BREAKING CHANGES**

NA

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1196)


job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -514,10 +515,47 @@ class WebApi(system: ActorSystem,
...
     //no authentication required
     get { ctx =>
       logger.info("Receiving healthz check request")
-      ctx.complete("OK")
+      var actorsAlive: Int = 0
+      val supervisorFuture = (supervisor ? GetContext("dummycontext")).mapTo[NoSuchContext.type]
+      val jobInfofuture = (jobInfoActor ? GetJobStatus("dummyjobid")).mapTo[NoSuchJobId.type]
+      val jobDaofuture = (daoActor ? GetJobInfos(1)).mapTo[JobInfos]
+      val statusActor = system.actorOf(JobStatusActor.props(daoActor))
+      val jobStatusfuture = (statusActor ? Unsubscribe("dummyjobid", null)).mapTo[NoSuchJobId.type]
```
valan4ik: I am not sure I see the point of creating statusActor and then sending a message to it here. I would say these two line should be removed.

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -514,10 +515,47 @@ class WebApi(system: ActorSystem,
...
+      val resultActor = GetResultActor("dummycontext")
+      val jobResultfuture = (jobInfoActor ? GetJobResult("dummyjobid")).mapTo[NoSuchJobId.type]
+
+      val listOfFutures = Seq(supervisorFuture, jobInfofuture, jobDaofuture, jobStatusfuture, jobResultfuture)
+      val futureOfList = Future.sequence(listOfFutures)
+
+      try {
+        val results = Await.result(futureOfList, 60 seconds)
```
valan4ik: Our general spray timeout is 40 seconds (in application.conf `request-timeout = 40 s`). 
Also I think it's enough to have 10-15 seconds timeout here. It's a general timeout in Jobserver and futures are executed in parallel, so 6 futures shouldn't make timeout 6 times longer.

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -514,10 +515,47 @@ class WebApi(system: ActorSystem,
...
+        for (result <- results) {
+          result match {
+            case NoSuchContext => actorsAlive += 1
+            case NoSuchJobId => actorsAlive += 1
+            case JobInfos(jobInfos) => actorsAlive += 1
+            case _ => logger.error("Unexpected response!")
```
valan4ik: Logger message is not super helpful, would be nice to log unexpected response itself.

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -514,10 +515,47 @@ class WebApi(system: ActorSystem,
...
+        case ex: Exception => logger.error(ex.getMessage())
+      }
+
+      if (actorsAlive == 5) {
+        logger.info("Required actors alive")
+        ctx.complete(StatusCodes.OK)
+      }
+      else {
+        logger.error("Required actors not alive")
+        ctx.complete(500, errMap("Required actors not alive"))
```
valan4ik: Just as a suggestion: Maybe after removing a check for some of the actors above, you can also easier track which ones are not here.  Otherwise you will need to waste time debugging which actor is not present.

SrivigneshM:
> 
> 
> @SrivigneshM ping

I'm taking a look at the comments

bsikander:
Related https://github.com/spark-jobserver/spark-jobserver/issues/323

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -514,10 +515,47 @@ class WebApi(system: ActorSystem,
...
     get { ctx =>
       logger.info("Receiving healthz check request")
-      ctx.complete("OK")
+      var actorsAlive: Int = 0
+      val supervisorFuture = (supervisor ? GetContext("dummycontext")).mapTo[NoSuchContext.type]
+      val jobInfofuture = (jobInfoActor ? GetJobStatus("dummyjobid")).mapTo[NoSuchJobId.type]
+      val jobDaofuture = (daoActor ? GetJobInfos(1)).mapTo[JobInfos]
+      val statusActor = system.actorOf(JobStatusActor.props(daoActor))
+      val jobStatusfuture = (statusActor ? Unsubscribe("dummyjobid", null)).mapTo[NoSuchJobId.type]
+      val resultActor = GetResultActor("dummycontext")
```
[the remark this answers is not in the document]
SrivigneshM: Removed resultActor

job-server/src/main/scala/spark/jobserver/JobServer.scala
```diff
@@ -155,9 +156,10 @@ object JobServer {
   }
 
   def startWebApi(system: ActorSystem, supervisor: ActorRef, jobDAO: JobDAO,
-      webApiPF: (ActorRef, ActorRef) => WebApi) {
+      webApiPF: (ActorRef, ActorRef, ActorRef) => WebApi) {
     val jobInfo = system.actorOf(Props(classOf[JobInfoActor], jobDAO, supervisor), "job-info")
-    webApiPF(supervisor, jobInfo).start()
+    val daoActor = system.actorOf(Props(classOf[JobDAOActor], jobDAO), "dao-manager")
```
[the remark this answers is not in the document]
SrivigneshM: Reused daoActor

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -514,10 +515,47 @@ class WebApi(system: ActorSystem,
...
+        for (result <- results) {
+          result match {
+            case NoSuchContext => actorsAlive += 1
+            case NoSuchJobId => actorsAlive += 1
+            case JobInfos(jobInfos) => actorsAlive += 1
+            case _ => logger.error("Unexpected response!")
+         }
+        }
+      } catch {
+        case ex: Exception => logger.error(ex.getMessage())
```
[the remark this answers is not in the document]
SrivigneshM: Addressed

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: Can you please elaborate on why this is not so useful?

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: I reduced the timeout to 30 seconds.

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: Replaced the wild card with a default variable and logged it.

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: I shall take up this suggestion as an enhancement to this API.

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: ping @valan4ik

valan4ik: If I understand it correctly, you create a new status actor here and try to send him a message.
But this actor is just artificially created here and is not part of the live system.
I am not sure I get the idea and what value does it bring, if the request fails it doesn't say anything about current cluster status.
In case of other actors, you check the ones, which form the akka cluster (like supervisor, which is a core of jobserver and of course if he is not here, system is unhealthy).

bsikander:
I had one question, in which case, you think that actors won't be alive?
Once they are started, they stay. If an exception occurs, they transparently restart.

SrivigneshM:
> 
> 
> I had one question, in which case, you think that actors won't be alive?
> Once they are started, they stay. If an exception occurs, they transparently restart.

Like in below issue, I have noticed that even if they try to restart, it doesn't succeed and end up shutting down.
https://github.com/spark-jobserver/spark-jobserver/issues/1149

bsikander:
@SrivigneshM do you know why it shuts down? I was trying to look but didn't find anything yet.

SrivigneshM:
> 
> 
> @SrivigneshM do you know why it shuts down? I was trying to look but didn't find anything yet.

That is unknown for me as well. Only thing different here from other issues is that the actor never recovers.

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: Taken care of this

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -530,9 +532,44 @@ class WebApi(system: ActorSystem,
...
    */
   def healthzRoutes: Route = pathPrefix("healthz") {
+    import ContextSupervisor._
+    import JobManagerActor._
+    import JobInfoActor._
+    import JobDAOActor._
     //no authentication required
     get { ctx =>
-      ctx.complete("OK")
+      logger.info("Receiving healthz check request")
```
bsikander: Your change looks okay, the only issue I see is that the definition of health check varies. Some people might want to check DB + Spark + HDFS also as part of health check and others might just want to know if WebApi is okay.

I would suggest, to hide your implementation behind an interface and provide 2 implementations. One could be the old one and other can be your implementation and it can be configurable via config file. 
An example of making it configurable can be found [here](https://github.com/spark-jobserver/spark-jobserver/blob/d705751418d65fd75a8ba2f928bab8baf2c2725e/job-server/src/main/resources/application.conf#L37) and its usage can be seen [here](https://github.com/spark-jobserver/spark-jobserver/blob/d705751418d65fd75a8ba2f928bab8baf2c2725e/job-server/src/main/scala/spark/jobserver/JobServer.scala#L70) and [here](https://github.com/spark-jobserver/spark-jobserver/blob/d705751418d65fd75a8ba2f928bab8baf2c2725e/job-server/src/main/scala/spark/jobserver/JobServer.scala#L115)

bsikander: Please don't hesitate to ask if you get stuck somewhere. Thanks.

bsikander:
ping @SrivigneshM

valan4ik:
Btw, there is Akka Management extension. I still didn't have time to try it out, but it seems like it can replace pinging the actors and be used for health check: https://doc.akka.io/docs/akka-management/current/cluster-http-management.html

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: @bsikander Please review and merge if the changes are good.
- Made the health check configurable. 
- By default this returns OK, health check will be executed only if it is explicitly configured in application.conf
- Test is failing currently but it is not relevant to this change.

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
```diff
@@ -0,0 +1,81 @@
...
+import spray.httpx.SprayJsonSupport.sprayJsonMarshaller
+import spray.json.DefaultJsonProtocol._
+import spray.http.StatusCodes
+import spray.routing.{HttpService, RequestContext}
+import spark.jobserver.io.JobInfo
+import spark.jobserver.io.JobDAOActor
+
+trait HealthCheckService {
+
+  def healthCheckImpl(ctx: RequestContext, supervisor: ActorRef, jobInfo: ActorRef,
```
bsikander: `healthCheckImpl` is a strange name for a function. It doesn't tell me what it is trying to do.

`isHealthy` seems to be more appropriate with a return type of `boolean`.
I do not like that we have to pass the request object down to this class. This function should just return a boolean and the `WebApi` should convert the boolean to `ctx.complete(....)`

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
```diff
@@ -0,0 +1,81 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.control.NonFatal
+import spray.httpx.SprayJsonSupport.sprayJsonMarshaller
+import spray.json.DefaultJsonProtocol._
+import spray.http.StatusCodes
+import spray.routing.{HttpService, RequestContext}
+import spark.jobserver.io.JobInfo
+import spark.jobserver.io.JobDAOActor
+
+trait HealthCheckService {
```
bsikander: nit:
`HealthCheck` is enough (in my opinion)

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
bsikander: Further, the arguments that you are taking in are very specific to your implementation of actors health check. Can we improve this?

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -561,9 +562,18 @@ class WebApi(system: ActorSystem,
...
-      ctx.complete("OK")
+      try {
+          val healthCheckClass = Class.forName(config.getString("spark.jobserver.healthcheck"))
+          val instance = healthCheckClass.newInstance.asInstanceOf[HealthCheckService]
+          instance.healthCheckImpl(ctx, supervisor, jobInfoActor, daoActor)
+        }
+        catch {
+          case ex: Exception => {
+            logger.error("Exception in healthz", ex)
+            ctx.complete(500, errMap("Exception while invoking health check"))
```
bsikander: nit:
Instead of 500, you should use `StatusCodes......`.

job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -561,9 +562,18 @@ class WebApi(system: ActorSystem,
    *    GET /healthz              - return OK or error message
    */
   def healthzRoutes: Route = pathPrefix("healthz") {
-    //no authentication required
     get { ctx =>
-      ctx.complete("OK")
+      try {
+          val healthCheckClass = Class.forName(config.getString("spark.jobserver.healthcheck"))
```
bsikander: Health check is something that is called very often, wouldn't it be better to create a global variable in this class and instantiate the healthcheck object once?

job-server/src/main/resources/application.conf
```diff
@@ -130,6 +130,9 @@ spark {
     # Start embedded H2 Server (usefull in cluster deployment)
     startH2Server = false
 
+    # Health Check Class to be invoked for healthz API
+    healthcheck = spark.jobserver.APIHealthCheck
```
bsikander: What is your opinion on putting this HealthCheck class in utils package?
`spark.jobserver.util.....`

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def healthCheckImpl(ctx: RequestContext, supervisor: ActorRef, jobInfo: ActorRef,
+                      daoActor: ActorRef)
+
+}
+
+class ActorsHealthCheck extends HealthCheckService {
```
bsikander: Since you want to verify that all user created actors are alive, have you considered sending an `Identify` message to all actors e.g. `/users/* ! Identify` ? If you receive response from supervisor, dao, info and result actor then everything is fine. By doing this your code will be simplified and you won't need to pass all the actors inside.

bsikander:
@Nibooor @valan4ik could you also review it once?

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
SrivigneshM: I am fine with renaming the function and changing the return type. But I do not understand your comment about arguments. Can you explain little more about arguments?

bsikander: I meant that the implementations of HealthCheckService can vary. Some might have an implementation to check jobserver + hdfs + zookeeper others might just have a check for actors. The arguments seem specific to your implementation. I was saying that the trait function arguments should be a bit more generic.

job-server/src/main/scala/spark/jobserver/WebApi.scala
valan4ik: +1 on this as well. There is a function `logAndComplete`

job-server/src/main/scala/spark/jobserver/WebApi.scala
valan4ik: Would be possible to create HealthCheck class still in Jobserver.scala and pass it to WebApi?
Then there is no need to pass WebApi daoActor, which is a dependency of HealthCheckService.

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
SrivigneshM: @bsikander I tried this approach. But seems like this will work only if the sender is also an actor. In this case, WebApi or HealthCheck is not an actor. So I might have to send message to individual actors.

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: @valan4ik I get the point that I can create HealthCheck class in JobServer.scala itself and pass to WebApi. But I do not understand why there is no need to pass daoActor. While invoking the healthCheckImpl function from WebApi, I might still need to pass daoActor right?

SrivigneshM: ping @valan4ik

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
SrivigneshM: ping @bsikander

job-server/src/main/scala/spark/jobserver/WebApi.scala
valan4ik: From my understanding it should look like this:
- Create helthCheckImpl instance in JobServer.scala with all needed parameters: daoActor and any other actor we may add later.
- Pass already created object to WebApi. WebApi doesn't need daoActor as far as I remember. All calls to daoActor happen inside healthCheckImpl, so you just call needed methods of healthCheckImpl.

A huge advantage is that if healthCheckImpl takes any additional parameters later, there is no need to add new parameters for WebApi. Only change the initialization of the object in JobServer.scala.

Hope I didn't miss anything during review :)

SrivigneshM: Thank you. I got the point. I will construct the object in JobServer.scala.

valan4ik: Okay, I looked into the code again. It's maybe a bit following @bsikander comment as well. 
At the moment you create a new instance here each time. But there is no need for it.
supervisor, jobInfoActor, daoActor don't change.
It's enough to initialize class with this parameters:
```
class ActorsHealthCheck(ctx: RequestContext, supervisor: ActorRef, jobInfo: ActorRef, daoActor: ActorRef) 
extends HealthCheckService
...
```
save them as class variables and just call `healthCheckImpl()` function on a class, without passing `supervisor, jobInfoActor, daoActor` each time.

Meaning depending on implementation class which extends `HealthCheckService` may take different parameters, but main function `healthCheckImpl` may be renamed just to `executeHealthCheck` and don't take any parameters.
Then we just need to be careful during the initialization of the health check instance.

Not sure if @bsikander had smth similar in mind.

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
SrivigneshM: Addressed

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: Addressed

job-server/src/main/scala/spark/jobserver/HealthCheck.scala
SrivigneshM: Addressed

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: Addressed

job-server/src/main/resources/application.conf
SrivigneshM: Addressed

job-server/src/main/scala/spark/jobserver/JobServer.scala
```diff
@@ -158,15 +160,23 @@ object JobServer {
...
   }
 
   def startWebApi(system: ActorSystem, supervisor: ActorRef, jobDAO: JobDAO,
-      webApiPF: (ActorRef, ActorRef) => WebApi) {
+      webApiPF: (ActorRef, ActorRef, HealthCheck) => WebApi, daoActor: ActorRef, config: Config) {
     val jobInfo = system.actorOf(Props(classOf[JobInfoActor], jobDAO, supervisor), "job-info")
-    webApiPF(supervisor, jobInfo).start()
+    val healthClass = Class.forName(config.getString("spark.jobserver.healthcheck"))
+    var healthCheckInst: HealthCheck = null
+    if (healthClass.getName() == "spark.jobserver.util.ActorsHealthCheck") {
```
Nibooor: nit:
This instantiation could be a bit more dynamic I guess (like have a common constructor in the trait `HealthCheck` - currently it would have the arguments supervisor, jobInfo, daoActor - that every instance needs to implement). Especially if there would be multiple implementations in the future this would become a bit blown here otherwise.

But this can also be done as part of later refactorings, I see you have put a lot effort in this PR already!

job-server/src/main/scala/spark/jobserver/WebApi.scala
SrivigneshM: Addressed

job-server/src/main/scala/spark/jobserver/JobServer.scala
SrivigneshM: I understand the point that this might get blown in future. But having common constructor with actors as arguments may not be relevant for other implementations of healthCheck right? For ex, we have APIHealthCheck class in HealthCheck.scala. This does not take any args like supervisor, jobInfo, daoActor as they are not relevant for a simple API health check.

noorul:
Thanks @SrivigneshM  for the patience.

SeaAndHillMe:
can some give a demo  for how to call the health api

Issue #1148, opened 2019-01-22
Proposal: a new design for DAO layer


valan4ik:
The purpose of this issue is to propose a new design for DAO layer while keeping backward compatibility in mind.

Problems with JobDAO interface:
- Some functions have no return type e.g saveBinary/deleteBinary/saveJobInfo/saveJobConfig
Returns types are different for different cases ranging from Future to Unit.
- Among generic functions, we have specific functions like getBinaryFilePath/getLastUploadTimeAndType/cleanRunningJobInfosForContext
- Function names are inconsistent
- Storing BLOBs in database is considered bad practice. From a practical perspective, people are uploading 50MB jar files to DB on regular basis. These large objects cause all sorts of issues e.g.
DB getting stuck on some large blob, DB corruption (H2) and not really scalable.

The new design introduces 2 new interfaces MetaDataDAO and BinaryDAO and a new implementation of JobDAO named CombinedDAO.

- MetaDataDAO represents all the metadata within Jobserver including binaries.
- BinaryDAO represents a simple interface for CRUD operations related to storing BLOBs.
- CombinedDAO acts as a glue between MetaDataDAO and BinaryDAO and implements JobDAO interface.
The foreign key between MetaDataDAO and BinaryDAO would be the hash of binary.

Ideally, the whole JobDAO interface should be broken but it would be too big of a change as it would have ripples across the whole project. Once both MetaDataDAO and BinaryDAO interfaces are mature enough, they should replace JobDAO.

The new proposed design is shown below.

![combineddao](https://user-images.githubusercontent.com/1217732/51547472-ac6f8d00-1e66-11e9-9df7-61c32eebb68e.png)

The final goal, as mentioned, is to replace JobDAO with MetaDataDAO and BinaryDAO. Having proper return types for all DAO functions will also allow us not to propagate DAO exceptions up to WebAPI, but to handle them correctly before. A bit of glue code (needed only for 2-3 operations, which update both meta and binary) could be done already in JobDAOActor itself.

The proposed end state design is shown below.

![daosplit](https://user-images.githubusercontent.com/1217732/51547493-bd200300-1e66-11e9-8009-1a1c5540ad9d.png)

bsikander:
I am ok with the design.
100MB jars are a big problem in DB and separation will help.

valan4ik:
DAO split is finished and last commits were recently merged into the master. Closing the issue.

PR #1199, opened 2019-05-10
Create JobInfo with full BinaryInfo object

valan4ik:
**Pull Request checklist**

- [x] The commit(s) message(s) follows the contribution [guidelines](doc/contribution-guidelines.md#commit-message-format)
- [x] Tests for the changes have been added (for bug fixes / features)

**Current behavior :** 

After introduction of a [new dao split](https://github.com/spark-jobserver/spark-jobserver/issues/1148), there is binary hash value in BinaryInfo. For the new DAO
it's a connection between meta data and BLOB storage.

**New behavior :**

Create a job using BinaryInfo instead of only timestamp and type.
Although only timestamp and type are needed to start the job,
for the consistency and full information in JobInfo it is better
to use full BinaryInfo.

Make JobSqlDAO/MetaDataSqlDAO return JobInfo including binaryStorageId.
Before the fix getBinaryInfo in MetaDataSqlDAO was returning BinaryInfo
including full information, but getJob/getJobs in both DAOs were
returning JobInfo objects with BinaryInfo without binHash.


bsikander:
LGTM. Merging. Thanks.

PR #1210, opened 2019-06-03
fix(tests): Fix saveJobConfig tests

bsikander:
saveJobConfig function in JobDAO has Unit return
value. This means that if we save a config
and read it again in next line then it boils down
to timing and the test becomes flaky.

This change adds a retry while reading the config
to fix the test.

Further, SqlCommonSpec uses the saveJobConfig from
SqlCommon which has a return value. So, in this
test, we just add Await.

**Pull Request checklist**

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1210)

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/util/Utils.scala
+++ b/job-server/src/main/scala/spark/jobserver/util/Utils.scala
@@ -1,6 +1,7 @@
 package spark.jobserver.util
 
 import java.io.{Closeable, File}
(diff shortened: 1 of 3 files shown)


bsikander:
The Utils class doesn't have a spec file but it gets tested indirectly. 

@noorul can you have a look at it?

Issue #1236, opened 2019-07-24
Jobserver HA proposal


Nibooor:
During the last months, @pk-work, @ecandreev, @valan4ik, @bsikander and I made some changes that in its entirety result in Jobserver being more or less highly available.

We have already contributed parts of it and keep contributing the other parts stepwise, so the community can review it / give feedback more easily.

This issue is meant to share our rationale behind the architecture and give a platform for questions, discussions and feedback.

# Roadmap
To give the changes a context and share our thoughts behind the design, here's the roadmap of sequential steps we took to achieve Jobserver HA:

### CombinedDAO with HDFS
The first set of changes to achieve Jobserver HA was to make the underlying database setup highly available.
As discussed in more detail in [#1148](https://github.com/spark-jobserver/spark-jobserver/issues/1148), we first split the JobDAO into a BinaryDAO and MetadataDAO avoiding the persistence of blobs in a regular database.
We prefered an **HDFS** setup to realize highly available binary persistence, but the abstraction of a BinaryDAO can also work with other blob storing techniques.

### MetadataDAO with Zookeeper
Using the aforementioned CombinedDAO architecture the highly available database for metadata of our choice is **Zookeeper**. 
As the jobserver API is synchronous and the eventual consistency of zookeeper would lead to "flaky" behavior in the API, we synchronize all zookeeper nodes before every read operation, which ensures consistency in our case (meaning every previously completed write operation is reflected in the next read). 
Some of our fixes (e.g. autopurge feature) addressed the performance difference and modified timeouts that come with a distributed system compared to a relational database.
Of course, other HA implementations of a MetadataDAO are possible in the future.

### Cluster Singleton
With an HA database as a foundation we decided to go for an **active-active** setup of two or more jobservers. The reason is that most parts of the jobserver are stateless and the WebAPI and actors of multiple jobservers could work in parallel without interfering with each other, synchronizing via the database. You then don't have to care about a proper active-passive switch plus you get the possibility of an external load balancing mechanism.

The only thing that has to be coordinated across multiple jobservers is the akka cluster supervision. More specifically, the MemberUp and Terminated event of the AkkaClusterSupervisor should not be handled by multiple jobservers to avoid state machine problems.
To realize this we promoted the AkkaClusterSupervisor to an [Akka Cluster Singleton](https://doc.akka.io/docs/akka/current/cluster-singleton.html), which ensures that there is always only one actor of this type in the cluster.

### Split Brain Resolver (SBR)
[...]

PR #1239, opened 2019-08-09
feat(jobserver): Make AkkaClusterSupervisorActor a singleton

Nibooor:
Following the [HA proposal mentioned in issue #1236](https://github.com/spark-jobserver/spark-jobserver/issues/1236), this singleton change represents step 3/5 to achieve jobserver HA:

**Details**

* Create AkkaClusterSupervisorActor as an akka cluster singleton
  on jobserver startup
* Adapt application.conf for a cluster singleton
* Adapt JobserverSpec to also check for singleton infrastructure
  in case of context-per-jvm=true
* Temporarily adapt changes to jobManager zombie kill features as
  long as we have no split brain resolver.

In jobserver HA mode the logic following a MemberUp event
(initializing a context) or Terminated event (updating context
and associated job states in the database) can conflict when
handled by two or more active jobservers. Moving this logic to a
cluster singleton ensures that this logic is executed exclusively
by one actor.
See: https://doc.akka.io/docs/akka/current/cluster-singleton.html

Making the AkkaClusterSupervisorActor a singleton is a relatively
simple, straight-forward solution, which requires few changes and
is therefore a good initial solution for HA. The downside is
that a singleton now handles more than just the necessary
AkkaClusterEvents (i.e. context management parts, which could be
handled locally by each jobserver), which represents a potential
bottleneck in the future.

A recommended optimization for the future would be to refactor
the AkkaClusterSupervisorActor into two actors (cluster and
context managing parts), where only the cluster parts will be
handled by a singleton. To not overburden the initial HA of
jobserver, this refactoring is left out for now.

**Pull Request checklist**

- [X] Docs have been added / updated (for bug fixes / features) ?

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1239)

### DIFF
--- a/job-server/src/main/resources/application.conf
+++ b/job-server/src/main/resources/application.conf
@@ -255,6 +255,19 @@ akka {
     failure-detector.acceptable-heartbeat-pause = 3s
     failure-detector.threshold = 12.0
   }
+
+  cluster.singleton {
+    singleton-name = "context-supervisor"
+    role = "supervisor"
+  }
+
+  cluster.singleton-proxy {
+    singleton-name = ${akka.cluster.singleton.singleton-name}
+    role = ${akka.cluster.singleton.role}
+    singleton-identification-interval = 1s
+    buffer-size = 1000
+  }
+
 }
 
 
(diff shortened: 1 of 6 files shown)


bsikander:
LGTM.

PR #1241, opened 2019-08-21
feat(jobserver): Add HA setup for jobserver

bsikander:
With recent changes, spinning up multiple
jobserver instances is possible. This change
adds required changes to setup an HA jobserver
cluster.

The address(es) can be specified in comma
separated format in spark.jobserver.ha.masters
section.

At startup, jobserver automatically forms
a new cluster with all the masters or joins
the old cluster of slaves. While forming the
cluster for the first time, the first node
shouldn't fail to come. If it does then the
secondary node will be hanging forever.

kill-context-on-supervisor-down is used in
client mode for killing the slaves if supervisor
is down. This was implemented because at the
time, we didn't had any good split brain resolver.
We were relying on auto-down to mark the node
as DOWN and kill-context-on-supervisor-down to
actually kill the node. Split brain will take
care of both now. Currently, this HA solution
partially breaks the kill-context-on-supervisor
feature because slaves react only on the state of
1 master. In client mode, if HA setup is required
then a proper split brain resolver should be used
with kill-context-on-supervisor-down disabled.

On the slave side, I had 2 options
Option#1: Use the new ha.masters section and use it to
directly to join the masters. If this section is not
defined then use the arguments pass in JVM
Option#2: From the master pass all masters in JVM
arguments

I opted for Option#2 because it is simple and as a
developer, I don't have another branch in my mind
on how master addresses reach the slaves.

Testing:
- Tested with 1 and 2 master setup with cluster +
  supervise mode
- Tested with client mode

Notes:
- If master goes down in cluster mode,
  and comes up with a new dynamic port, then H2
  database on slave will fail to connect. Not
  a problem in ZK dao implementation.
- If port is dynamic in HA mode, then on slave
  restart, it won't know the seed nodes to connect to.
  So, port hardcoding is required.

Change-Id: I0553d0e400d471a6de70f86b9e94096231f43939

**Pull Request checklist**

**New behavior :**

**BREAKING CHANGES**

**Other information**:

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1241)

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/JobServer.scala
+++ b/job-server/src/main/scala/spark/jobserver/JobServer.scala
@@ -1,17 +1,16 @@
 package spark.jobserver
 
-import akka.actor.{ActorContext, ActorNotFound, ActorRef, ActorSystem, AddressFromURIString, Props}
+import akka.actor.{ActorContext, ActorNotFound, ActorRef, ActorSystem, Address, AddressFromURIString, Props}
 import akka.util.Timeout
 import akka.pattern.ask
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent.{InitialStateAsEvents, MemberEvent}
 import com.typesafe.config.{Config, ConfigFactory, ConfigValueFactory}
-
 import java.io.File
 import java.util.concurrent.{TimeUnit, TimeoutException}
 
 import spark.jobserver.io._
-import spark.jobserver.util.{ContextReconnectFailedException, DAOCleanup}
+import spark.jobserver.util.{ContextReconnectFailedException, DAOCleanup, Utils, WrongFormatException}
 import org.joda.time.DateTime
 import org.slf4j.LoggerFactory
 
@@ -46,6 +45,8 @@ object JobServer {
   implicit val ec = scala.concurrent.ExecutionContext.Implicits.global
 
   class InvalidConfiguration(error: String) extends RuntimeException(error)
+
+  val ACTOR_SYSTEM_NAME = "JobServer"
 
   // Allow custom function to create ActorSystem.  An example of why this is useful:
   // we can have something that stores the ActorSystem so it could be shut down easily later.
@@ -115,6 +116,8 @@ object JobServer {
       logger.info("Embeded H2 server started with base dir {} and URL {}", rootDir, h2.getURL: Any)
     }
 
+    val haSeedNodes = getAndValidateHASeedNodes(config)
+
     // Start actors and managers
     val ctor = jobDaoClass.getDeclaredConstructor(Class.forName("com.typesafe.config.Config"))
     val jobDAO = ctor.newInstance(config).asInstanceOf[JobDAO]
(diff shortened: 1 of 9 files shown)


bsikander:
https://github.com/spark-jobserver/spark-jobserver/issues/1236

bsikander:
@noorul @velvia could you please have a look at this change?

job-server/src/main/scala/spark/jobserver/JobServer.scala
```diff
@@ -166,15 +169,37 @@ object JobServer {
     webApiPF(supervisor, jobInfo).start()
   }
 
-  private def joinAkkaCluster(cluster: Cluster, seedNodes: List[ActorRef]) {
-    if (seedNodes.isEmpty) {
+  /**
+    * This function is responsible for joining the Akka cluster dynamically
+    * based on current situation of Akka cluster.
+    *
+    * - If there are no old slave nodes and HA section of config is not
```
noorul: By slave nodes, did you mean peer master nodes? I am confused by the terminology slave here.

bsikander: Sorry for confusion, by slaves, I mean driver JVMs.

This case just means that if HA is not configured and no driver jvms are alive then the current jobserver node will join itself.

job-server/src/test/scala/spark/jobserver/util/UtilsSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+      intercept[WrongFormatException] {
+        Utils.getHASeedNodes(config)
+      }
+
+      intercept[WrongFormatException] {
+        Utils.getHASeedNodes(config1)
+      }
+    }
+
+    it("should return seed nodes") {
```
noorul: Can we have one test for multiple nodes?

job-server/src/test/scala/spark/jobserver/JobServerSpec.scala
```diff
@@ -374,6 +375,26 @@ class JobServerSpec extends TestKit(JobServerSpec.system) with FunSpecLike with
...
+      cluster.subscribe(testActor, ClusterEvent.InitialStateAsSnapshot, classOf[ClusterEvent.MemberEvent])
+      expectMsgClass(classOf[ClusterEvent.CurrentClusterState])
+
+      JobServer.joinAkkaCluster(cluster, List.empty, List.empty)
+
+      expectMsgClass(classOf[ClusterEvent.MemberUp])
+
+      within(5.seconds) {
+        awaitAssert {
+          cluster.state.members.size should ===(1)
```
noorul: Can we leave space after === ?

job-server/src/test/scala/spark/jobserver/JobServerSpec.scala
```diff
@@ -374,6 +375,26 @@ class JobServerSpec extends TestKit(JobServerSpec.system) with FunSpecLike with
...
+          cluster.state.members.head.status should ===(MemberStatus.Up)
```
noorul: Same as above. I wonder why scalastyle is not catching this?

noorul:
re-triggered the build

bsikander:
@noorul I have retriggered the build multiple times, there are a few flaky tests which are causing the build to fail. Maybe we can go ahead and merge this change. I will push the fix for flaky changes soon.

bsikander:
@noorul the tests passed, you can review the change again.

PR #1251, opened 2019-09-30
fix(sbr): Should subscribe to events early and exit with -1 without hooks

valan4ik:
**Pull Request checklist**
- [x] The commit(s) message(s) follows the contribution [guidelines](doc/contribution-guidelines.md#commit-message-format)

**Current behavior :** 
Previous implementation with `sys.exit(-1)` was calling Spark shutdown hooks
and sometimes putting context in KILLED stated (depending on success of the
DAO operation).

**New behavior :**
The `java.lang.Runtime.halt(int status)` method forcibly terminates the currently
running Java virtual machine. This method never returns normally. Unlike the
exit method, this method does not cause shutdown hooks to be started and does
not run uninvoked finalizers if finalization-on-exit has been enabled.

Split brain resolver class is initialised once the cluster is created.
When cluster is in inconsistent state, node may be in a JOINING
state and never reach UP state (e.g. node was joining the cluster and
right at the same moment all other nodes in the cluster went down).
To avoid this situation split brain resolver should start to react on
cluster events right after initialisation.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1251)

### DIFF
--- a/job-server/src/main/scala/akka/downing/KeepOldestAutoDown.scala
+++ b/job-server/src/main/scala/akka/downing/KeepOldestAutoDown.scala
@@ -135,7 +135,7 @@ class KeepOldestAutoDown(preferredOldestMemberRole: Option[String],
       Await.result(context.system.terminate(), 10 seconds)
     } else {
       logger.info("ShutdownSelf: exiting the system with -1")
-      sys.exit(-1)
+      Runtime.getRuntime.halt(-1)
     }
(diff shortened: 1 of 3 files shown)


bsikander:
Tests are failing due to unrelated issues. I will push the fix for tests in a separate change. Merging this one.

Issue #1151, opened 2019-01-28
dependent-jar-uris not picked up


walmir:
**Used Spark version**: 2.3.1

**Used Spark Job Server version**: Master
(Released version, git branch or docker image version)

**Deployed mode**: Cluster - Spark Standalone

**Actual (wrong) behavior**: dependent-jar-uris parameter is not picked up. 

**Steps to reproduce**
1. Placed `base.jar` on SJS machine under `/tmp/base.jar`.
2. Created new context with `dependent-jar-uris`
3. Started Job that requires `base.jar`
-> Always results in `ClassNotFoundException`

I've tried using `dependent-jar-uris` as a URL parameter and in the POST body when creating the context. I've tried setting it in the context-settings as: `context-settings.dependent-jar-uris = ["file:///tmp/base.jar"]`
And I've tried submitting it when deploying a job, that requires this jar. Non worked.

I would really appreciate your help on this.

Thanks.

walmir:
I just noticed the following log statements when I try to start a Job:
```
[2019-01-28 20:36:32,950] INFO  AkkaClusterSupervisorActor [] [] - SparkContext new-context joined
[2019-01-28 20:36:40,484] WARN  ReliableDeliverySupervisor [] [akka.tcp://JobServer@ec2----.compute.amazonaws.com:38386/system/endpointManager/reliableEndpointWriter-akka.tcp%3A%2F%2FJobServer%4010.10.2.61%3A40499-6] - Association with remote system [akka.tcp://JobServer@10.10.2.61:40499] has failed, address is now gated for [5000] ms. Reason: [Disassociated]
[2019-01-28 20:36:44,760] WARN  .cluster.ClusterCoreDaemon [] [akka.tcp://JobServer@ec2----.compute.amazonaws.com:38386/system/cluster/core/daemon] - Cluster Node [akka.tcp://JobServer@ec2----.compute.amazonaws.com:38386] - Marking node(s) as UNREACHABLE [Member(address = akka.tcp://JobServer@10.10.2.61:40499, status = Up)]. Node roles [supervisor]
[...]

PR #1279, opened 2020-02-17
fix(dependent-jars): Dependent-jar-uris for context config

valan4ik:
Add mapping to the old dependent-jar-parameter for the
context config. Currently posting dependent jars for the context with `dependent-jar-uris` parameter doesn't work.

More information can be found in this issue: https://github.com/spark-jobserver/spark-jobserver/issues/1151

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1279)


bsikander:
+1

PR #1295, opened 2020-04-23
doc(jmx): Add example how to down a node from jmx-cli

valan4ik:
The command can be very useful in some cases, so I thought it would be good to add it to jmx documentation in Jobserver to avoid long search in Google in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1295)


bsikander:
This is just a documentation change so merging. Tests are not required.

PR #1329, opened 2020-11-04
fix: Fail postgres migration on exception

Ennosigaeon:
**Pull Request checklist**

**Current behavior :** Previously, exceptions during postgres migration where only logged but did not abort migration. For example, if the postgres extension lo is not present, V0.7.3 and V0.7.5 log an exception but are recorded as successful. V0.7.6 finally fails, as an expected table is not present. As a result, the migration will always continue from V0.7.6 but never succeed, even when the extension was manually added.

**New behavior :** The flyway migration will correctly record a migration step as failed if an exception is raised. Subsequent migrations will continue from the failed step.

I think the changes are only relevant for the jobserver-0.10.x branch and not for master. But if the changes are also relevant for master I can provide an additional PR.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1329)

### DIFF
--- a/job-server/src/main/scala/db/postgresql/migration/V0_7_3/V0_7_3__Migrate_Blobs.scala
+++ b/job-server/src/main/scala/db/postgresql/migration/V0_7_3/V0_7_3__Migrate_Blobs.scala
@@ -2,21 +2,17 @@ package db.postgresql.migration.V0_7_3
 
 import java.sql.Connection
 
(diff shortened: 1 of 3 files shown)


valan4ik:
@Ennosigaeon Could you please rebase you PR on the current `jobserver-0.10.x`?

Ennosigaeon:
Done

PR #1343, opened 2021-01-26
misc(Python): Document wheel support

Ennosigaeon:
Jobserver already supports Python wheel files by implicitly treating them as eggs. Yet, this ability is neither documented nor reflected in the code (i.e. variable names). This commit makes the support of wheel files more explicit.

**Pull Request checklist**

I seems like Python wheels are already supported, kind of... whl files are imported and stored as egg files and than passed to spark. As whl and egg both are just zip files this works just fine but is very ugly. I have added explicit support for whl files such that they are recognized and stored correctly. Furthmore, I have adapted the docu and variable names to reflect that both whl and egg files can be processed.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1343)


doc/python.md
```diff
@@ -140,7 +140,7 @@ would be fine.
 
 ### Packaging a job
 
-In order to be able to push a job to the the job server, it must be packaged into a Python Egg file. Similar to a Jar,
+In order to be able to push a job to the job server, it must be packaged into a Python Egg file. Similar to a Jar,
```
bsikander, line 143: should we mention about wheel file here?

doc/python.md
```diff
@@ -163,32 +163,42 @@ In the most basic setup, a job ready for packaging would be structured as:
 
 Then, running `python setup.py bdist_egg` will create a file `dist/my_job_package-0.0.0-py2.7.egg`.
 
+## Uploading a Package
+
+Whereas Java and Scala jobs are packaged as Jar files, Python jobs need to be packaged as `Egg` or `Wheel` files. A set
+of example jobs can be build using the `job-server-python/` sbt task `job-server-python/buildPyExamples`. This builds an
+examples Egg and Wheel in `job-server-python/target/python` so we could push this to the server as a job binary:
+
+    curl --data-binary @dist/my_job_package-0.0.0-py2.7.egg \
```
bsikander, line 172: As wheel is the recent format and we want to encourage people to use wheels, it might be better to have example commands using wheels instead of eggs. What is your opinion?

job-server/src/main/scala/spark/jobserver/util/JarUtils.scala
```diff
@@ -66,7 +66,7 @@ object JarUtils {
     objectRef
   }
 
-  //Jars and Eggs are both zip files
+  //Jars, Eggs and wheels are both zip files
```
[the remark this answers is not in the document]
Ennosigaeon: Done

doc/python.md
Ennosigaeon: Good point, I missed that. I have adapted the paragraph.

doc/python.md
Ennosigaeon: I have updated all examples with whl files.

bsikander:
In [this](https://github.com/Ennosigaeon/spark-jobserver/tree/wheel_support#binaries) section on readme, i think we should replace the content type with wheel also.

Ennosigaeon:
@bsikander Ok, I have made according changes.

PR #1350, opened 2021-02-15
refactor(Context Reset): Make context resets async and report exceptions

Ennosigaeon:
**Pull Request checklist**

**Current behavior :** Previously, context resets blocked the thread to wait for the result. Additionally, if any context could not be stopped, the request is still reported as handled successfully.

**New behavior :** Do not block during context shutdown. Check if all contexts have been shutdown successfully, otherwise return InternalServerError.

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1350)

### DIFF
--- a/job-server/src/main/scala/spark/jobserver/WebApi.scala
+++ b/job-server/src/main/scala/spark/jobserver/WebApi.scala
@@ -1,8 +1,5 @@
 package spark.jobserver
 
-import java.net.MalformedURLException
-import java.util.NoSuchElementException
-import java.util.concurrent.TimeUnit
 import akka.actor.{ActorRef, ActorSystem}
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
 import akka.http.scaladsl.model._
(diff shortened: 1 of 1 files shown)


job-server/src/main/scala/spark/jobserver/WebApi.scala
```diff
@@ -908,7 +917,7 @@ class WebApi(system: ActorSystem,
                             case _: MalformedURLException => completeWithException(
                               ctx, "JOB LOADING FAILED: Malformed URL", StatusCodes.BadRequest, err)
                             case _ => completeWithException (
-                              ctx, "JOB LOADING FAILED", StatusCodes.BadRequest, err)
+                              ctx, "JOB LOADING FAILED", StatusCodes.InternalServerError, err)
```
valan4ik, line 920: Why do you think that InternalServerError is better?
Job loading error usually shows that request was incorrect, some problem with a job.

It feels a bit unconnected to the PR (POST /jobs endpoint) and it changes REST API responses a bit (which should be reflected in the https://github.com/spark-jobserver/spark-jobserver/blob/master/doc/WebApi/swagger.yaml#L405)

Ennosigaeon: Hm, I can't recall why I changed it in the first place. I will revert it tomorrow.

Ennosigaeon: Ok, it is gone now.

valan4ik (approved):

Issue #1342, opened 2021-01-15
Update JavaSparkJob to reflect new API and/or clarify verify usage


murraytodd:
I'm curious if there's a reason the JavaSparkJob interface hasn't been updated for the new API style where validation returns a generic JobData type that gets used by the run method.

To what I understand, the current Java API is designed so the application writer is supposed to write their verify methods to throw a RuntimeException if a validation problem is detected. Two reactions to this:
1. It's weird. I usually associate runtime exceptions with things like null pointers or indexoutofbounds mistakes. I assume this choice was made to avoid having to declare the exception in a "throws" signature.
2. It would be really helpful to document this somewhere. (Javadoc comment on the verify method and/or in the `javaapi.md` file.

I'd be happy to contribute to improve this, but I figured I would ask first in case there's a specific reason we've taken this approach.

murraytodd:
Sorry, but I was looking as issue #692 and see that the new API has actually been created. I just didn't notice it because (a) there's no mention in the main README.md and more importantly (b) the doc/javaapi.md document (which is where I first went to look into how to do Java) tells to use the old JSparkJob method and (c)`job-server-api/src/main/java/spark/jobserver/api/TestJob.java` also demonstrates using the old JSparkJob interface.

If someone can confirm that I'm not missing something, I'll be happy to assembly a pull request to tidy this up.

valan4ik:
Hi @murraytodd ,

Your code snippet in the second comment is from `JavaJob`. I briefly checked the project and there are `JavaJob` and `JavaSparkJob`:
https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server-api/src/main/scala/spark/jobserver/JavaSparkJob.scala
https://github.com/spark-jobserver/spark-jobserver/blob/master/job-server-api/src/main/scala/spark/jobserver/japi/JavaJob.scala

`JavaSparkJob` seems to have a new API:

valan4ik:
Regarding your Java/Scala question, Scala is highly compatible with Java (if that is what was confusing you):
> The standard Scala backend is a Java VM. Scala classes are Java classes, and vice versa. You can call the methods of either language from methods in the other one. You can extend Java classes in Scala, and vice versa. The main limitation is that some Scala features do not have equivalents in Java, for example traits.

https://www.scala-lang.org/old/faq/4

murraytodd:
[...]

PR #1357, opened 2021-03-18
Refractor java job interface

Ennosigaeon:
**Pull Request checklist**

**Current behavior :** #1342

**New behavior :**
I have taken a closer look at the Java API in preparation for the Spark 3 migration and give the API redesign a shot. 

Important Changes:
- Rename `spark.jobserver.japi.JavaJob` to `spark.jobserver.japi.WrappedJavaJob` to indicate that this is not the base class for Java jobs
- Add a new generic to `spark.jobserver.japi.BaseJavaJob`. `D` = `JobData`, `R` = `JobOutput` and `C` = `SparkContext`. `verify` takes a `Config` and returns `D`. `run` takes a `D` and returns `R`. I also tried to rename both methods to reflect the spark API even closer, however this breaks downward compatibility.
- Add a new `spark.jobserver.japi.JavaSparkJob` that replaces the previous `spark.jobserver.japi.JSparkJob` but with `R` and `D`
- Adapt existing `spark.jobserver.japi.JSparkJob` to default `D` to `Config` mimicking the previous behavior.
- Deprecating `spark.jobserver.SparkJobValidation` as it is only used in the deprecated `spark.jobserver.SparkJobBase`


murraytodd (commented):
In the java-api.md documentation, in the JavaConfigurableJob Basic Example, could we make it so that the data we extract is an Int, and could we put a hypothetical requirement that the number be positive? (I.e. show how the user is idiomatically supposed to handle verification failures by throwing a runtime exception—which is what I _believe_ the intent is!)

That would not only help the reader understand how to do validation, but it would make it easier to see the D vs R typing in play.

One last really trivial thing: in the markdown headers, I think GitHub isn't displaying them correctly because there aren't any spaces between the hashes and the titles. (I.e. "###Basic Example" should be "### Basic Example".

murraytodd:
Okay, I've got one more question about the intended (idiomatic) approach to coding validation problems: should the user throw a specific exception type (i.e. `spark.jobserver.japi.ValidationException`) or just make up any RuntimeException? If the former, would it make sense to add a `throws` clause to the verify method signature? I know that by establishing a convention of "throw a runtime exception so that we don't have to declare it" has a level of convenience, but since this is already a Java API, maybe it's best to follow Java-style conventions.

Also, while I'm digging in this rabbithole... what if we designed the `spark.jobserver.japi.ValidationException` to be optionally configurable with multiple Validation issues? I.e. if the user wants to write validation code carefully enough to catch a variety of configuration problems rather than just the "first one caught" it would give the ability to leverage the fuller parity with the Scala API.

Thoughts?

Ennosigaeon:
@murraytodd: My biggest issue is keeping the API backwards compatible, so I can't make all the changes that I would like to make. As users were previously able to raise any RuntimeException, we have to keep the interface that way. So raising any Exception is the _intended_ way to go. To make this more explicit, I have added `throws RuntimeException` to the verify method.

The idea of being able to have multiple Validation Issues is neat. I am not sure how to implement it best/correctly, so I will leave it for later. If you would like, you can give it a try.

murraytodd:
@Ennosigaeon I understand about backwards compatibility, and I'm fine with keeping the throwing of any RuntimeExceptions an acceptable pattern. My biggest concern is just making sure we provide the users some clear instructions in the documentation (java-api.md) as to the preferred idiomatic usage.

I'll think about the multiple Validation issue approach. We could suggest the idiomatic approach is to throw that `spark.jobserver.japi.JavaValidationException` for a single issue, and we could create a parallel `spark.jobserver.japi.MultiJavaValidationException(issues: Seq[String])` that allowed multiple separate issues to be returned, and then the `WrappedJavaJob#verify` wrapper code could have a pattern match case that caught that and converted it into the proper `Bad(Many(...))` structure.

doc/java-api.md
```diff
@@ -33,29 +37,30 @@ For instance the JSparkJob uses the JavaSparkContextFactory.
...
-    public String run(JavaSparkContext sc, JobEnvironment runtime, String data) {
-        return data;
+    public String run(JavaSparkContext sc, JobEnvironment runtime, Config data) {
+        return "Hi!";
     }
 
-    public String verify(JavaSparkContext sc, JobEnvironment runtime, Config config) {
-        return config.getString("input");
+    public Config verify(JavaSparkContext sc, JobEnvironment runtime, Config config) throws RuntimeException{
+        return ConfigFactory.empty();
```
murraytodd, line 60: Again, I'm worried that the reader will wonder why the configuration information is not getting passed-through to run in this example. Either, could we add an end-of-line comment saying `// No configuration information needed by run **or** we could have verify check that `! config.getString("name").isEmpty` and then pass that same config object through to run?

murraytodd:
By the way, I looked over the changes to the documentation. On behalf of the potential Java users, I appreciate you're taking the time. I added one comment on the JavaHelloWorldJob file. Apologies if I'm being nit-picky.

Ennosigaeon:
@murraytodd: Ok, I tried to implement like you suggested. Please have a look again.

Issue #1227, opened 2019-07-04
Memory leak with JobResultActor


bklostermann:
**Used Spark version**
2.2.0

**Used Spark Job Server version**
0.9.0

**Deployed mode**
EMR
context-per-jvm = true

**Actual (wrong) behavior**
We create 3 Spark Contexts and run jobs against those contexts.  After awhile those Spark Contexts will be deleted, and new ones created in its place.  SJS will eventually run out of memory.  When inspecting the heap dump, all JobResultActors remain in memory permanently (even for the Spark Contexts that were deleted).  Consequently those JobResultActors also retain their LRU cache.  Is there any way to have those old JobResultActors released?

**Steps to reproduce**
Create Spark Contexts, run jobs, and then delete contexts. Repeat.

Nibooor:
Hi,

you are right, it seems like the jobserver (or jobmanager) never terminates any result actor after the contexts are stopped.

Probably most SJS users don't encounter this problem as 
  a) job results are often written to a database directly (so resultActors typically have a minor load)
  b) a restart of the spark slave will free the memory eventually.

So unless you want to circumvent the issue by persisting job results in a database, I think a code modification would be necessary to address this leak.
Contributions are of course always welcome.

Nibooor:
See https://github.com/spark-jobserver/spark-jobserver/pull/1375

Issue #1098, opened 2018-07-24
can support to save result field in database?


lxhao:
**Used Spark version**
2.2
**Used Spark Job Server version**
(Released version, git branch or docker image version)
0.8
**Deployed mode**
(client/cluster on Spark Standalone/YARN/Mesos/EMR or default)
client 
**Actual (wrong) behavior**
 Sometimes results are very small, maybe just a HDFS path. I think it is reasonable for users to configure whether the result to be stored in the database, or if the result exceeds a certain size, it can not be saved to the database.
**Steps to reproduce**

**Logs**

```
some log
```

or as attached file (see below)

Unused parts of this template should be removed (including this line).

velvia:
Sure, this is not a bad idea

Nibooor:
See https://github.com/spark-jobserver/spark-jobserver/pull/1375

Issue #585, opened 2016-09-08
delete finished/ failed jobs from database??


jorgeakanieves:
is any possibility to remove finished/ failed jobs from database? In case that executing several times the same job (same appName) and quering the api rest, it returns the same appName (className) but with different status (FINISHED/ RUNNING/ FAILED) so it´s a bit complicated to check the "current" status of the appName...

Any way to do this via API? (Not a workaround to perfom this like bash script, jdbc client to h2,...)

noorul:
Are you not using jobId which is unique?

jorgeakanieves:
Over the same appName (same jar)  I execute several iterations so always a new jobid is created.. And the jobs list becomes huge and can't get the last execution unless I capture and control the time of execution and this is not a very smart solution...

noorul:
@jorgeakanieves Which version of SJS are you using?

jorgeakanieves:
I´m deploying jobserver with velvia/spark-jobserver:0.6.2.mesos-0.28.1.spark-1.6.1 docker. 

Another issue is that sometimes, running jobserver against yarn by yarn-client mode, jobs are finishing in yarn (by an error or other fact) but it doesn´t finish in jobserver (there´s a sync problem) so I can´t stop it in jobserver and always stay in running status.

¿?

BigFishBall:
Hi ，
     Is the problem resolved? I have the same issue: is there any way to clean up job list via APIs？
     This job list become Huge, and ... I search no answer.

thanks.

BigFishBall:
I deleted /tmp/spark-jobserver before starting the jobserver.
and joblist is cleanup。
Refer to https://github.com/spark-jobserver/spark-jobserver/issues/516#issuecomment-227986726

But I still hope there is a easy way~~   :p

Nibooor:
See https://github.com/spark-jobserver/spark-jobserver/pull/1375

PR #1375, opened 2021-06-24
Persist job results in database

Nibooor:
## Motivation

This relatively big change that @vglagoleva and I worked on is addressing shortcomings of the current implementation of the `JobResultActor` as well as some feature requests around heaving persistent job results.
As a summary (see related issues), the shortcomings are:

- Job results are currently stored in an in-memory `LRUCache` within the `JobResultActor`:
  - results are limited in size
  - results are lost over time
  - cache leads to memory leaks
  - results are lost after actor restart (specifically this means results are lost as soon as the context is terminated for every non-ad-hoc-contexts)
- Actor complications
  - HA failover (restart of a JobServer) leads to wrong/lost actor addresses & actors. I.e. both global as well as local result actors are wrong after an HA failover, meaning results are lost even though the job/context is still running.

The idea was to simply persist job results in a supported database using and extending the existent DAO infrastructure.

### Related issues
- #1098 - (Feature request) to save result in database
- #1227 - (Bug) Memory leak with JobResultActor
- #585 - (Feature request) to delete finished/failed jobs from database

## Introduced changes
- Make Job results persistent
  - Add result persistence to `JobDAOActor` by adding new messages `SaveJobResult` and `GetJobResult` that delegate to the binary DAO
  - Rename BinaryDAO to BinaryObjectsDAO
  - Extend BinaryObjectsDAO interface with saveJobResult / getJobResult methods
  - Supply SQL migrations to create a new table storing JobResults
- Replace JobResultActor with DAO
  - Remove the `JobResultActor` completely
  - Replace every previous call to JobResultActor with the new DAO messages
- Cleanup
  - Add cleanup methods to DAO that delete jobs and contexts in a final state older than a certain age
  - Expose config property `dao-cleanup-after` that - if set - schedules a regular cleanup of jobs and contexts
  - Implement new deleteJobs and deleteContexts MetadataDAO methods for all (Zookeeper, SQL, InMemory) DAOs
  - Make job cleanup also delete configs and results
  - Implement new deleteJobResult BinaryObjectsDAO method for all (HDFS, SQL, InMemory) DAOs
  - Remove now obsolete `AutoPurgeActor`

## Summary
Overall, after this change
 * JobResults can be saved and retrieved via a database more reliably
 * Old contexts and jobs (including config & results) can be regularly purged based on configuration

---
This change is [Reviewable](https://reviewable.io/reviews/spark-jobserver/spark-jobserver/1375)


vglagoleva (approved):
@Nibooor LGTM, but documentation should be updated: https://github.com/spark-jobserver/spark-jobserver/blob/master/doc/job-server-flow.md (maybe some other files as well)
Could be addressed in a separate PR though.

